PR #6, opened 2015-11-10
Fixed "*" dependencies, fixed some warnings

bvssvni:


bvssvni:
Glium is broken at the moment, so the tests doesn't compile. Merging anyway since this blocks the image library.

PR #10, opened 2015-12-29
Added Eq traits to public enums

fsh:
Adding Eq trait to public enums.

See PistonDevelopers/image-png#9


fsh:
Sorry, I fucked up~ I'm new to this thing...

PR #14, opened 2016-01-14
Fix build issue under 1.5/1.6 and get show working again

nathanaeljones:
Perhaps glium's API changed? ./show now works for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,11 @@
 # PNG Decoder
 
 (WIP) PNG decoder. Currently `pngcheck` is the most usable part. Build with `cargo build --release --example pngcheck`.
+
+
+
+Render a PNG to the display: 
+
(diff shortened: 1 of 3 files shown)


nwin:
It's the glium api. I'll have a look at this looks weird that you had to change the function signature.

tomaka:
What changed in glium is that you now have to explicitly destroy the `Frame` object by calling `finish()` or you get a panic.

The current code just does `display.draw()`, which doesn't do that.

nwin:
Yes, so one doesn’t need to change the fill_v_flipped function

        let surface = &mut display.draw();
        fill_v_flipped(&opengl_texture.as_surface(), surface, glium::uniforms::MagnifySamplerFilter::Linear);
        surface.set_finish().unwrap();

will do it.

PR #25, opened 2016-03-09
Make flate2 dependency (and PNG encoding) optional

jonas-schievink:
This allows pure-Rust PNG decoding


bvssvni:
Looks good!

jonas-schievink:
Oh, and this is a breaking change in case `default-features = false` is used

bvssvni:
Yeah, I'll merge this and publish an upgrade.

PR #26, opened 2016-03-12
Updated crates.io

bvssvni:
- Bumped to 0.4.2


bvssvni:
0.4.1 was uploaded Dec 29, 2015. I've looked through the commits since then and it seems to be non-breaking changes.

PR #37, opened 2016-07-20
Add ancillary pHYs chunk

baskerville:


nwin:
Thank you for contributing.

Can you please

- Rename the struct/enum variant to `PixelDimensions` I don’t see any benefit from using an abbreviation here. 
- Remove the line break at the end of chunk.rs, we should probably do that change anyway but in a separate PR.
- Introduce an enum for the unit? E.g.

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[repr(u8)]
/// Physical unit of the pixel dimensions
pub enum Unit {
    Unspecified = 0,
    Meter = 1,
}

impl Unit {
    /// u8 -> Self. Temporary solution until Rust provides a canonical one.
    pub fn from_u8(n: u8) -> Option<Unit> {
        match n {
            0 | 1 => Some(unsafe { mem::transmute(n) }),
            _ => None
        }
    }
}
```

As it is implemented now we both accept and output invalid values. See `BitDepth` handling for more information.

baskerville:
Updated, thanks.

nwin:
Thank you.

Issue #43, opened 2016-11-28
Empty PNG causes panic


nwin:
Reported in https://github.com/PistonDevelopers/image/issues/593.

PR #45, opened 2017-01-04
Fix empty image panic

kosinix:
Fix for panic when saving a 0x0 PNG. #43


nwin:
Thank you for contributing!

PR #46, opened 2017-01-05
Use deflate rather than flate2 for deflate encoding. Fixes #2

### LINKED ISSUE #2: Use pure Rust library instead of flate2
Since `png` depends on `flate2`, which relies on `miniz.c`, this means that to build this crate people need to have a C/C++ compiler setup, which is annoying for users on Windows who just install Rust and expect things to work.

https://github.com/PistonDevelopers/image/issues/419

oyvindln:
This changes the encoder to use deflate-rs, which is written in rust rather than flate2 which is wrapper to the c library miniz. This would fix #2.

I don't know if it should be merged yet, the roundtrip unit test seems to work fine, but it may need some more testing on other images. (There's a failing doc test, but it failed before I did any changes so I presume it's unrelated, as it complains about a missing file). I've tested the decoder on a fair number of different inputs, but there could still be bugs in it (but feel free to help out with the decoder if you find any). It's also going to be a bit slower than using flate2 for now.

My editor also ran rustfmt on the file I changed, so if that's an issue I'll change the PR to avoid that if you want.

Lastly, I wasn't sure if I should pin the full version, or if I should only depend on "0.7" to avoid having to update image-png in case of bugfixes to deflate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ exclude = [
 
 [dependencies]
 inflate = "0.1.0"
-flate2 = { version = "0.2.11", optional = true }
+deflate = { version = "0.7.1", optional = true }
 num-iter = "0.1.32"
 bitflags = "0.7"
 
@@ -26,6 +26,6 @@ features = ["glutin"]
 default-features = false
 
 [features]
-png-encoding = ["flate2"]
+png-encoding = ["deflate"]
 default = ["png-encoding"]
 unstable = []
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1,4 +1,4 @@
-extern crate flate2;
+extern crate deflate;
 
 use std::borrow::Cow;
 use std::io::{self, Write};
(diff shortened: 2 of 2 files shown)


bvssvni:
Perhaps we could use a Cargo feature to enable flate2? This way people could test it in real projects.

nwin:
I think we should just merge it like it is. The test we have passes. Encoders are usually less critical than decoders. We should have a look at the doctest nevertheless.

oyvindln:
I suspect the failing doctest is due to [this](https://github.com/PistonDevelopers/image-png/pull/44/files#diff-b4aea3e418ccdb71239b96952d9cddb6R30). Anyhow, that's a separate issue.

bvssvni:
Nice! Merging.

eddyb:
I mentioned this to @alexcrichton on IRC and he seems fine with adding a feature to `flate2` to expose its API on top of these Rust implementations - that seems like it may scale better to other existing users.

eddyb:
@bvssvni LOL that timing.

bvssvni:
@eddyb 😄

PR #70, opened 2017-10-09
update inflate 0.3.2 -> 0.3.3

tigercosmos:
update dependencies `inflate` to the latest version.
It will solve [`servo`](https://github.com/servo/servo) current panic.

r? @bvssvni

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ exclude = [
 ]
 
 [dependencies]
-inflate = "0.3.2"
+inflate = "0.3.3"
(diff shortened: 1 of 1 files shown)


bvssvni:
Have you tried `cargo update`?

tigercosmos:
Yes, I have.
I'm sorry for not quite understanding. 
So, It is fine not to change `.toml` when we have used `cargo update`?
Thanks for your answer.

nwin:
It will probably fix it. Nevertheless we need  to update the dependency. Just no need to release a new version immediately.

PR #71, opened 2017-10-11
Update bitflags to 1.0 and bump version

Eijebong:


bvssvni:
cc @nwin

bvssvni:
@Eijebong It looks good, except for small thing: Do you think `Transformations` should be used when that's more custom to use on iterators in Rust?

Eijebong:
Not sure I understand your question. Do you want me to rename that enum to something else ?

bvssvni:
I was thinking about calling it `Transformation` instead of `Transformations`.

Eijebong:
Oh... I don't think I ever saw a bitflags struct that wasn't a plural, so I think it's better to stick with `Transformations`.

bvssvni:
OK. Merging.

Issue #80, opened 2018-06-27
Unbounded memory consumption on malformed inputs


Shnatsel:
There is an integer overflow in https://github.com/PistonDevelopers/image-png/blob/99383650e1a440bb14c54987938676c8f54d3bc6/src/decoder/mod.rs#L51

Aside of posing dangers for unsafe code (which shouldn't rely on this value anyway), this overflow causes enormous amounts of memory to be _actually_ allocated when fed to `png` crate via the [fuzzing harness](https://github.com/Shnatsel/image-png/blob/master/png-afl/src/main.rs). Not just virtual memory - actual physical memory.

The worst part is, fixing this correctly requires changing the external API: the function should use [checked_mul()](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_mul) which returns `Option<usize>`, and actually return either an Option or Result to the outside.

Testcase: [integer_overflow_in_multiplication](https://user-images.githubusercontent.com/291257/41970114-4dcd90f6-7a12-11e8-9393-03251e376dc4.png) found via [afl-rs](https://github.com/rust-fuzz/afl.rs). Steps to reproduce the crash can be found in #79 except you need to build in debug mode, without the `--release` flag.

I would appreciate advice on how to proceed with fixing this issue. Is adding "deprecated" marker to this function in 0.12 series and releasing 0.13 with a breaking fix appropriate? Do we need the [semver trick](https://github.com/dtolnay/semver-trick) here?

**Update:** libpng itself also had similar issues; see https://libpng.sourceforge.io/decompression_bombs.html for more info. Among other things, they have introduced limits on the possible size of an image by default. In Rust we can easily allow the API user to override these limits via the [builder pattern](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html).

Shnatsel:
It is also possible that the overflow is not to blame, and is merely a side effect of incorrect handling of malformed files.

newpavlov:
I think good solution will be to add `Limits` struct and check against it, as for example was done in [`flif`](https://github.com/dgriffen/flif.rs) crate

Issue #79, opened 2018-06-21
Panic on malformed input


Shnatsel:
Previous fuzzing attempts did not bypass crc32 checks in `png` and adler32 checks in `inflate` crates. Thus they never actually exercised the png decoding code.

I have disabled checksum verification in fuzzing mode via conditional compilation and ran [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz). I got a crash in less than a second.

---
Details on this specific panic: it happens at line https://github.com/PistonDevelopers/image-png/blob/1266ec2/src/decoder/mod.rs#L450

Steps to reproduce:
```
git clone https://github.com/Shnatsel/png-leak.git
cd png-leak
env RUST_BACKTRACE=1 RUSTFLAGS='--cfg fuzzing' cargo run
```
This repo uses [my fork of image-png](https://github.com/Shnatsel/image-png) with crc32 conditionally disabled. `--cfg fuzzing` in the above command disables it so that fuzzer-generated files can be tested. The file that's causing the crash is included in the repo and can be downloaded [here](https://github.com/Shnatsel/png-leak/blob/8e9f299/in.png).

Shnatsel:
Another panic on malformed input: out of bounds access on line https://github.com/PistonDevelopers/image-png/blob/1266ec2967b49a4c5c0a80605e19e3268a7b7bef/src/filter.rs#L79 when bpp > len

I have a feeling that there's gonna be a lot of these, so I'm just going to put them all in this issue instead of creating a separate issue for each panic.

Shnatsel:
There is an integer overflow in https://github.com/PistonDevelopers/image-png/blob/master/src/decoder/mod.rs#L51

Any unsafe code relying on has a security vulnerability - either [information disclosure](http://seclists.org/fulldisclosure/2013/Nov/83) or arbitrary code execution. 

The worst part is, fixing this correctly requires changing the external API: the function should use [checked_mul()](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_mul) which returns `Option<usize>`. For now I'll have to make it panic on overflow, which is bad, but not as bad as the other options.

Testcase: [integer_overflow_in_multiplication](https://user-images.githubusercontent.com/291257/41970114-4dcd90f6-7a12-11e8-9393-03251e376dc4.png)

Shnatsel:
[...]

PR #81, opened 2018-06-27
Fix panics on malformed inputs, support fuzzing

Shnatsel:
Bypass crc32 and adler32 checks when compiled with fuzzing instrumentation. This lets fuzzers reach actually interesting code.

Add a starting corpus for fuzzing, obtained initially by fuzzing other tools, then fed to afl-fuzz on image-png to get more inputs specific to it, and subsequently minified with afl-cmin.

Fix panics on malformed inputs discovered via fuzzing (#79):
 - Panic on malformed paletted images in `expand_paletted()`: return `Result` instead of calling `unwrap()`
 - Out of bounds access in `unfilter()`: now checks bounds and returns Result
 - Panic in `next_interlaced_row():` `info` from previous row was erroneously used to decode the current row. Fixing this no longer triggers a panic and likely fixes decoding of some real-world PNGs.
 - Overflow on left shift described in https://github.com/PistonDevelopers/image-png/issues/79#issuecomment-400658905

Not fixed in this PR:
 - Overflow in calculation of output buffer size (#80), which allows an attacker to exhaust system memory and/or the address space of the process. Fixing this would be a breaking change for the external API.
 - Possibly some other overflows and debug mode panics that I cannot reach because of the above overflow happening before that code is reached


png-afl/Cargo.toml
```diff
@@ -1,14 +1,10 @@
 [package]
 name = "png-afl"
-version = "0.1.0"
-authors = ["nwin <nwin@users.noreply.github.com>"]
+version = "0.2.0"
+authors = ["Sergey Davidoff <shnatsel@gmail.com>", "Paul Grandperrin <paul.grandperrin@gmail.com>"]
```
bvssvni: I think nwin removed as author here by accident.

Shnatsel: I have entirely replaced the afl/ subfolder without using anything that's in there previously. Its current incarnation is based only on https://github.com/rust-fuzz/targets, which is why Paul Grandperrin is credited.

I agree nwin should be credited for his contribution in some way, but I have removed him from the copyright notice for the source files the current afl/ folder is not based on his work in any way, and such misattribution is deliberately prohibited under some licenses, e.g. some BSD variants. This crate is dual-licensed under MIT and Apache, which do not have such a clause, so crediting nwin here would not be in conflict with the license.

I will re-add him as an author if you believe that is the best way to go about it.

bvssvni: Ah, it's under the "png-afl" directory, didn't saw that.

@nwin OK with this?

Shnatsel: Should I split the panic fixes to a separate PR so they would not be held up by a copyright notice?

bvssvni: I think it's OK. @nwin can just add himself back as author if he disagrees.

bvssvni:
Merging.

PR #82, opened 2018-09-01
Limit decoded images to 2^26 pixels by default

### LINKED ISSUE #80: Unbounded memory consumption on malformed inputs

zealousidealroll:
Fixes #80

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -53,21 +53,61 @@ impl OutputInfo {
     }
 }
 
+#[derive(Clone, Copy, Debug)]
+pub struct Limits {
+    /// max number of pixels: `width * height` (default: 67M = 2<sup>26</sup>)
+    pub pixels: u64,
+}
+
+impl Default for Limits {
+    fn default() -> Limits {
+        Limits {
+            pixels: 1 << 26,
+        }
+    }
+}
+
 /// PNG Decoder
 pub struct Decoder<R: Read> {
     /// Reader
     r: R,
     /// Output transformations
     transform: Transformations,
+    /// Images that are considered too big
+    limits: Limits,
 }
 
 impl<R: Read> Decoder<R> {
     pub fn new(r: R) -> Decoder<R> {
+        Decoder::new_with_limits(r, Limits::default())
+    }
+    
+    pub fn new_with_limits(r: R, l: Limits) -> Decoder<R> {
         Decoder {
             r: r,
             transform: ::Transformations::EXPAND | ::Transformations::SCALE_16 | ::Transformations::STRIP_16,
-
+            limits: l,
         }
+    }
+
+    /// Images that are considered too big
+    ///
+    /// ```
+    /// use std::fs::File;
+    /// use png::{Decoder, Limits};
+    /// // This image is 32x32 pixels, so it's more than four pixels in size.
+    /// let mut limits = Limits::default();
+    /// limits.pixels = 4;
+    /// let mut decoder = Decoder::new_with_limits(File::open("tests/pngsuite/basi0g01.png", limits).unwrap());
+    /// assert!(decoder.read_info().is_err());
+    /// // This image is 32x32 pixels, so it's exactly 1024 pixels in size.
+    /// let mut limits = Limits::default();
+    /// limits.pixels = 1024;
+    /// let mut decoder = Decoder::new_with_limits(File::open("tests/pngsuite/basi0g01.png", limits).unwrap());
+    /// assert!(decoder.read_info().is_ok());
+    /// ```
+    pub fn set_limits(&mut self, limits: Limits) {
+        self.limits = limits;
     }
 
     /// Reads all meta data until the first IDAT chunk
@@ -85,6 +125,12 @@ impl<R: Read> Decoder<R> {
                 line_size: r.output_line_size(info.width),
             }
         };
+        let (width, height, pixels) = (info.width as u64, info.height as u64, self.limits.pixels);
+        if width.checked_mul(height).map(|p| p > pixels).unwrap_or(true) {
+            // DecodingError::Other is used for backwards compatibility.
+            // In the next major version, add a variant for this.
+            return Err(DecodingError::Other(borrow::Cow::Borrowed("pixels limit exceeded")));
+        }
         Ok((info, r))
     }
 }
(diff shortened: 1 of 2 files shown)


bvssvni:
Looks good to me.

HeroicKatora:
The limit appears somewhat arbitrary to me. Is it referring to a limit already established in another library?

HeroicKatora:
Considering that it triggers on a quadratic image of width and length each `2^13 = 8192` which is already well in range of some of the better cameras and might even become a monitor resolution in the foreseeable future. For example, IMAX is [estimated at just about 10000x7000 according to Wikipedia](https://en.wikipedia.org/wiki/Image_resolution). And Japan wants to record the 2020 olympics in [16k resolution](https://en.wikipedia.org/wiki/16K_resolution).

oyvindln:
In the BMP decoder in image there is a [max](https://github.com/PistonDevelopers/image/blob/master/src/bmp/decoder.rs#L53) of ``2^16``, though one may not expect bmp images to be quite as large as a png image (as they are not, or only very slightly compressed). 

There is also a limit on how much memory the decoder allocates at a first try, and it will only allocate more if there is actually image data for it. Maybe something similar could be done here.

Ideally though it would be nice if a limit could be specified, so the caller could say give some reasonable expected max size based on say the file size.

HeroicKatora:
The max in BMP is enforced for width and height individually, so in total limit the size to something slightly smaller than 2^32. The limit proposed here is for the actual pixel count, `width*height`.

oyvindln:
Yeah I meant to write ``2^16`` in one direction but I somehow missed it.

zealousidealroll:
The limit was copied from the [flif](https://github.com/dgriffen/flif.rs) crate.

HeroicKatora:
Ah, I didn't realize that. [In the corresponding pull request](https://github.com/dgriffen/flif.rs/pull/28), the choice was explained as follows:

> Default values allow maximum size of DecodingImage's data equal to 2^26, which is equal to 512 MB.

Now I wonder if we should differentiate between a pixel and a memory limit. `png` allows bit depths of up to 16 bit and up to 4 samples per pixel, leading to up to `16=2^4` bytes per pixel in the most extreme case. Setting a limit on the number of pixels would enforce somewhere between `32MiBi` and `1024MiBi` of memory usage, depending on the colour format.

However, that is more of a bikeshedding issue for now, with prior art demonstrated it might be sensible to just merge it for now. Increasing the default limit in the future or adding alternative and optional additional limits should not be a breaking change anyways.

src/decoder/mod.rs
```diff
@@ -53,23 +53,67 @@ impl OutputInfo {
     }
 }
 
+#[derive(Clone, Copy, Debug)]
+pub struct Limits {
+    /// max number of pixels: `width * height` (default: 67M = 2<sup>26</sup>)
+    pixels: u64,
+}
```
newpavlov: I think there shouldn't be any problems with making `pixels` field public. This will also allow us to remove `set_pixels` setter.

src/decoder/mod.rs
```diff
@@ -53,23 +53,67 @@ impl OutputInfo {
...
+    /// assert!(decoder.read_info().is_err());
+    /// // This image is 32x32 pixels, so it's exactly 1024 pixels in size.
+    /// let mut limits = Limits::default();
+    /// limits.set_pixels(1024);
+    /// let mut decoder = Decoder::new(File::open("tests/pngsuite/basi0g01.png").unwrap());
+    /// decoder.set_limits(limits);
+    /// assert!(decoder.read_info().is_ok());
+    /// ```
+    pub fn set_limits(&mut self, limits: Limits) {
+        self.limits = limits;
```
newpavlov: How about using `new_with_limits(r: R, limits: Limits)` method instead? It will allow to create decoder in one line and `new` can be defined as `Self::new_with_limits(r, Default::default())`.

Shnatsel:
For reference, libpng enforces 1,000,000 by 1,000,000 pixels limit by default, which is roughly 2^40 pixels or 2^44 bytes per image in the extreme case.

zealousidealroll:
> 2^40 pixels

* Even with a measly 256 colors (so it would be 2^40 bytes), that's bigger than the entire address space on a 32-bit system. If nothing else, the limits should never be higher than 2^29 on a 32-bit system (making 2^31 bytes at 2^2=4 bytes per pixel, which will exactly fill the userland-accessible address space on a 32-bit system where the upper half is reserved for the operating system). If you think anyone's ever going to need higher, then make it conditional on the platform.

* According to my quick Wikipedia visit, an [8K picture](https://en.wikipedia.org/wiki/8K_resolution) is [2^25 pixels](https://duckduckgo.com/?q=log(7680*4320)+%2F+log(2)&t=ffab&ia=calculator). In the immortal words of Bill Gates, 8K should be enough for anybody.

bvssvni:
Merging.

PR #83, opened 2018-10-01
WindowEvent::Close to WindowEvent::CloseRequested

eclipseo:
It was a breaking change in winit 0.13.0:

>Breaking: the Closed event has been replaced by CloseRequested and 
>Destroyed. To migrate, you typically just need to replace all usages of 
>Closed with CloseRequested; see example programs for more info.


bvssvni:
Thanks!

PR #84, opened 2018-10-01
Use the new LogicalSize with set_inner_size

eclipseo:
Change introduced in winit 017.2:

>- **Breaking:** The entire API for sizes, positions, etc. has changed. 
In the majority of cases, winit produces and consumes positions and 
sizes as `LogicalPosition` and `LogicalSize`, respectively.


bvssvni:
Thanks!

PR #88, opened 2018-10-13
 check data size before encoding

lovebug356:
copy_from_slice (used in write_image_data) panics if the line length is not as expected.

This patch checks if the data_size is as expected.


bvssvni:
Thanks! Merging.

PR #89, opened 2018-10-13
bump dependencies

lovebug356:


bvssvni:
There is no point in bumping patch versions. A `cargo update` is sufficient.

lovebug356:
Indeed, this only adds explicitly that this newer dependencies work, where this is implicitly assumed (but not validated) by semantic versioning. This came "for free" with an editor plugin, feel free to close if you don't like unnecessary version bumps.

bvssvni:
Closing.

PR #91, opened 2018-10-18
Replace slice conversion with vec drain

acgaudette:
I was running into a memory issue on my end, and this fixed it. Incidentally should be more performant as well (no extra allocation).


bvssvni:
Thanks! Merging.

PR #93, opened 2018-10-31
Add APNG Encode Support

rrbutani:
Missing support for a couple of features, but seems to work.

This probably isn't ready to merge as is, but I thought I'd make the PR anyways to get some feedback.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -9,9 +9,10 @@ use std::mem;
 use std::result;
 
 use crc32fast::Hasher as Crc32;
+use deflate::write::ZlibEncoder;
 
 use crate::chunk;
-use crate::common::{BitDepth, ColorType, Compression, Info};
+use crate::common::{AnimationControl, BitDepth, ColorType, Compression, FrameControl, Info};
 use crate::filter::{filter, FilterType};
 use crate::traits::WriteBytesExt;
 
@@ -67,6 +68,25 @@ impl<W: Write> Encoder<W> {
         Encoder { w, info }
     }
 
+    pub fn new_animated(w: W, width: u32, height: u32, frames: u32) -> Result<Encoder<W>> {
+        if frames > 0 {
+            let mut encoder = Encoder::new(w, width, height);
+
+            let animation_ctl = AnimationControl { num_frames: frames, num_plays: 0 };
+            let mut frame_ctl = FrameControl::default();
+            frame_ctl.width = width;
+            frame_ctl.height = height;
+
+            encoder.info.animation_control = Some(animation_ctl);
+            encoder.info.frame_control = Some(frame_ctl);
+
+            Ok(encoder)
+        } else {
+            Err(EncodingError::Format("invalid number of frames for an animated PNG".into()))
+        }
+
+    }
+
     pub fn write_header(self) -> Result<Writer<W>> {
         Writer::new(self.w, self.info).init()
     }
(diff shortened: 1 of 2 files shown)


abonander:
@rrbutani can you rebase against latest master? Looks like there's some merge conflicts.

@bvssvni @nox I've run into a use-case for this, I'd like to see some movement on it. I'll be testing this fork myself to see how it works once it's been rebased.

bvssvni:
@abonander Nice if you can get a look.

abonander:
@rrbutani `FrameControl` should be on a per-frame basis since each frame can have configurable parameters. In fact, I would almost suggest an entirely different `Encoder` struct that's constructed directly with an `AnimationControl` instance.

src/encoder.rs
```diff
@@ -136,7 +191,110 @@ impl<W: Write> Writer<W> {
...
+            Info { animation_control: Some(AnimationControl { num_frames: 0, ..}) , frame_control: Some(_), ..} => {
+                Err(EncodingError::Format("exceeded number of frames specified".into()))
+            },
+            Info { animation_control: Some(anim_ctl), frame_control: Some(frame_control), ..} => {
+                match frame_control.sequence_number {
+                    0 => {
+                        let ret = if self.separate_default_image { // If we've already written the default image we can write frames the normal way
+                            // fcTL + fdAT
+
+                            self.write_fcTL().and(self.write_fdAT(data))
```
abonander: This kind of chaining could lead to data corruption as the `write_fdAT()` operation will be executed regardless of whether or not `write_fcTL()` succeeds. These should instead be wrapped with `try!()` like all the other fallible operations (although the `?` operator has been stable for a long time now, `try!()` would be consistent with the rest of the crate).

src/encoder.rs
```diff
@@ -136,7 +191,110 @@ impl<W: Write> Writer<W> {
...
+        (&mut data[22..]).write_be(frame_ctl.delay_den)?;
+        data[24] = frame_ctl.dispose_op as u8;
+        data[25] = frame_ctl.blend_op as u8;
+
+        self.write_chunk(chunk::fcTL, &data)
+    }
+
+    #[allow(non_snake_case)]
+    fn write_fdAT(&mut self, data: &[u8]) -> Result<()> {
+        // println!("Writing fdAT:{:?}", self.info.frame_control.unwrap().sequence_number+1);
```
abonander: These printlns should be removed once the PR is ready to be merged.

rrbutani:
Thanks for the feedback @abonander; I'll rebase and make the suggested changes later today.

fintelia:
Is this ready to be merged? (Other than the couple conflicts)

HeroicKatora:
There seem to be two potentially independent changes here:

* Interpreting `DisposeOp` and `BlendOp` and erroring accordingly during decoding.
* Support writing animated data stream and encode `fcTL` block.

As far as I can see only the former is SemVer relevant while the latter can be done in a point-release after the first has landed. I'd like to release a new version of `image-png` and `image` relatively soon so it would be nice to either get this merged but only having the public representation changes ready is also fine.

Rimpampa:
I opened a pull request on the @rrbutani fork: https://github.com/rrbutani/image-png/pull/1

I made it work but some features are missing, like the option to set a different framerate for each frame or dividing each frame imago into more fcTL chunks (like with the `StreamWriter`)

I'm willing to continue contribuiting on this but I'd like to hear some opinions on how to make it better, as right now it doesn't feel right.

HeroicKatora:
@Rimpampa I think you should make your own PR so we don't need to wait on the fork's author to respond (it's been quite a while). Then also rebase those commits to resolve the merge conflict. Let me know if you'd like any help with that, I've done it once before on this one :)

HeroicKatora:
Closed/Moved to #255

PR #94, opened 2018-11-13
Use io::Write::write_all

nox:
It's incorrect to use `io::Write::write` and ignore how many bytes were actually written in the underlying writer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+/.idea
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ bitflags = "1.0"
 getopts = "0.2.14"
 term = "0.4"
 glob = "0.2.11"
+rand = "0.5.5"
 
 [dev-dependencies.glium]
 version = "0.22"
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -94,7 +94,7 @@ impl<W: Write> Writer<W> {
     }
 
     fn init(mut self) -> Result<Self> {
-        try!(self.w.write(&[137, 80, 78, 71, 13, 10, 26, 10]));
+        try!(self.w.write_all(&[137, 80, 78, 71, 13, 10, 26, 10]));
         let mut data = [0; 13];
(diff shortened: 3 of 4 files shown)


marshallpierce:
This snippet demonstrates the issue, extracted from me tinkering in base64 to see what was going on:

```

#[test]
fn png() {
    let mut rng = rand::thread_rng();
    let mut input_bytes = Vec::<u8>::new();
    let mut png_data = Vec::<u8>::new();

    // A really ugly 200x200 rgba
    for _ in 0..(200 * 200 * 4) {
        input_bytes.push(rng.gen());
    }

    // encode into a Vec
    self::image::png::PNGEncoder::new(&mut png_data)
        .encode(&input_bytes, 200, 200, self::image::ColorType::RGBA(8))
        .unwrap();

    // and b64 the vec
    let simple_b64 = encode_config(&png_data, ::STANDARD);

    // encode into a b64 streaming encoder
    let mut stream_png_b64 = Vec::new();

    {
        let mut b64_encoder = EncoderWriter::new(&mut stream_png_b64, ::STANDARD);

        self::image::png::PNGEncoder::new(&mut b64_encoder)
            .encode(&input_bytes, 200, 200, self::image::ColorType::RGBA(8))
            .unwrap();

        b64_encoder.finish().unwrap();
    }

    assert_eq!(simple_b64, ::std::str::from_utf8(&stream_png_b64).unwrap());
}
```

With debug logging in base64 added to show when the streaming encoder can't write the full input it's given, I get this:

```
input = 160236, consumed = 768
```

I'm working on getting a failing test case and possibly a fix.

marshallpierce:
I changed the `roundtrip` test to use all available test images, which worked fine but only for images whose `line_size` was 32, so the test skips all other images. For the affected images, encoding fails with errors like `"wrong data size, expected 65536 got 196608"`. What's up with that? Just a guess, but perhaps the decode buffer is sized conservatively, but since `next_frame` doesn't return the number of bytes decoded, there's no way to only encode the relevant portion of the decode buffer?

bvssvni:
Merging.

nox:
@bvssvni Can you make a new png release, please?

PR #98, opened 2019-01-01
[+] Allow caller to specify compression level via setter.

JackMordaunt:
> For context, refer to this [issue](https://github.com/PistonDevelopers/image/issues/849)

- Compression and filter fields added to existing Info struct, since that seems to be it's purpose.
- Default to `deflate::Compression::Fast` and `filter::FilterType::Sub` to maintain backward compatible output.
- Compression enum is defined by deflate crate.

Implemented as a generic setter through the `Parameter` trait.
This simplifies the usage and conforms to the established
pattern.

One caveat with this approach is requiring the caller to import
`deflate` and `png::HasParameters`.

Example use of the api:

```rust
use std::io::prelude::*;
use png::{self, HasParameters};
use deflate;

/// Encodes the image ```image```
/// that has dimensions ```width``` and ```height```
/// and ```ColorType``` ```c```
pub fn encode<W: Write>(
    w: W,
    data: &[u8],
    width: u32,
    height: u32,
    ct: png::ColorType,
    bits: png::BitDepth,
) -> io::Result<()> {
    let mut encoder = png::Encoder::new(w, width, height);
    encoder
        .set(ct)
        .set(bits)
        // Note that we can now set the compression level and filter type.
        // If we don't manually set these, they default to
        // `deflate::Compression::Fast` and `png::FilterType::Sub` respectively,
        // which were the previously hard-coded values.
        .set(deflate::CompressionOptions::default())
        .set(png::FilterType::Paeth);
    let mut writer = encoder.write_header()?;
    writer.write_image_data(data).map_err(|e| e.into())
}
```

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -9,7 +9,7 @@ use std::result;
 
 use chunk;
 use crc::Crc32;
-use common::{Info, ColorType, BitDepth};
+use common::{Info, ColorType, BitDepth, Compression};
 use filter::{FilterType, filter};
 use traits::{WriteBytesExt, HasParameters, Parameter};
 
@@ -81,9 +81,12 @@ impl<W: Write> Parameter<Encoder<W>> for BitDepth {
     }
 }
 
-impl<W: Write> Parameter<Encoder<W>> for deflate::CompressionOptions {
+/// Set compression param for a ```Compression``` or any type that can transform 
+/// into a ```Compression```, notably ```deflate::Compression``` and 
+/// ```deflate::CompressionOptions``` which "just work".
+impl<W: Write, C: Into<Compression>> Parameter<Encoder<W>> for C {
     fn set_param(self, this: &mut Encoder<W>) {
-        this.info.compression = self
+        this.info.compression = self.into()
     }
 }
 
@@ -139,7 +142,7 @@ impl<W: Write> Writer<W> {
             let message = format!("wrong data size, expected {} got {}", data_size, data.len());
             return Err(EncodingError::Format(message.into()));
         }
-        let mut zlib = deflate::write::ZlibEncoder::new(Vec::new(), self.info.compression);
+        let mut zlib = deflate::write::ZlibEncoder::new(Vec::new(), self.info.compression.clone());
         let filter_method = self.info.filter;
         for line in data.chunks(in_len) {
             current.copy_from_slice(&line);
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -81,6 +81,12 @@ impl<W: Write> Parameter<Encoder<W>> for BitDepth {
     }
 }
 
+impl<W: Write> Parameter<Encoder<W>> for deflate::CompressionOptions {
+    fn set_param(self, this: &mut Encoder<W>) {
+        this.info.compression = self
+    }
+}
+
 /// PNG writer
 pub struct Writer<W: Write> {
     w: W,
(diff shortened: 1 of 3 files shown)


JackMordaunt:
Ignore the force-push. I undid a commit that was me attempting to figure out how to import a crate from a fork.

src/lib.rs
```diff
@@ -60,5 +60,6 @@ pub use common::*;
 pub use decoder::{Decoder, Reader, OutputInfo, StreamingDecoder, Decoded, DecodingError, Limits};
 #[cfg(feature = "png-encoding")]
 pub use encoder::{Encoder, Writer, EncodingError};
+pub use filter::FilterType;
```
HeroicKatora: Regarding this, all other members of `Info` are public members in `pub use common::*`. Not having a public re-export of `deflate::Compression` but using that as a public member in a struct feels a bit strange. Two possibilities:

1. Commit to `deflate::Compression` and add a public export.
2. Create a simple `struct` wrapper like `pub struct Compression(deflate::Compression);` to `common` which hides the representation, and add a `impl From<deflate::Compression> for Compression` as well as a replica of the constants. Replace the info attribute to use that type instead, and unpack its representation in the encoder code.

On the plus side 1. is simpler but one should consider that [`deflate::CompressionOptions`](https://docs.rs/deflate/0.7.19/deflate/struct.CompressionOptions.html) allow even more control over the output compression. It would be easier to extend option 2. in a backwards compatible way to introduce these additional possibilities at a future point. At the same time, method 2. would not require adding an extra crate to every users dependency list (in edition 2015) or crate usage (in edition 2018). Basically a question of whether the additional consistency of 2. comes at a cost of convenience.

src/encoder.rs
```diff
@@ -139,7 +142,7 @@ impl<W: Write> Writer<W> {
             let message = format!("wrong data size, expected {} got {}", data_size, data.len());
             return Err(EncodingError::Format(message.into()));
         }
-        let mut zlib = deflate::write::ZlibEncoder::new(Vec::new(), self.info.compression);
+        let mut zlib = deflate::write::ZlibEncoder::new(Vec::new(), self.info.compression.clone());
```
JackMordaunt: I think it's possible to avoid using clone but I thought the convenience is worth it, since:

1. It's only a once-off initialisation cost.
2. A tagged enum variant with no internal data is basically just an integer (I would assume), so the allocation is tiny.

HeroicKatora: I wouldn't worry about it too much for any static size.

JackMordaunt:
Is there anything blocking this from being merged? :)

bvssvni:
Merging.

bvssvni:
Thanks!

HeroicKatora:
Yep, thanks a lot :heart:

nox:
This introduced calls to `deflate` methods which aren't in `deflate` 0.7.2, which is the minimum version of the dependency requested by `image-png`. A build should be added to CI that checks that the project still compiles with the oldest versions possible of dependencies.

nox:
I filed https://github.com/PistonDevelopers/image-png/pull/101

src/encoder.rs
fintelia: Bit late now, but if you had also derived Copy the clone call wouldn't be needed. (The generated code will be the same either way)

HeroicKatora: `Copy` is a commitment that it will stay `Copy`. I'm not sure.

PR #99, opened 2019-01-02
Remove instances of unsafe code

fintelia:
This commit changes the interface of StreamingDecoder to not return references
to internal state. This makes it much easier to implement in a way that
satisfies the borrow checker, without resorting to unsafe.

It isn't completely clear whether this change is completely desirable because it
can incur additional copying overhead, but probably won't for common use cases.


bvssvni:
Travis CI fails on Rust stable with the following error:

```
error[E0554]: #![feature] may not be used on the stable release channel
  --> src/lib.rs:44:1
   |
44 | #![feature(nll)]
   | ^^^^^^^^^^^^^^^^
```

PR #101, opened 2019-01-22
Bump deflate to 0.7.12

nox:
CompressionOptions::huffman_only doesn't exist in deflate 0.7.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: rust
-rust:
-  - stable
-  - nightly
+matrix:
+  include:
+    - rust: stable
+    - rust: nightly
+    - rust: nightly
+      env: FLAGS="-Z minimal-versions"
 script:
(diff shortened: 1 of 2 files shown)


HeroicKatora:
As bad as it seems, the functions in `deflate` [simply fill in some members](https://docs.rs/deflate/0.7.19/src/deflate/compression_options.rs.html#171-178). I think we'd rather duplicate their implementations and make a note to upgrade at the next semver breaking version bump.

nox:
What? Bumping deflate is not a breaking bump, given it's a compatible bump. There is no need to duplicate anything here.

Making a minor bump to a dependency is not a brealking bump.

HeroicKatora:
Ah you're right, I was under the impression this bumped more than the minor version.. Sorry, yeah go ahead then.

jdm:
It would be nice to merge this!

PR #102, opened 2019-02-01
Add a streaming writer

### LINKED ISSUE #75: Images must be written all at once or not at all
Hello,

I'd like to write images so large I cannot store them uncompressed in memory.

As far as I can tell, with the current state of the library, this is impossible. It seems to me that you must either call `png::Writer::write_image_data` with the entire image or not at all. No other function seems to be a candidate for the functionality I'm searching for - we seem to have a `StreamingDecoder` but no `StreamingEncoder`.

I'd love to be proven wrong obviously :-) or any advice you can give for this request. It *is* technically possible - libpng implements it via the `png_write_rows` function.

birktj:
This adds a streaming png writer so that you can create images that do not fit in memory. I choose to not update `Writer::write_image_data` as the `StreamWriter` has a few layers of buffering that may reduce performance a little bit. I selected a default chunk size of 4k, this number was a bit out of thin air, however there are only a few extra bytes per chunk so I don't think it will be a problem.

Fixes #75

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -3,7 +3,7 @@ extern crate deflate;
 use std::borrow::Cow;
 use std::error;
 use std::fmt;
-use std::io::{self, Write};
+use std::io::{self, Read, Write};
 use std::mem;
 use std::result;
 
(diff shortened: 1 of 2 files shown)


fintelia:
This all looks good to me. Other than that one conflict in lib.rs, is there anything else needed before merging this?

birktj:
@fintelia I have rebased on master now. I am reasonably sure that this should be ready for merging.

fintelia:
Thanks!

udscbt-wsx:
Hi, when will `StreamWriter` be in a future release?
Also, is it possible to make `StreamWriter` to own a `Writer` underneath instead of borrowing it, so that we can eliminate the lifetime parameter in `StreamWriter<'a, W>`?

PR #109, opened 2019-04-14
Update crate to Rust 2018

Atul9:
Update crate to Rust 2018 using 'cargo fix --edition'


fintelia:
Hm, the automated results don't really look that nice. The tool seems to have converted all the `try!` macros to `r#try!`. I think it would be worth making a manual pass cleaning things up and switching to using `?` everywhere.

Atul9:
@fintelia done. I have updated the PR.

fintelia:
Thanks!

PR #112, opened 2019-04-19
Fix pngcheck and travis tests

### LINKED ISSUE #111: Travis ci does not run `cargo test`
It seems like the travis ci script does not run `cargo test`, is this intentional? Currently `cargo test` fails because of some syntax errors in `examples/pngcheck.rs`, this should probably be fixed asap.

HeroicKatora:
Closes #111 and the associated error in tests. Use the opportunity to 
rework some code flow, avoiding deeply nested loops and the need
for ad-hoc lambda function that is instantly called for error handling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,4 @@ matrix:
 script:
   - cargo build -v $FLAGS
   - cargo doc -v $FLAGS
+  - cargo test -v $FLAGS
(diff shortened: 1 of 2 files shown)


HeroicKatora:
`cargo test` fails on `minimal-versions`, but not the other builds.

Issue #113, opened 2019-04-19
Travis-CI fails on minimal versions


HeroicKatora:
The command `cargo +nightly test -Z minimal-versions` does not succeed on Windows machines, due to resolving to version `0.0.1` of `winapi`. This crate version contains invalid syntax. Two options:

1. Do not execute `cargo test` on `nightly` with `minimal-versions`
2. Declare minimum versions that resolve correctly

HeroicKatora:
There are in fact a lot of crates with dependencies that do not work properly with `minimal-versions`:

* `term = 0.4` -- `winapi = 0.0.1`!
* `glium = 0.22` -- `winapi = 0.0.1`!
* `glium = 0.24` -- `lazy_static = 0.1.0`!

See #119 for the full dependency chains. Especially `glium` complicates the matter as there is no newer version available and the project seems to be in maintenance-only mode. This makes the first option more appealing.

PR #115, opened 2019-04-23
Update limits to be byte based

### LINKED ISSUE #105: Limit buffer size instead of image dimensions
The `Limits` struct should limit buffer size instead of image dimensions. With https://github.com/PistonDevelopers/image/pull/860 I had remove the limits in order to decode large images, with a limit on buffer size one could decode huge images while keeping a reasonable limit on memory usage. 

Also see discussion in https://github.com/PistonDevelopers/image-tiff/issues/29.

birktj:
Limiting allocated bytes is more effective and useful than image
size. Especially when incrementally decoding an image which may not
fit in memory, limiting pixels doesn't make much sense. Also see the discussion in https://github.com/image-rs/image-tiff/issues/29

This will change the public api in that the `Limits` struct now has a different field.

This fixes #105 

@HeroicKatora @fintelia @Shnatsel  what do you guys think?

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -54,15 +54,16 @@ impl OutputInfo {
 }
 
 #[derive(Clone, Copy, Debug)]
+/// Limits on the resources the `Decoder` is allowed too use
 pub struct Limits {
-    /// max number of pixels: `width * height` (default: 67M = 2<sup>26</sup>)
-    pub pixels: u64,
+    /// maximum number of bytes the decoder is allowed to allocate, default is 64Mib
+    pub bytes: usize,
 }
 
 impl Default for Limits {
     fn default() -> Limits {
         Limits {
-            pixels: 1 << 26,
+            bytes: 1024*1024*64,
         }
     }
 }
@@ -73,7 +74,7 @@ pub struct Decoder<R: Read> {
     r: R,
     /// Output transformations
     transform: Transformations,
-    /// Images that are considered too big
+    /// Limits on resources the Decoder is allowed to use
     limits: Limits,
 }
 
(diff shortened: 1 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -425,9 +422,16 @@ impl<R: Read> Reader<R> {
         len + match extra { 0 => 0, _ => 1 }
     }
 
-    fn allocate_out_buf(&mut self) {
+    fn allocate_out_buf(&mut self) -> Result<(), DecodingError> {
         let width = get_info!(self).width;
-        self.processed = vec![0; self.line_size(width)]
+        let bytes = self.limits.bytes;
+        if bytes < self.line_size(width) {
+            // FIXME: DecodingError::Other is used for backwards compatibility.
```
HeroicKatora: This reasoning doesn't make much sense, as this is a backwards incompatible commit set–a public member of `Limits` was renamed from `pixels` to `bytes`.

birktj: I simply copied code from the previous limit stuff, but yeah I realize that this doesn't really make sense. I will add a new error variant.

src/decoder/mod.rs
```diff
@@ -54,15 +54,16 @@ impl OutputInfo {
 }
 
 #[derive(Clone, Copy, Debug)]
+/// Limits on the resources the `Decoder` is allowed too use
 pub struct Limits {
-    /// max number of pixels: `width * height` (default: 67M = 2<sup>26</sup>)
-    pub pixels: u64,
+    /// maximum number of bytes the decoder is allowed to allocate, default is 64M
```
Shnatsel: It might be helpful to clarify that this is Mib, not Mb. See https://en.wikipedia.org/wiki/Mebibyte

Shnatsel:
Overall I think this is definitely a step in the right direction.

CI failure on nightly seems to be a rustc bug, could anyone familiar with the CI setup comment on that?

HeroicKatora:
@Shnatsel That's just #113, a side effect of having the tests ran in all configurations in Travis.

src/decoder/mod.rs
birktj: Thanks, I always wondered which was which

HeroicKatora: There's also 'the decoder will allocate less than 10K bytes' somewhere below, as `10*1024`.

birktj:
Since nobody else has commented this I will merge it.

Issue #116, opened 2019-04-25
Support for encoding metadata chunks


birktj:
Currently the encoder lacks built-in support for writing metadata chunks, you can do it manually with `Writer::write_chunk`. However I believe that there should be a nicer high-level way of doing this. 

See also https://github.com/image-rs/image/issues/911

jansol:
There is an addition to the spec that adds support for PQ-encoded HDR images: https://www.w3.org/TR/png-hdr-pq/ . This relies on `iCCP`, with fallbacks to `cHRM` and `gAMA`. Krita reportedly supports this format for HDR exports.

However adding support for HDR images is probably going to require some API redesign...

zlstringham:
+1 for encoder AND decoder support of text chunks.
I'm currently working on a library to support a proprietary file format for a game engine with the following specifications:
- Image is tiled (think like a sprite sheet)
- Each frame is specified by metadata in the zTXt chunk

Unfortunately, without access to read/write the zTXt chunk, I currently can't use this crate for my needs.
(Image format is the DMI image format used by [BYOND](http://www.byond.com).)

de-vri-es:
`gAMA` and `cHRM` have been merged with #244 :tada: I opened a PR #260 for the sRGB chunk. I can't tick the boxes, but it would be nice to update the list for potential contributors :)

SirWindfield:
Hey!

Since I am going to use the library (especially the metadata part of it), I'd like to contribute back by working on some of the chunks here. Maybe some guidance on how to approach a PR would be nice, for example for the iTXT chunk.

Is the overall goal to add some new struct and a field to `Info` so that decoder/encoder can populate it? If so would the appropriate type for iTXT be `Vec<NewiTXTChunkStruct>`?

de-vri-es:
> Is the overall goal to add some new struct and a field to `Info` so that decoder/encoder can populate it? If so would the appropriate type for tEXT be `Vec<NewtEXTChunkStruct>`?

That sounds logical to me (but note that I'm not a maintainer).

You could still wonder if you want to provide the chunk data without processing, or if you want to decompress the text during decoding. Getting decompressed data would be nice, but might allow malicious PNGs to contain small compressed text that expands to large amounts of memory.

For that reason, it might be wise to store the data compressed in the info struct, and provide an easy interface to perform the decompression as needed. That also prevents wasting time and memory on decompression when the application doesn't care about the text chunk.

PR #118, opened 2019-04-28
Add write_to_dpi and write_image_data_with_dpi

3442853561:
https://github.com/image-rs/image-png/issues/116

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -161,7 +161,7 @@ impl<W: Write> Writer<W> {
     }
 
     pub fn write_image_data_with_dpi(&mut self, x: u32, y: u32, unit: bool, data: &[u8]) -> Result<()> {
-        self.write_to_dpi(x, y, unit)?
+        self.write_to_dpi(x, y, unit)?;
         self.write_image_data(data)
     }
 
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -153,6 +153,17 @@ impl<W: Write> Writer<W> {
         }
         self.write_chunk(chunk::IDAT, &zlib.finish()?)
     }
+    
+    pub fn write_to_dpi(&mut self, x: u32, y: u32, unit: bool) -> Result<()>  {
+        let _x = x.to_be_bytes();
+        let _y = y.to_be_bytes();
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -153,6 +153,17 @@ impl<W: Write> Writer<W> {
         }
         self.write_chunk(chunk::IDAT, &zlib.finish()?)
     }
+    
+    pub fn write_to_dpi(&mut self, x: u32, y: u32, unit: bool) -> Result<()>  {
```
HeroicKatora: That design makes it possible to write a `dpi` chunk after an image chunk. However, [the specification](https://www.w3.org/TR/PNG/#5ChunkOrdering) only permits it before the first `IDAT` chunk. This *at least* deserves better documentation. Maybe a better place would be to place it into the `Encoder` and write the `pHYs` information as part of the header in https://github.com/image-rs/image-png/blob/e2c1152f9138d6f441c1356fc7444482bab85a3a/src/encoder.rs#L65 ?

And `unit` should not be a `bool` but a proper enum (with conversion to `u8`). Otherwise, the values leave one guessing which meaning they have.

src/encoder.rs
```diff
@@ -153,6 +153,17 @@ impl<W: Write> Writer<W> {
...
+        let _x = x.to_be_bytes();
+        let _y = y.to_be_bytes();
+        self.write_chunk(chunk::pHYs, &[_x[0], _x[1], _x[2], _x[3], _y[0], _y[1], _y[2], _y[3], unit as u8])
+    }
+
+    pub fn write_image_data_with_dpi(&mut self, x: u32, y: u32, unit: bool, data: &[u8]) -> Result<()> {
```
HeroicKatora: I think this misinterprets what `x` and `y` in an `pHYs` chunk are. For the `write_image_data` call they need to specify the size of the image pixel matrix, but for the `pHYs` they specify the (aspect-ratio of) the *physical* size of each individual pixel. Most devices would have at least somewhat quadratic pixels and reusing the image size is definitely wrong.

I don't think this method is necessary for a convincing PR.

3442853561:
@HeroicKatora Thank you for your suggestion! I will submit another PR later. There's something wrong with GitHub Desktop, and I have been unable to clone this project.

3442853561:
https://github.com/image-rs/image-png/pull/122

PR #119, opened 2019-04-28
[Don't merge yet] Update glob dependency

HeroicKatora:
Tries to fix the tests under minimal-versions in the Travis matrix. Some dependency
resolves to `winapi=0.0.1` in that condition and this is a faulty version that was not
yanked. The guess here is that `getopts` is affected since it did depend on `tempdir`
but now that has been relegated to a `dev-dependency`. We'll see.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,12 +18,12 @@ bitflags = "1.0"
 
 [dev-dependencies]
 getopts = "0.2.14"
-term = "0.4"
-glob = "0.2.11"
+term = "0.5"
+glob = "0.3"
 rand = "0.5.5"
 
 [dev-dependencies.glium]
-version = "0.22"
+version = "0.24"
 features = ["glutin"]
 default-features = false
 


HeroicKatora:
No, it is in fact `advapi32-sys` which declares:
* `winapi = *`

Now, to find out which package depends on advapi32-sys

fintelia:
I recommend giving [`cargo tree`](https://github.com/sfackler/cargo-tree) a try for tracking down issues like this

HeroicKatora:
Yes, but I need to track down what happens on Travis... And can't reproduce it locally :grimacing:

HeroicKatora:
Oh god, it is in fact `winapi = 0.2.4` which depends on `advapi32-sys = *` which depends on `winapi = *`

HeroicKatora:
So, effectively I should have read about `minimal-versions` in `cargo` before this since https://github.com/rust-lang/cargo/pull/5757 has the same problem ^^

HeroicKatora:
`libloading = 0.4.0` also depends on `winapi = 0.2`

HeroicKatora:
```
   Compiling lazy_static v0.1.0
     Running `rustc --crate-name lazy_static /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/lazy_static-0.1.0/src/lazy_static.rs --color always --crate-type lib --emit=dep-info,link -C debuginfo=2 -C metadata=b92219cd1010870c -C extra-filename=-b92219cd1010870c --out-dir /home/travis/build/image-rs/image-png/target/debug/deps -L dependency=/home/travis/build/image-rs/image-png/target/debug/deps --cap-lints allow`
error[E0658]: non-builtin inner attributes are unstable
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/lazy_static-0.1.0/src/lazy_static.rs:74:1
   |
74 | #![license = "MIT"]
   | ^^^^^^^^^^^^^^^^^^^
   |
   = note: for more information, see https://github.com/rust-lang/rust/issues/54726
   = help: add #![feature(custom_inner_attributes)] to the crate attributes to enable
```

Oh no

HeroicKatora:
* `glium = 0.24` (newest) depends on
* `glutin = ^0.20` depends on
* `shared_library = ^0.1.0` depends on
* `lazy_static = ^0.1` which fails

PR #123, opened 2019-04-29
Make it possible to checkout the repo under Windows

HeroicKatora:
File paths containing ':' can not be checked out under Windows. This was
probably the reason that cargo refuses to publish a package containing
files with such file names. The exact seed name should however not
matter much, in the worst case it can probably be restored locally.

See https://github.com/image-rs/image/issues/915


HeroicKatora:
@Shnatsel This is kind of awkward as it affects all repos which contain `afl` fuzzing seeds generated from a minimized, but automatically discovered fuzzing corpus. It occurred to me that you might be interested in this issue, and potentially have ideas on workaround, or better workflows to change that?

Shnatsel:
Ouch. Uuuh, you can safely rename the files, AFL doesn't really care what they're called. References to files from unit tests would have to be updated (if any), but that's it.

I don't think this is going to be solved in AFL upstream as there is no longer an AFL upstream, but you could open a feature request against afl-rs I guess?

PR #122, opened 2019-04-29
In order to process this issue #116

3442853561:
In order to process this issue https://github.com/image-rs/image-png/issues/116

PR #124, opened 2019-04-30
Yet another PR like #122

3442853561:
Yet another PR like https://github.com/image-rs/image-png/pull/122
Add set_dpi

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -9,7 +9,7 @@ use std::result;
 
 use crate::chunk;
 use crate::crc::Crc32;
-use crate::common::{Info, ColorType, BitDepth, Compression};
+use crate::common::{Info, ColorType, BitDepth, Compression, PixelDimensions, Unit};
 use crate::filter::{FilterType, filter};
 use crate::traits::{WriteBytesExt, HasParameters, Parameter};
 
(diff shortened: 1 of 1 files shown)


HeroicKatora:
You don't need to open a new PR by the way, `git` allows you to force push the branch and Github then updates the PR automatically. Not sure if this is possible through the web-interface.

fintelia:
Yeah, the commands would be something like:

```
$ git add foo.rs bar.rs
$ git commit --amend
$ git push --force-with-lease
```

HeroicKatora:
Unfortunately, the unchecked cast from `f64` to `u32` in `from_dpi` [may not be as safe as one would like](https://doc.rust-lang.org/reference/expressions/operator-expr.html#semantics). Even if, as currently expected, it were to adopt saturating semantics the value could be unexpectedly lower than intended. It's also slightly confusing to have `set_dpi` but no `PixelDimensions::from_dpi` to parallel it.

The PR is already good without `set_dpi` as it is pure utility. Can we address and design that in separate issue?

HeroicKatora:
The conversion from inch to meter also follows from the *definition* of an inch as `25.4mm`. Hence, the more precise number to use is `(1000.0/25.4)` but that shouldn't matter for most calculations.

fintelia:
Superseded by #403

PR #128, opened 2019-05-04
remove duplicate crc32 algorithm, fixes #127

### LINKED ISSUE #127: Remove duplicated crc32 algorithm
There's a crc32 implementation in src/crc.rs

A duplicate also lives in tests/check_testimages.rs

This issue is for reusing the code from src/crc.rs and deleting the duplicated code in tests/check_testimages.rs

martinlindhe:
Note that in order to have this, crc module is now public.

I'm not sure that's desired.

Another option could be to drop the internal crc32 algo completely and instead rely on a separate crate for this, such as https://crates.io/crates/crc


martinlindhe:
PR dropped in favor for other options as discussed in #127

PR #130, opened 2019-05-04
enable benchmark using libtest

martinlindhe:
Re-enabling commented out benchmark.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,4 @@ default-features = false
 png-encoding = ["deflate"]
 default = ["png-encoding"]
 unstable = []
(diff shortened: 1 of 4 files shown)


martinlindhe:
Noticed https://github.com/image-rs/image/tree/master/benches uses libtest for benchmarks.

I'll rework this PR to do the same instead.

martinlindhe:
PR rewritten to use libtest similar to how http://github.com/image-rs/image does it

martinlindhe:
Since no objections and this is not touching any critical parts of the code, I'll go ahead and merge.

PR #133, opened 2019-05-05
png-afl: document usage and minor tweaks

### LINKED ISSUE #132: png-afl fuzzer is broken
The afl fuzzer target in ./png-afl currently exits with SIGABRT as soon as any decoding errors occurs, causing AFL to think for example "CRC mismatch" is a crashing input file.

martinlindhe:
fixes #132


png-afl/Cargo.toml
```diff
@@ -4,7 +4,8 @@ version = "0.2.0"
 authors = ["Sergey Davidoff <shnatsel@gmail.com>", "Paul Grandperrin <paul.grandperrin@gmail.com>"]
 
 [dependencies]
-afl = "0.4.0"
+afl = "0.4.3"
 png = {path = "../"}
 
-
+[features]
+fuzzing = []
```
HeroicKatora: The feature would need to be placed in the `Cargo.toml` of the main library, and enabled by include the changing the reference line above to:

`png = { path = "../", features = ["fuzzing"] }

martinlindhe: Does your proposed way have advantages?

I tested this change locally by adding to png-afl/src/main.rs something like
```rust
fn main() {
    if cfg!(fuzzing) {
        panic!("fuzzing on!");
    }
    ...
}
``` 
rebuilt and invoked the binary directly

HeroicKatora: The code in `src/main.rs` will be compiled with `cfg(fuzzing)` but not the artifacts that come from the included `png` library as far as I am aware.

martinlindhe: That makes sense. I'm still not successfully fuzzing then... :-)

HeroicKatora: Features are local to the crate that declare them.

HeroicKatora: Hmmm. It seems that `cargo afl build` sets the `--cfg fuzzing` itself.

HeroicKatora: Because `cfg(fuzzing)` doesn't actualy select on a feature, the syntax for that would be `cfg(feature = "fuzzing")`.

HeroicKatora: Yeah, it does. https://github.com/rust-fuzz/afl.rs/blob/e6fe251b42e09a9b512da7c0db11041c0a5653bf/src/bin/cargo-afl.rs#L156

HeroicKatora: So, we've been fuzzing correctl all along, regardless of feature and RUSTFLAGS :+1:

martinlindhe: PR updated

martinlindhe: Huh. Missed your comments there :)

martinlindhe: Okay cool, dropping that stuff then

martinlindhe: PR updated with removed fuzzing feature

martinlindhe:
@HeroicKatora thank you for your help!

Shnatsel:
You probably also want the `-d` flag in that AFL command.

The default mode for AFL is "deterministic mode" which does a number of deterministic transformations of the input such as cycling every byte in the input through all values from 0 to 255, setting 4-byte chunks to interesting values, etc. When running on CI you'd probably want the "havoc" mode that just transforms the inputs randomly and doesn't run the exhaustive-ish deterministic search, which is what the `-d` flag activates.

The ideal setup is actually non-deterministic and deterministic modes running in parallel, run these commands in several terminal tabs / tmux shells / screen sessions:
```bash
cargo afl fuzz -M master -m 200 -i fuzzing_seeds -o out target/debug/png-afl
cargo afl fuzz -S slave1 -m 200 -i fuzzing_seeds -o out target/debug/png-afl
cargo afl fuzz -S slave2 -m 200 -i fuzzing_seeds -o out target/debug/png-afl
```
and so on for however many cores you have.

Shnatsel:
Oh yeah, fuzzing in release mode also works fine and is obviously faster. You'll be missing out on the extra checks from debug mode though.

Also you might want to enable Address Sanitizer to look for memory safety issues, see https://github.com/japaric/rust-san for more info. However, this is only relevant if your code or one of your dependencies uses unsafe code, and last time I checked image-png had no `unsafe` in it.

birktj:
According to https://doc.rust-lang.org/std/macro.debug_assert.html it should be possible to enable `debug_assert!` on release builds with `-C debug-assertions`, I would guess that is a good idea for fuzzing so that is both quite fast and you have the extra checks.

Shnatsel:
There is also a toggle for checked arithmetic, not sure if `-C debug-assertions` also toggles that or not. These tricks are getting quite generic, I think you should upstream this into Rust AFL guide: https://github.com/rust-fuzz/book

HeroicKatora:
Both of these checks are enabled as part of `cargo afl build`: https://github.com/rust-fuzz/afl.rs/blob/e6fe251b42e09a9b512da7c0db11041c0a5653bf/src/bin/cargo-afl.rs#L157-L167

Shnatsel:
Oh, that's quite surprising behavior! That should definitely be documented.

Issue #136, opened 2019-05-11
Writing PLTE chunk for indexed color files


antonok-edm:
It appears there's currently no support for writing a PLTE (palette) chunk to a new PNG image file.

e.g., the following code will write out a new image successfully:

```rust
    let file = File::create(PATH).unwrap();
    let ref mut w = BufWriter::new(file);

    let mut encoder = png::Encoder::new(w, WIDTH, HEIGHT);
    encoder.set(png::ColorType::Indexed).set(png::BitDepth::Eight);
    // Set color palette?
    let mut writer = encoder.write_header().unwrap();

    writer.write_image_data(&IMAGE_DATA).unwrap();
```

... but the resulting file is unusable because there is no palette to map the image data to colors. Opening the new file in any other program will fail because the PLTE chunk is missing.

Ideally, there should be a way to provide a palette to the encoder, and attempting to write an indexed color file without first setting a palette should fail.

cosmicchipsocket:
Looks like PLTE encoding is in, but tRNS is still needed.

kvark:
It would be great if this library also checked that the palette has the expected size. I happen to get the broken PNG saved when accidentally providing a size that is not expected - https://github.com/image-rs/image-png/issues/97#issuecomment-665429159

PR #138, opened 2019-05-22
Add support for encoding with fixed palette

malthe:
Note that the color type must be manually set to 'indexed'.

This fixes issue #136.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -189,6 +189,13 @@ impl<W: Write> Writer<W> {
     /// Writes the image data.
     pub fn write_image_data(&mut self, data: &[u8]) -> Result<()> {
         const MAX_CHUNK_LEN: u32 = (1u32 << 31) - 1;
+
+        if self.info.color_type == ColorType::Indexed && self.info.palette.is_none() {
+            return Err(EncodingError::Format(
+                "can't write indexed image without palette".into(),
+            ));
+        }
+
         let bpp = self.info.bytes_per_pixel();
         let in_len = self.info.raw_row_length() - 1;
         let prev = vec![0; in_len];
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -67,6 +67,10 @@ impl<W: Write> Encoder<W> {
         Encoder { w, info }
     }
 
+    pub fn set_palette(&mut self, palette: Vec<u8>) {
+        self.info.palette = Some(palette);
+    }
+
     pub fn write_header(self) -> Result<Writer<W>> {
         Writer::new(self.w, self.info).init()
     }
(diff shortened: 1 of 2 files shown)


src/encoder.rs
```diff
@@ -104,6 +109,17 @@ pub struct Writer<W: Write> {
     info: Info,
 }
 
+fn write_chunk<W: Write>(w: &mut W, name: [u8; 4], data: &[u8]) -> Result<()> {
```
birktj: Is there a reason for using this function instead of `Writer::write_chunk`?

HeroicKatora: It seems already used in `Writer::new` but the method requires `&self`. The freestanding function however does not.

birktj: Yeah, but this function was used in `init` which takes `&self`.

HeroicKatora: Oh stupid me. Github's doesn't actual give the right context. I thought line 135 was still within `new`.

birktj: No, sorry that was wrong, `init` takes `mut self`, but the function should still be usable.

malthe: I struggled to both borrow mutably self and immutably its palette. That's why I split it up so that I only had to borrow mutably the writer.

HeroicKatora: Ah I see. Well, one standard trick you can employ is to `take` from an `Option` in that case:
```
// Temporarily take remove it from `self`.
if let Some(palette) = self.info.palette.take() {
    self.write_chunk(chunk::PLTE, &palette)?;
    // Put it back in, we check that later.
    self.info.palette = Some(palette)
}
```

HeroicKatora: But there's nothing wrong with having it as a separate method, actually. I'd even go so far and say it is nicer? It was only a bit confusing to have it included in this feature addition and not as part of a code polish.

HeroicKatora: The only thing I'd change is the exact function definition:
```
// `w: W` instead of `w: &mut W`
fn write_chunk<W: Write>(w: W, name: [u8; 4], data: &[u8]) -> Result<()> {
```
That is because `Write` is automatically implemented for `&'_ mut impl Write` and so this is slightly more generic.

birktj: Yes, I have nothing against it. I only wondered why it ended up like that.

malthe: It won't accept that:
```
error[E0596]: cannot borrow `w` as mutable, as it is not declared as mutable
   --> src/encoder.rs:113:5
    |
112 | fn write_chunk<W: Write>(w: W, name: [u8; 4], data: &[u8]) -> Result<()> {
    |                          - help: consider changing this to be mutable: `mut w`
113 |     w.write_be(data.len() as u32)?;
    |     ^ cannot borrow as mutable
```

HeroicKatora: `fn ...(mut w: W, ` since you use it mutably within the function, yeah

malthe: Oh cool – learning something here :-)

nico-abram:
Is there a reason this hasn't been merged?

fintelia:
I'm not sure. I just made a pass through it, and everything seems reasonable to me.

@birktj @HeroicKatora This PR looks like it just needs a rebase and then should be ready to accept. Am I missing anything?

birktj:
@fintelia I think that is correct.

HeroicKatora:
I'm not sure about the added dependency but otherwise looks fine.

HeroicKatora:
Thank you! Took a while from our side.

PR #146, opened 2019-06-19
Remove call to dbg!()

HeroicKatora:


HeroicKatora:
Failure is from an ICE on `minimal-versions` target (reported [here](https://github.com/rust-lang/rust/issues/61804) as well). If false positives occur frequently it would be possible to mark it as allowed-to-fail.

PR #148, opened 2019-07-04
Use 'dyn' since trait objects without an explicit 'dyn' are deprecated

Atul9:


HeroicKatora:
Looks fine. Nothing major so feel free to self-merge this when checks have ran (you should have the access rights to do so).

HeroicKatora:
Note: `-Z minimal-version` being broken does not look like your fault to me, although it is a *very* interesting crash.

PR #150, opened 2019-07-07
Remove the num-iter dependency.

RazrFalcon:
`num-iter` is used just for `range_step` iterator. By using `std::iter::step_by` we can remove a lot of bloat:

```
`-- num-iter v0.1.39
    |-- num-integer v0.1.41
    |   `-- num-traits v0.2.8
    |       [build-dependencies]
    |       `-- autocfg v0.1.4
    |   [build-dependencies]
    |   `-- autocfg v0.1.4 (*)
    `-- num-traits v0.2.8 (*)
    [build-dependencies]
    `-- autocfg v0.1.4 (*)
```

This also decreases the build time: 4.9s vs 6.7s

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ exclude = [
 [dependencies]
 inflate = "0.4.2"
 deflate = { version = "0.7.12", optional = true }
-num-iter = "0.1.32"
 bitflags = "1.0"
 crc32fast = "1.2.0"
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,8 +44,6 @@
 
 #[macro_use] extern crate bitflags;
 
-extern crate num_iter;
-
 pub mod chunk;
 mod decoder;
 #[cfg(feature = "png-encoding")]
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,6 +1,5 @@
 //! Utility functions
 use std::iter::repeat;
-use num_iter::range_step;
 
 #[inline(always)]
 pub fn unpack_bits<F>(buf: &mut [u8], channels: usize, bit_depth: u8, func: F)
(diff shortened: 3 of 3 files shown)


HeroicKatora:
Do we even need this anymore? `(start..stop).step_by(step)` seems to achive the very same thing from the standard library. Is that a performance difference?

RazrFalcon:
I've wanted to use `step_by` too, but it can step only by `usize` and this code uses `isize`.

HeroicKatora:
`bit_depth` is a `u8` initially, so that should be no problem. In fact, the line should be `bit_depth.into()` to avoid making it seems like a lossy conversion.

RazrFalcon:
`range_step` used more than once.

HeroicKatora:
Oh, right :+1:

HeroicKatora:
@RazrFalcon I'm kind of confused now. You said it is required to be used for `isize` as well but all usages cast to `usize` eventually. It seems all using functions expect the actual range of values to always be positive because they are used as indices (only within `utils.rs`). So, it should be possible to replace the `isize` in all iterators with carful application of `.rev()`, `.step_by()` and `RangeInclusive` (`0..=buf.len()`).

HeroicKatora:
E.g.

```
pub fn expand_trns_line(buf: &mut[u8], trns: &[u8], channels: usize) {
    let i = (0..=buf.len()/(channels+1)*channels - channels).rev().step_by(channels);
    let j = (0..=buf.len() - (channels+1)).rev().step_by(channels+1);
    let channels = channels as usize;
    for (i, j) in i.zip(j) {
        if &buf[i..i+channels] == trns {
            buf[j+channels] = 0
        } else {
            buf[j+channels] = 0xFF
        }
        for k in (0..channels).rev() {
            buf[j+k] = buf[i+k];
        }
    }
}
```

RazrFalcon:
Yes. I've simply replaced the current functions without investigating the code that much. I'm not familiar with the code, so I don't know why `isize` is used. I guess it simply steps from `n` to `0`.

I will try to change it to `step_by`.

RazrFalcon:
Done.

HeroicKatora:
Great work, thank you! Want to join `image-rs`? You can  invited to the `image-rs` collaborators team, which allows you access to this and the other repos if you want. See [the contribution policy](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/image-rs/organization/blob/master/CODE_OF_CONDUCT.md).

RazrFalcon:
Thanks, but I'm pretty busy with my own project already :smile:

HeroicKatora:
If you ever change your mind, just ask (and probably link to this PR). Happy coding :smile:

PR #152, opened 2019-07-09
Allow -Z minimal-versions to fail

HeroicKatora:
The actual build fails already (hence it was disabled by a scripted if).
With these changes, the build is ran but allowed it to fail. In its
current state, minimal versions can at most give an indication on the
health of our dependencies but is not required (as indicated by the
disabled build). Since this is also prone to the most failures from
nightly, it has only little bearing on the decision of merging a PR and
releases. The standard nightly target is still enabled regardless.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,11 @@ matrix:
       env: FLAGS="--no-default-features"
     - rust: nightly
       env: FLAGS="-Z minimal-versions"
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Clarification: If `minimal-versions` becomes available on `stable` then it should be enabled again.

PR #153, opened 2019-07-09
Interpret more IHDR options during parsing

HeroicKatora:
This prepares the public api for encoding of animated PNGs without
implementing any of the code yet. All relevant options are de- and
encoded already but not otherwise relevant for computation.

Thanks to @rrbutani for #93

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1,6 +1,8 @@
 //! Common types shared between the encoder and decoder
 use crate::filter;
 
+use std::fmt;
+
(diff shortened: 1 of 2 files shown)


HeroicKatora:
@abonander Mind taking a look, you reviewed the other PR as well.

@rrbutani Your PR is not forgotten, it's just a breaking change and we might want to make a release at some point. This allows us to accept your PR in a point release.

PR #156, opened 2019-07-22
Change doc example to use set_* methods.

ian-hamlin:
Hi,

I just noticed the documentation was now slightly wrong with the new version, this change fixes that.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,15 +25,14 @@
 //! use std::path::Path;
 //! use std::fs::File;
 //! use std::io::BufWriter;
-//! // To use encoder.set()
-//! use png::HasParameters;
 //!
 //! let path = Path::new(r"/path/to/image.png");
 //! let file = File::create(path).unwrap();
 //! let ref mut w = BufWriter::new(file);
 //!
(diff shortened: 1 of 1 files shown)


ian-hamlin:
> Good catch! The test is marked `ignore` to avoid actually saving a file when running tests, maybe it should be marked `no-run` instead, to make sure it stays compilable after the fix?

That's an excellent idea, I've made the change.

ian-hamlin:
That broke one part of the build, so I went back to ignore for now :)

src/lib.rs
```diff
@@ -25,15 +25,14 @@
 //! use std::path::Path;
```
HeroicKatora: It is possible to wrap this whole test in a hidden `#[cfg(…)]` even in a doctest (note the preceding `#`). That should exclude the block when the `encoder` feature is not enabled and allow you to use `no_run` again.

```
//! # #[cfg(feature = "encoder")] {
…
//! # }
```

HeroicKatora:
Just noticed that I approved these but forgot to merge. Thank you!

PR #157, opened 2019-07-23
Fix typos

ConnorSkees:
Fixed some small errors I found while reading the documentation :)

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -233,7 +233,7 @@ impl<'a, W: Write> Drop for ChunkWriter<'a, W> {
 /// Streaming png writer
 ///
 /// This may silently fail in the destructor, so it is a good idea to call
-/// [`finish`](StreamWriter::finish) or [`flush`](StreamWriter::flush) before dropping. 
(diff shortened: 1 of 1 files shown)


HeroicKatora:
@ConnorSkees The automatic links derived from symbol names only work on `nightly` rust but do not yet work on `stable`. So unfortunately the links need to point to the exact docs directly. But the first link is on the same page and can utililize the standard url anchor syntax of `#method.finish` which resolves to the html id `method.finish` that is generated for the inherent `fn finish(&self)` method .

ConnorSkees:
It should be valid to write links in [this way](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md#additions-to-the-documentation-syntax). They both work correctly when testing just now (the former points to `.../image-png/target/doc/png/struct.StreamWriter.html#method.finish` and flush now goes to https://doc.rust-lang.org/nightly/std/io/trait.Write.html#tymethod.flush)

HeroicKatora:
The tracking issue for intra doc links is: https://github.com/rust-lang/rust/issues/43466

HeroicKatora:
Thanks for noticing @ConnorSkees and thank you for the review @kaj

PR #158, opened 2019-08-10
Fix documentation example

RazrFalcon:


src/lib.rs
```diff
@@ -33,7 +33,8 @@
 //! let ref mut w = BufWriter::new(file);
 //!
 //! let mut encoder = png::Encoder::new(w, 2, 1); // Width is 2 pixels and height is 1.
-//! encoder.set(png::ColorType::RGBA).set(png::BitDepth::Eight);
+//! encoder.set_color(png::ColorType::RGBA);
+//! encoder.set_depth(png::BitDepth::Eight);
```
HeroicKatora: This example is marked `ignore`, that's why CI hasn't caught it before. Replacing this with `no_run` might be better.

RazrFalcon:
@HeroicKatora Looks like it was disabled because `Encoder` is optional

HeroicKatora:
Two possible solutions:

* Add a guard within the example, e.g. wrap it in a block marked`#[cfg(feature="png-encoding")]`.
* Make the annotation (`no_run` vs. `ignore`) depend on the feature configuration. This is cleaner, and I've seen the trick myself only recently. But you can use `cfg_attr` to insert parts of the documentation. So you start the doc-test in two different manners depending on configuration:
  ```
  #[cfg_attr(feature = "png-encoding", doc = "```no_run")]
  #[cfg_attr(not(feature = "png-encoding"), doc = "```ignore")]
  ```

RazrFalcon:
Done, but looks like it was already fixed in the master.

HeroicKatora:
Huh. Sorry then @RazrFalcon

PR #161, opened 2019-08-31
Fix clippy warnings

JohnTitor:
- replace `as *` with `*::from()`
- use `write_all` instead of `write`

etc.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -16,9 +16,9 @@ pub enum ColorType {
 
 impl ColorType {
     /// Returns the number of samples used per pixel of `ColorType`
-    pub fn samples(&self) -> usize {
+    pub fn samples(self) -> usize {
         use self::ColorType::*;
-        match *self {
+        match self {
             Grayscale | Indexed => 1,
             RGB => 3,
             GrayscaleAlpha => 2,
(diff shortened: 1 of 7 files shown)


src/common.rs
```diff
@@ -16,9 +16,9 @@ pub enum ColorType {
 
 impl ColorType {
     /// Returns the number of samples used per pixel of `ColorType`
-    pub fn samples(&self) -> usize {
+    pub fn samples(self) -> usize {
```
JohnTitor: This changes the `samples`' argument from `&self` to `self`. I'm not sure it's reasonable.

fintelia: This is probably fine. I assume most users call this function via `c.samples()` rather than `ColorType::samples(&c)` so I wouldn't expect much/any breakage.

HeroicKatora: We could always check instead of guess. Get the larger reverse dependencies [from crates.io's list](https://crates.io/crates/png/reverse_dependencies) and try to compile them with the changes.

HeroicKatora: Those that actually depend on `0.15` anyways

JohnTitor: We should check manually?

HeroicKatora: It was just an idea to be extra cautious. I don't expect any larger breakage either, and it's certainly easy to fix, so this is optional.

HeroicKatora: Would be cool if there was a tool for automating this though ;)

JohnTitor: Okay, let's merge this and see whether anything happens.

PR #168, opened 2019-09-19
LibFuzzer, cargo-fuzz and Fuzzit

HeroicKatora:
Integrates CI checks on the dynamically updated corpus of decoder test cases from [fuzzit.dev](fuzzit.dev)¹. The end goal is to have fuzzing running with continually updated builds of the master branch but this will give us a large coverage in any case and to expand this to other repositories as well.

The seed of the current corpus was generated with a ~1 hour manual run of `cargo-fuzz`, with the `decoder` fuzzing harness included here as well.

-----

¹ More on that soon. @fintelia, can I add you as an additional admin on fuzzit simply to increase bus factor?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,12 @@ matrix:
       env: FLAGS="--no-default-features"
     - rust: nightly
       env: FLAGS="-Z minimal-versions"
+    - rust: nightly
+      env:
+        - FUZZIT_TARGET="png-decoder"
+        - FUZZIT_BINARY="decode"
+      services:
+        - docker
   allow_failures:
     - rust: nightly
       env: FLAGS="-Z minimal-versions"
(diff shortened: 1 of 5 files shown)


fintelia:
Feel free to add me as an admin on Fuzzit. I've created an account via GitHub login, so I think my username is just "fintelia".

HeroicKatora:
Found you, it's based on email. Well, and my display name is also my email and I can't find where to change that but it's not a big deal.

Shnatsel:
There is already a good corpus in png-afl/fuzzing_seeds. Was it incorporated in the corpus used for fuzzit?

HeroicKatora:
Yes, it was the initial seed for the much larger corpus that's being tested in CI right now.

PR #171, opened 2019-10-27
Update dev-dependency glium to 0.25

HeroicKatora:
The previous version could error in a very deep dependency `cocoa`,
only required on osx. This seems to have regressed silently.


HeroicKatora:
I don't think we can feasibly maintain a dependency of `glium` and a compiler version of `1.34.2`. Note: this concerns **only the examples** and not the main code. To keep the CI status meaningful, I'm thinking of checking the examples on stable compilers only.

PR #174, opened 2019-11-13
Only expand bit depths less than 8 to 8

aschampion:
When using the `EXPAND` transform, match the libpng behavior by only
changing the bit depth to 8 if it is less than 8. This also fixes
broken/wrong buffer size decodings when the bit depth is greater than 8
but the `EXPAND` transform is set.

This is also necessary for supporting 16 bit per channel `ImageBuffer`s and `DynamicImage` in a PR I'm nearly done with for `image`, without breaking the current behavior for 1/2/4 bit images.

Unfortunately tests are currently broken on OSX, so I'll see if this passes CI.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -351,7 +351,7 @@ impl<R: Read> Reader<R> {
                 16 if t.intersects(
                     crate::Transformations::SCALE_16 | crate::Transformations::STRIP_16
                 ) => 8,
-                _ if t.contains(crate::Transformations::EXPAND) => 8,
+                n if n < 8 && t.contains(crate::Transformations::EXPAND) => 8,
                 n => n
             };
             let color_type = if t.contains(crate::Transformations::EXPAND) {


aschampion:
Rebasing on `master` fixed the tests due to #171

fintelia:
Thanks!

Shnatsel:
> This also fixes broken/wrong buffer size decodings when the bit depth is greater than 8 but the EXPAND transform is set.

Could you clarify what exactly happens in this case? Does the decoding panic, or does it leave part of the buffer uninitialized, or something else?

aschampion:
This code gets called several times in a row by functions calling each other, so it's a bit convoluted, but it [returns an error here](https://github.com/image-rs/image-png/blob/master/src/decoder/mod.rs#L253), because [here](https://github.com/image-rs/image-png/blob/master/src/decoder/mod.rs#L384) the bit depth is used to determine the output line size for a check. If `EXPAND` sets the bit depth to 8 instead of 16, it thinks the line size has to be twice as big as it does.

I should add some 16 bit tests to this crate and not just in the `image` crate (especially given all the "TODO: 16 bit" comments in the code), but I'm not sure when I'll find time to do that.

Shnatsel:
Oh, so it merely says "supplied buffer is too small to hold the image". Not a security concern then! Thanks!

HeroicKatora:
I'm not precisely sure what happens if the buffer is large enough (it might be in general, it's supplied by the user, but not in the case of usage through `image`). But even if the buffer is large enough, there will be no uninitialized memory involved as the buffer must already be initialized and there is no `unsafe` code involved.

PR #175, opened 2019-11-17
Release 0.15.1

HeroicKatora:


HeroicKatora:
Look at the diff from `v0.15` there was a very minor breaking interface change that I've reverted. Clippy won't be happy be it's not part of our travis policy (yet). Probably my review comment [here](https://github.com/image-rs/image-png/pull/161#discussion_r320019810) was badly worded. I meant to say that checking other crates were option when that change was made was optional, it seems that it was interpreted as saying that the change itself is fine for minor version updates.

PR #176, opened 2019-11-24
Remove incorrect notice in Transformations documentation

Osspial:
Previously, it claimed that only IDENTITY and TRANSFORM_EXPAND | TRANSFORM_STRIP_ALPHA worked, but this is no longer true.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -331,9 +331,7 @@ impl Info {
 
 
 bitflags! {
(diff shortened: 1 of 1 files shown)


fintelia:
It would probably be better to still somehow indicate which transformation do and don't work

HeroicKatora:
See: #252

PR #178, opened 2019-12-17
Update to deflate 0.8

cuviper:


fintelia:
Thanks!

HeroicKatora:
This requires a major version bump next release, there are some public `From<deflate::>` implementations.

cuviper:
Good catch, I didn't check for things like that.

Issue #179, opened 2020-01-06
Panic during save_with_format when width == 0


filcuc:
To reproduce:
1) load an image with size 1x150
```
let reader = BufReader::new(file);
let image = image::load(reader, image_format).map_err(|_| "Failed to load file".to_owned())?;
let w = image.width();  // 1
let h = image.height(); // 150
```
2) create thumbnail with size 128x128.
```
let thumbnail = Some(image.thumbnail(128, 128));
let w = thumbnail.width();   // 0
let h = thumbnail.height();  // 128
```
3) Save image
```
thumbnail.save_with_format(
   "/path/to/somwhere,
   image::ImageFormat::PNG,
)
```
4) Panic

HeroicKatora:
Thanks for the report. I was able to distill the reproduction:

```
image::RgbImage::new(0, 0).save_with_format("/tmp/foo.png", image::ImageFormat::PNG);
```

<details>
<summary>A relevant portion from the stacktrace (also with version numbers):</summary>
<pre>
  16: png::encoder::Writer<W>::write_image_data
             at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/png-0.15.2/src/encoder.rs:154
  17: image::png::PNGEncoder<W>::encode
             at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.22.3/./src/png.rs:160
  18: image::io::free_functions::save_buffer_with_format_impl
             at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.22.3/./src/io/free_functions.rs:198
  19: image::dynimage::save_buffer_with_format
             at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.22.3/./src/dynimage.rs:794
  20: image::buffer::ImageBuffer<P,Container>::save_with_format
             at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/image-0.22.3/./src/buffer.rs:786
</pre>
</details>

filcuc:
Related to this is that i've obtained the image with width = 0 during thumb creation. I think that maybe resize should take in account this corner case a put a `max(computed_width, 1)`

HeroicKatora:
This is offtopic but documented accordingly:

> The image's aspect ratio is preserved.

You can always workaround by using your own size calculation and `thumbnail_exact` instead. If you want to propose changing the calculated sizes then you should create an issue in `image`.

kaj:
[...]

PR #180, opened 2020-01-07
Error out when trying to encode an empty image

HeroicKatora:
The encoder expected all lines to be non-empty to iterate over them with
a chunk calculation. This should hold true for all valid PNG files so this 
precondition needed to be checked before writing the header data.

Closes: #179

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -115,6 +115,14 @@ impl<W: Write> Writer<W> {
     }
 
     fn init(mut self) -> Result<Self> {
+        if self.info.width == 0 {
+            return Err(EncodingError::Format("Zero width not allowed".into()));
+        }
+
+        if self.info.height == 0 {
+            return Err(EncodingError::Format("Zero height not allowed".into()));
+        }
+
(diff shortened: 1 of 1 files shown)


fintelia:
The PNG spec forbids zero size images ([source](https://www.w3.org/TR/2003/REC-PNG-20031110/#11IHDR)):

> Width and height give the image dimensions in pixels. They are PNG four-byte unsigned integers. Zero is an invalid value.

We should instead trigger an error at a different point to indicate that the dimensions requested by the user are invalid.

HeroicKatora:
Good catch. We can't error in `Encoder::new` but instead `write_header` provides a hook to insert the necessary checks. It now asserts both preconditions before even writing the signature bytes.

HeroicKatora:
Note: probably should get released as `0.15.3` which means another branch from before #178 had been applied. There was a long time without many created issues so I had hoped (and wrongly assumed) we'd get a new major version before the next bug report :grinning: I'll do the necessary work once merged.

PR #185, opened 2020-03-03
Change Cargo.toml to an explicit include list

cuviper:
The published `png 0.16.0` package contains a rather large `seed/`
directory which is presumably only needed for fuzzing.

    $ dutree -s
    [ png-0.16.0 2.93 MiB ]
    ├─ seed                   │ ██████████████│  95%      2.81 MiB
    └─ <aggregated>           │               │   4%    126.78 KiB

To prevent this from happening accidentally, I've changed the `exclude`
list in `Cargo.toml` to an explicit `include` list, containing only the
licenses, documentation, and `/src/` files.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,8 +9,12 @@ authors = ["nwin <nwin@users.noreply.github.com>"]
 repository = "https://github.com/image-rs/image-png.git"
 
 edition = "2018"
-exclude = [
-    "tests/*",
+include = [
+    "/LICENSE-MIT",
+    "/LICENSE-APACHE",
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Probably even worth a `0.16.1`. Looks reasonable, I'm wondering how useful it would be to nevertheless include `examples` and `benches`?

cuviper:
> I'm wondering how useful it would be to nevertheless include `examples` and `benches`?

I'm not sure how useful they are, but they only add 28K to the raw source, ~5K to the compressed package, so I think we might as well include them. Done.

PR #187, opened 2020-03-13
Format code using 'cargo fmt'

Atul9:


src/decoder/mod.rs
```diff
@@ -349,10 +346,13 @@ impl<R: Read> Reader<R> {
...
-                    crate::Transformations::SCALE_16 | crate::Transformations::STRIP_16
-                ) => 8,
+                    crate::Transformations::SCALE_16 | crate::Transformations::STRIP_16,
+                ) =>
+                {
+                    8
+                }
                 n if n < 8 && t.contains(crate::Transformations::EXPAND) => 8,
-                n => n
+                n => n,
```
HeroicKatora: Sometimes `cargo fmt` seems so arbitrary...

Issue #189, opened 2020-03-30
Make Writer::stream_writer(_) take ownership of the underlying Writer or provide a Writer.into_strea_writer()


hank-der-hafenarbeiter:
Hey everyone,

I'm currently trying to use StreamWriter and Actix's SyncArbiter to calculate some pretty big images. The problem I run into is that StreamWriter borrows the underlying writer making it hard to use across threads. This would be a lot easier if I could get a StreamWriter that owns the underlying Writer.

HeroicKatora:
Tagging this as `hard` as it potentially requires revising the data model if we want to avoid a breaking release. Since most of the internals are private we might be able to cheat our way and enable an owned writer that itself with the lifetime parameter `'static`.

HeroicKatora:
As a general concept, we could take inspiration from `Cow::Owned`. Since it owns the object, the lifetime parameter can be chosen arbitrarily as it is only used in the borrowed state. We could introduce a similar enum internally, only requiring some matching on use which should be negligible overhead compared to the IO.

hank-der-hafenarbeiter:
Going a route similar to the standard library might be a good idea: Vec::iter() borrows the vector Vec::into_iter() owns the vector. So there would - from my perspective - be no need to change any of the existing code. Add a 'into_stream_writer()' method to Writer and create a OwnedStreamWriter. I don't know how much work this would be, but it seems to avoid some of your concerns.

PR #190, opened 2020-03-30
Add an owning stream buffer

HeroicKatora:
To avoid duplication or breaking the interface the internal
representation is an enum that may either hold a reference with lifetime
or owns the buffer with `'static` lifetime. This allows adding a new
interface that should be as good as a truly owning the writer, down to
the lifetime and even `Send` and `Sync` bounds (as contrary to Cow there
is no shared reference involved).

Closes: #189 (maybe it wasn't so hard afterall)

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -112,6 +112,8 @@ pub struct Writer<W: Write> {
     info: Info,
 }
 
+const DEFAULT_BUFFER_LENGTH: usize = 4 * 1024;
+
 impl<W: Write> Writer<W> {
     fn new(w: W, info: Info) -> Writer<W> {
         Writer { w, info }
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -112,6 +112,8 @@ pub struct Writer<W: Write> {
     info: Info,
 }
 
+const DEFAULT_BUFFER_LENGTH: usize = 4 * 1024;
```
nabijaczleweli: NB: stdlib [recommends](https://github.com/rust-lang/rust/blob/1.42.0/src/libstd/sys_common/io.rs#L1) 8k

HeroicKatora: Moving that to its own issue #

HeroicKatora: It's an interesting point though, I wonder where the constant came from.

PR #194, opened 2020-03-31
Test on a big-endian platform; use stdlib funxions for BE conversions

nabijaczleweli:
The bespoke `read_be()` functions could probably be replaced with `byteorder`.

Opened this to get Travis to run for a BE platform, WIP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ matrix:
   include:
     - rust: stable
       env: FLAGS="--no-default-features"
+    - rust: stable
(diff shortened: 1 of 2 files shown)


HeroicKatora:
Testing on some BE platform is a good way to ensure compatibility. It should nevertheless be said that the `byteorder` crate does little different than the implementation here and targets and even older version of Rust than `png`. In fact, it even points out that since `1.32` there are alternatives available within the standard library.

nabijaczleweli:
Indeed! Now uses `to_be_bytes()` and friends from stdlib

PR #195, opened 2020-03-31
Test and build on s390x

HeroicKatora:
Supported by Travis Beta, this does no cross build but actually runs on
big endian hardware including executing our tests. The platform is
listed as Tier 2, same as the mips to which previously cross compiled.


HeroicKatora:
Looks like it is beta for a reason.

PR #198, opened 2020-04-12
Filter speedups

HeroicKatora:
Speedup filters and buffering of raw scanlines. This improves performance between **2%** and **30%**. Best reviewed commit-by-commit.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -59,61 +59,124 @@ pub fn unfilter(
     assert!(bpp > 0);
     let len = current.len();
 
+    fn require_length(slice: &[u8], length: usize) -> Result<&[u8], &'static str> {
+        match slice.get(..length) {
+            None => Err("Filtering failed: not enough data in previous row"),
+            Some(slice) => Ok(slice),
+        }
+    }
+
     match filter {
         NoFilter => Ok(()),
         Sub => {
+            let current = &mut current[..len];
             for i in bpp..len {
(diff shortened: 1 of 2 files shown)


Shnatsel:
On the attached image I see a speedup in filtering of roughly 10%. It also sees a dramatic speedup on miniz_oxide branch. In fact, on miniz_oxide branch `png::filter::unfilter` takes up roughly 70% of decoding time, so any further speedups in filtering would be very impactful.

src/filter.rs
```diff
@@ -59,61 +59,123 @@ pub fn unfilter(
...
-                    current[i] = current[i].wrapping_add(filter_paeth(0, previous[i], 0));
-                }
+            let current = &mut current[..len];
+            let previous = require_length(previous, len)?;
+            if bpp > len {
+                return Err("Filtering failed: bytes per pixel is greater than length of row");
+            }
+
+            for i in 0..bpp {
+                current[i] = current[i].wrapping_add(filter_paeth(0, previous[i], 0));
```
Shnatsel, line 154: I recall reading that LLVM will never unroll loops which use the iteration variable in the body. A crate with macros to work around that has been created: https://docs.rs/crunchy/0.2.2/crunchy/

I wonder if this is relevant here?

Shnatsel:
Ah nevermind, my comment refers to Paeth filter which is barely touched in this PR. That's also what the attached image uses most of the time, so no wonder it still spends a lot of time in unfiltering.

PR #207, opened 2020-05-18
Prevent encoding images with invalid bit depth/color type combination

InvalidUsernameException:
The PNG standard specifies certain combinations of pixel depth and color type as invalid, [see here](https://www.w3.org/TR/PNG/#11IHDR). Previously the encoder would accept any combination of these two values and produce image files that cannot be read by other tools:
```rust
use std::path::Path;
use std::fs::File;
use std::io::BufWriter;

fn main() {
    let path = Path::new(r"test.png");
    let file = File::create(path).unwrap();
    let w = BufWriter::new(file);

    let mut encoder = png::Encoder::new(w, 6, 1);
    encoder.set_color(png::ColorType::RGBA);
    encoder.set_depth(png::BitDepth::Four);

    // red, green, blue, black, white, transparent
    let data = [0xf0, 0x0f, 0x0f, 0x0f, 0x00, 0xff, 0x00, 0x0f, 0xff, 0xff, 0x00, 0x00];
    writer.write_image_data(&data).unwrap();
}
```

```
$ identify  test.png 
identify: Invalid color type/bit depth combination in IHDR `test.png' @ warning/png.c/MagickPNGWarningHandler/1748.
identify: Invalid IHDR data `test.png' @ error/png.c/MagickPNGErrorHandler/1715.
```

Instead it will now produce an error.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -128,19 +128,11 @@ impl<W: Write> Writer<W> {
             return Err(EncodingError::Format("Zero height not allowed".into()));
         }
 
-        let bit_depth = self.info.bit_depth;
-        let color_type = self.info.color_type;
-
-        if (bit_depth == BitDepth::One || bit_depth == BitDepth::Two || bit_depth == BitDepth::Four)
-            && (color_type == ColorType::RGB
-                || color_type == ColorType::GrayscaleAlpha
-                || color_type == ColorType::RGBA)
-            || bit_depth == BitDepth::Sixteen && color_type == ColorType::Indexed
-        {
+        if self.is_bit_depth_color_type_combination_invalid() {
             return Err(EncodingError::Format(
                 format!(
                     "Invalid combination of bit-depth '{:?}' and color-type '{:?}'",
-                    bit_depth, color_type
+                    self.info.bit_depth, self.info.color_type
                 )
                 .into(),
             ));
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -128,6 +128,24 @@ impl<W: Write> Writer<W> {
...
         }
 
+        let bit_depth = self.info.bit_depth;
+        let color_type = self.info.color_type;
+
+        if (bit_depth == BitDepth::One || bit_depth == BitDepth::Two || bit_depth == BitDepth::Four)
+            && (color_type == ColorType::RGB
+                || color_type == ColorType::GrayscaleAlpha
+                || color_type == ColorType::RGBA)
+            || bit_depth == BitDepth::Sixteen && color_type == ColorType::Indexed
```
HeroicKatora, line 138: I believe this can be written more clearly and separated into small helper function. And be it just in case we want to offer disabling the check.

HeroicKatora: And a comment referring to the relevant section would be nice (via number, not link) to not confused readers why certina combinations are 'arbitrarily' disallowed.

InvalidUsernameException: Moved to a helper function now. Good point on the comment.

HeroicKatora:
Thank you!

Issue #215, opened 2020-06-02
Panic when reading multiple images


HeroicKatora:
Calling `next_image` after the main IDAT based image has been read can sometimes panic.

Note that this could only happen when reading multiple frames, which is not fully supported yet in any case. What happens is that `Decoded::finished_decoding` expects to be within some active chunk and that it has to read until its end. However, when the main frame was finished and its iterator is consumed the library will mistakenly read no image data. Since the reading loop itself is what consumes the chunk start, the (successfully reached) call to `finished_decoding` will then hit this `Decoded::ChunkStart` that would have otherwise been consumed while iterating over the rows of the image.

A similar scenario could occur when reading `fcTL`/`fdAT` chunks, of an empty image or an empty subframe. While those are not permited by spec, there was no pressure to reject them until now. But these again would not read any scanline before the above mentioned call. We can avoid this by keeping track of subframes and reading at least `Decoded::ChunkStart`, similar to how `init` reads until over the start of the `IDAT` chunk. This would be beneficial in any case as it at the same time keep exact track of subframes.

Issue #214, opened 2020-06-02
Decoder parses, but doesn't consume, some buffer data


HeroicKatora:
After parsing the image data of a chunk the decoder will seek forward in the stream until it is fully decoded. For this it will fill its read buffer. But critically it fails to inform the underlying buffered reader of the consumed data such that it might later be confused for other events. 

Note that this could only happen when reading multiple frames, which is not fully supported yet in any case. In some other scenarios of corrupt images the error message might differ.

PR #216, opened 2020-06-02
Fix chunk end iteration and track APNG frames

HeroicKatora:
Closes: #214 
Closes: #215


HeroicKatora:
I've fuzzed after this and no results for ~5 hours of fuzzing while the faults introduced were found after 20 minutes each. Should be fine.

Issue #222, opened 2020-06-15
out-of-bounds panic in utils::expand_pass


Shnatsel:
`thread '<unnamed>' panicked at 'index out of bounds: the len is 56 but the index is 56', png-0.16.5/src/utils.rs:248:17`

Sample file triggering the panic: [mysterious-oob-crash](https://user-images.githubusercontent.com/291257/84704783-a4802380-af5a-11ea-97ec-4a1d447f9ae4.png)

The coolest thing about this is that this bug is not reproducible with png's current cargo-fuzz fuzzing harness for decoding. I get the panic only with the following code:

```rust
#![no_main]
use libfuzzer_sys::fuzz_target;

use std::io;
extern crate png;

fn decode_png(data: &[u8])
    -> io::Result<Vec<u8>>
{
    let limits = png::Limits { bytes: 1 << 16 };
    let decoder = png::Decoder::new_with_limits(data, limits);
    let (info, mut reader) = decoder.read_info()?;

    if info.buffer_size() > 5_000_000 {
        return Err(std::io::Error::new(std::io::ErrorKind::Other, "memory limit exceeded"));
    }

    let mut data = vec![0u8; info.buffer_size()];
    reader.next_frame(&mut data)?;

    Ok(data)
}

fuzz_target!(|data: &[u8]| {
    decode_png(data);
});

```

Backtrace:

```
thread '<unnamed>' panicked at 'index out of bounds: the len is 56 but the index is 56', /home/shnatsel/.cargo/registry/src/github.com-1ecc6299db9ec823/png-0.16.5/src/utils.rs:248:17
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1069
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1504
[...]

PR #223, opened 2020-06-16
Fix decoding invalid color combinations

HeroicKatora:
The calculation for buffer size used a different path than the color
used during actual decoding, especially when the flags would expand
colors. This would only panic during the decoding of interlaced images.
For standard images the remaining data would simply get dropped.

Closes: #222

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -49,6 +49,16 @@ impl ColorType {
             }
             + 1 // filter method
     }
+
+    pub(crate) fn is_combination_invalid(self, bit_depth: BitDepth) -> bool {
+        // Section 11.2.2 of the PNG standard disallows several combinations
+        // of bit depth and color type
+        ((bit_depth == BitDepth::One || bit_depth == BitDepth::Two || bit_depth == BitDepth::Four)
+            && (self == ColorType::RGB
+                || self == ColorType::GrayscaleAlpha
+                || self == ColorType::RGBA))
+            || (bit_depth == BitDepth::Sixteen && self == ColorType::Indexed)
+    }
 }
 
 /// Bit depth of the png file
--- a/src/common.rs
+++ b/src/common.rs
@@ -37,6 +37,18 @@ impl ColorType {
             _ => None,
         }
     }
+
+    pub(crate) fn raw_row_length_from_width(&self, depth: BitDepth, width: u32) -> usize {
+        let bits = width as usize * self.samples() * depth as usize;
+
+        let extra = bits % 8;
+        bits / 8
+            + match extra {
+                0 => 0,
+                _ => 1,
+            }
+            + 1 // filter method
+    }
 }
 
 /// Bit depth of the png file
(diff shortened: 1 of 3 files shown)


HeroicKatora:
That apparently fixed it but I'm having a hard time figuring out exactly which combination of inputs would differ, and which one is intended or correct..

src/common.rs
```diff
@@ -37,6 +37,18 @@ impl ColorType {
...
+
+    pub(crate) fn raw_row_length_from_width(&self, depth: BitDepth, width: u32) -> usize {
+        let bits = width as usize * self.samples() * depth as usize;
+
+        let extra = bits % 8;
+        bits / 8
+            + match extra {
+                0 => 0,
+                _ => 1,
+            }
```
fintelia, line 49: This is a convoluted way to say `(bits + 7) / 8`

src/common.rs
```diff
@@ -37,6 +37,18 @@ impl ColorType {
             _ => None,
         }
     }
+
+    pub(crate) fn raw_row_length_from_width(&self, depth: BitDepth, width: u32) -> usize {
+        let bits = width as usize * self.samples() * depth as usize;
```
fintelia, line 42: Does this need to use `u64` instead of `usize` to prevent possible overflow on 32-bit platforms?

src/common.rs
HeroicKatora: It's the overflow safe way. But actually all of the computation of buffer size disregards pointer size and overflow in any case already.. Eh.

src/common.rs
HeroicKatora: Tbh we should catch this overflow __way__ earlier if anything. Do you have some test cases in mind, since this mainly tries to address the single issue reported with mismatching color/depth between different size calculations and not much else.

fintelia: What I'm worried about would be something like an 8bpp 600M x 1px image gray scale image. That would only take ~1GB so it should be a valid PNG and pass the other checks. However, when we compute the number of _bits_ required to store a row we could get an overflow.

HeroicKatora: We could probably test that if we use the procedure outlined in #224 of ensuring that everything is computed in advance. Then a test image containing only the header and no IDAT chunk would be enough to check for the early decoding error. We need to check early against the limits in any case as well!

fintelia: I'm not sure I follow. We shouldn't be returning any error in this case, the function should just correctly return the number of bytes required to store a single row (which will always be less than usize::MAX). The issue is that intermediate computation needs a few extra bits of precision (35 vs 32). The fix should be as simple as writing some `as u64` casts on this line, and then cast back to a `usize` after doing the division on the next line.

HeroicKatora: Seems like a good idea. Let's fork that in a separate issue, okay?

HeroicKatora:
Actually, the color combination in the failing image (RGBA/2-bit) is [not allowed](http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html) per specification. This makes the whole fix appear under a different light, even our encoder rejects it since a recent commit ;)

PR #226, opened 2020-06-17
Turn bpp into a proper enumeration

HeroicKatora:
The filter/prediction algorithm has only a small set of possible values for the length of byte chunks it operates on. This would previously be passed as a `usize` but still match against the discrete value choices in several different locations with differing methods. We turn it into a proper enum to avoid this issue. 

- [x] Evaluated if this influences optimization. Based on our benchmarks, there was nothing too significant.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -72,6 +72,20 @@ pub enum BitDepth {
     Sixteen = 16,
 }
 
+/// Internal count of bytes per pixel.
+/// This is used for filtering which never uses sub-byte units. This essentially reduces the number
+/// of possible byte chunk lengths to a very small set of values appropriate to be defined as an
+/// enum.
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum BytesPerPixel {
+    One,
+    Two,
+    Three,
+    Four,
+    Six,
+    Eight,
+}
+
 impl BitDepth {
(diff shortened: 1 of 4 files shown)


src/common.rs
```diff
@@ -343,6 +374,19 @@ impl Info {
     }
 }
 
+impl Bpp {
+    pub(crate) fn into_usize(self) -> usize {
```
fintelia, line 378: If I remember correctly, can't you just do `self as usize`?

fintelia:
BPP often means bits per pixel, but here seems to stand for bytes per pixel. It might be better to be explicit and call the type BytesPerPixel?

src/common.rs
HeroicKatora: Yes and no. `self as u8` works as well for any enum but isn't always what you meant and will silently wrap. This is just slightly more explicit and affirms that the value range is surely right for the cast.

HeroicKatora:
> call the type BytesPerPixel?

Good point, yes.

Shnatsel:
You should be able to instead use an [enum with custom discriminant values](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations). It will allow for slightly better optimization and perhaps less boilerplate.

HeroicKatora:
@Shnatsel It will be part of #227

Issue #225, opened 2020-06-16
Overflow in byte length calculation


HeroicKatora:
The byte length of rows is calculated through the _bit_length. This requires 3 bits more than the final value which means an intermediate of the calculation might overflow even if the result could be represented on a 32-bit target.

https://github.com/image-rs/image-png/blob/5c00d5af103a3ac12ed288bc2d3afac2667ebfc2/src/common.rs#L318-L334

Previous conversation on the issue: https://github.com/image-rs/image-png/pull/223#discussion_r441019014

PR #227, opened 2020-06-19
Check buffer sizes for overflow

HeroicKatora:
Fixes an issue affecting mainly platforms with 32-bit (or smaller)
pointer size where the calculation of required buffer sizes would
use unchecked math ops that could silently overflow.

We now try to calculate the maximum of the necessary lengths up-front,
as part of converting to a Reader, which ensures that all other
computations have no overflow.

Closes: #225

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -17,8 +17,12 @@ pub enum ColorType {
 impl ColorType {
     /// Returns the number of samples used per pixel of `ColorType`
     pub fn samples(&self) -> usize {
+        self.samples_u8().into()
+    }
+
+    pub(crate) fn samples_u8(self) -> u8 {
         use self::ColorType::*;
-        match *self {
+        match self {
             Grayscale | Indexed => 1,
             RGB => 3,
             GrayscaleAlpha => 2,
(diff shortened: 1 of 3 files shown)


src/common.rs
```diff
@@ -38,16 +38,36 @@ impl ColorType {
         }
     }
 
-    pub(crate) fn raw_row_length_from_width(&self, depth: BitDepth, width: u32) -> usize {
-        let bits = width as usize * self.samples() * depth as usize;
+    pub(crate) fn checked_raw_row_length(&self, depth: BitDepth, width: u32) -> Option<usize> {
```
fintelia, line 41: This whole function could just be:
```rust
   TryFrom::try_from((width as u64 * (self.samples() * depth as usize) as u64 + 7) / 8 + 1).ok()
```
LLVM is even able to produce [quite reasonable code](https://rust.godbolt.org/z/bWrmjd) for the on 32-bit targets, and there aren't issues with a hypothetical 128-bit target either because _width_ is a u32 so the full product can't overflow.

HeroicKatora: Good point, introducing internal converters to `u8` (we can discuss making them public later) makes this even clearer. Then this multiplication is straightup `u32*u8*u8` which surely fits in `u64`.

Issue #233, opened 2020-06-27
Docs state that reader.next_frame() can only be called once


Shnatsel:
The code example on docs.rs for 0.16.5 release states:

```
// Read the next frame. Currently this function should only called once.
```

Now that this crate supports APNG, that's no longer the case. The documentation should be updated.

PR #234, opened 2020-06-27
Ergonomics

HeroicKatora:
Closes: #233

### DIFF
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -1,0 +1,10 @@
+
+# Fuzzing `i686` on native `x86_64` Linux
+
+It's possible and quite straightforward to fuzz 32-bit architecture on 64-bit
+Linux systems with the following two steps:
+
+```bash
+$ rustup target add i686-unknown-linux-gnu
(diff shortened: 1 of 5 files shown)


fuzz/README.md
```diff
@@ -0,0 +1,10 @@
+
+# Fuzzing i686 on native `x86_64` linux
```
nabijaczleweli, line 2: The codification of i686 vs `x86_64` is inconsistent here (also linux might be better off as Linux)

HeroicKatora: It's consistent with the names in `rustup target list`. I'll change the capital of linux, good point.

nabijaczleweli: I mostly meant the code blocks around `x86_64` and lack thereof around i686

HeroicKatora: Ohhhh. That makes sense then xd

HeroicKatora: Just checked that Github's markdown flavor can render inline code in titles so I'll add it to both. I vaguely remember that jekyll (the Github pages app) could not.

HeroicKatora:
@nabijaczleweli Thanks for the review.

PR #236, opened 2020-07-03
Add setter to encoder for tRNS chunk

cosmicchipsocket:
Accompanies #136 and #138. Without the tRNS chunk, indexed files cannot be saved with transparency. This adds a simpler setter to Encoder so that this chunk can be specified.


cosmicchipsocket:
Hi, would it be at all possible to make a new point release with this enhancement? I am working on a library that depends on this feature.

Issue #213, opened 2020-06-01
Intended breakage in `0.17`


HeroicKatora:
A list of all the interface incompatible changes we could clean up in the next major version:

- [ ] `Decoder::read_info` returns an overly condensed form of `Info`
- [ ] `Reader::next_row` should have named types
- [ ] `Reader::next_interlaced_row` should use wrapper struct instead of tuple
- [ ] `DecodingError::Format` is far to unspecific, could use a boxed error interface with better than stringly typing of all publicly inspectable error variants
- [ ] `ChunkType` could be a wrapper type, not a raw array. This would allow customized debug and display implementations to properly present the usually ascii nature of its contents. Also the free standing functions of `chunk` could then be methods.
- [ ] `ColorType::samples` should take `self` by value.
- [ ] `FrameControl::delay_den` could be a proper non-zero type. Secondary.
- [ ] `Adam7Iterator::Item` should be stictly typed with a struct.
- [ ] Test interlaced APNG
- [ ]

yanchith:
Hey there! Sad to see you drop `#![no_std]` from the list. Anything I can do to help? What is the major complication there?

HeroicKatora:
The problem with `#[no_std]` is usage of `io::Error`. It is difficult to have interface that interact with both streams _and_ are compatible with no-std usage, relying only on in memory buffers. It would be slightly easier had the crate been engineered from the ground up for this but retro-fitting it proves difficult. (I _think_ I've found a reasonably good interface in `weezl` which I can explain a bit further—in a blog post probably—but it's not applicable to the style of `png`). There is hopefully more opportunity in the next major version.

HeroicKatora:
To expand on the problematic design (if you want to call it that, it's certainly not problematic in blocking code): All decoding methods get handed a _borrowed_ output buffer. The code within is then written in such a way that it assumes it can keep this borrow until either a) the image has been read b) an error occurs. Both `async` and no-std subvert this expectation because we return to the caller to fetch addition input data, which gives up our current borrow of the buffer. (Note: technically we could make _everything_ a coroutine and thus avoid the return for `async` specifically but this is then very unwieldy for all synchronous usage). 

To fix this we need to make the layers between the ergonomic interface and the underlying, sans-io `StreamDecoder` statemachine aware of potentially exiting and storing some progress information for resumption. Removing this assumption also makes it harder to use the output as a temporary buffer which could otherwise be used to avoid a small number of allocations but I don't think it's used that way right now.

[...]

PR #250, opened 2020-10-03
Expose semantic result types for next_{interlaced_,}row

WorldSEnder:
As mentioned in #213 this replaces the tuples returned from `next_row` and `next_interlaced_row` with types carrying more semantic meaning. Should probably also replace `<Adam7Iterator as Iterator>::Item` in the same commit, but lemme know.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -68,22 +68,30 @@ pub struct Decoder<R: Read> {
 }
 
 /// A row of data with interlace information attached.
-#[derive(Debug)]
-#[non_exhaustive]
+#[derive(Clone, Copy, Debug)]
 pub struct InterlacedRow<'data> {
-    pub data: &'data [u8],
-    pub interlace: InterlaceInfo,
+    data: &'data [u8],
+    interlace: InterlaceInfo,
+}
+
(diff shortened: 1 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -67,16 +67,41 @@ pub struct Decoder<R: Read> {
...
+pub struct InterlacedRow<'data> {
+    pub data: &'data [u8],
+    pub interlace: InterlaceInfo,
 }
 
-enum InterlaceInfo {
-    None,
+/// PNG (2003) specifies two interlace modes, but reserves future extensions.
+#[derive(Debug)]
+#[non_exhaustive]
```
HeroicKatora, line 80: In the tradeoff between marking it `Copy` and `#[non_exhaustive]`, I'd choose to have `Copy` etc. derives. The probability that the standard introduces a new interlace method before we bump major version is quite small.

HeroicKatora (changes requested):

HeroicKatora:
Of course, if structs are not exhaustive then we need to decide if the members should be public or not. It's only slightly more ergonomic to access members and the caller is not expected to ever construct those structs.

HeroicKatora (approved):

PR #251, opened 2020-10-03
Encapsulated error

HeroicKatora:
Reworks the error type such that

* we can add new variants easily
* the top-level variants represent different blame of failure, similar to the `image` error.
* the creation of an error no longer allocates

Addresses part of: #213

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -531,12 +531,12 @@ pub struct ParameterError {
 
 #[derive(Debug)]
 pub(crate) enum ParameterErrorKind {
-    /// A provided buffer must be large enough to hold the image data. Where the buffer can be
-    /// allocated by the caller, it must ensure that it has a minimum size as hinted previously.
+    /// A provided buffer must be have the exact size to hold the image data. Where the buffer can
+    /// be allocated by the caller, they must ensure that it has a minimum size as hinted previously.
     /// Even though the size is calculated from image data, this does counts as a parameter error
     /// because they must react to a value produced by this library, which can have been subjected
     /// to limits.
-    ImageBufferSize(usize, usize),
+    ImageBufferSize { expected: usize, actual: usize },
     /// A bit like return `None` from an iterator.
     /// We use it to differentiate between failing to seek to the next image in a sequence and the
     /// absence of a next image. This is an error of the caller because they should have checked
@@ -556,8 +556,8 @@ impl fmt::Display for ParameterError {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         use ParameterErrorKind::*;
         match self.inner {
-            ImageBufferSize(expected, got) => {
-                write!(fmt, "wrong data size, expected {} got {}", expected, got)
+            ImageBufferSize { expected, actual } => {
+                write!(fmt, "wrong data size, expected {} got {}", expected, actual)
             }
             PolledAfterEndOfImage => write!(fmt, "End of image has been reached"),
         }
--- a/src/common.rs
+++ b/src/common.rs
@@ -531,8 +531,19 @@ pub struct ParameterError {
 
 #[derive(Debug)]
 pub(crate) enum ParameterErrorKind {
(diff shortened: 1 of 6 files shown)


src/decoder/stream.rs
```diff
@@ -68,9 +68,21 @@ pub enum Decoded {
     ImageEnd,
 }
 
+/// Any kind of error during PNG decoding.
+///
+/// This enumeration provides a very rough analysis on the origin of the failure. That is, each
+/// variant corresponds to one kind of actor causing the error. It should be understood as a direct
```
kaj, line 74: Should this be "It *shouldn't* be ..." ?

kaj (commented):
Documentation is nice!  I've got one question:

src/decoder/stream.rs
HeroicKatora: That is correct.

src/encoder.rs
```diff
@@ -20,7 +20,23 @@ pub type Result<T> = result::Result<T, EncodingError>;
 #[derive(Debug)]
 pub enum EncodingError {
     IoError(io::Error),
-    Format(Cow<'static, str>),
+    Format(FormatError),
```
HeroicKatora, line 23: I'm only a bit puzzled by the asymmetry of variants between `DecodingError` and `EncodingError`. I'm not sure if we should introduce at least `ParameterError` here—and keep it unused for now. Maybe a method to manually close the png (i.e. write and IEND chunk) by `&mut self` could be useful to avoid fallible writes in Drop. After it, all other methods should return that error I think.

src/decoder/stream.rs
```diff
@@ -99,7 +98,7 @@ pub enum DecodingError {
     /// number of calls
     ///
     /// If you're an application you might want to signal that a bug report is appreciated.
-    Parameter(Cow<'static, str>),
+    Parameter(ParameterError),
```
fintelia, line 101: Would it be better to call this DecodingError::InvalidParameter?

HeroicKatora: I was trying to imitate the naming in `image`, mainly. I'm also not convinced the `invalid` qualifier adds anything substantial, it's an error so of course something wasn't right. The [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order) recommend verb-object-error as a word order but there is no verb to apply here.

fintelia: You're probably right. Part of my confusion initially was from the variants of ParameterError: "Wrong data size" or "end of image has been reached" sound more like file corruption than something caused by passing a bad parameter.

HeroicKatora: I can improve those of course :+1:

src/common.rs
```diff
@@ -524,6 +524,46 @@ bitflags! {
...
+}
+
+#[derive(Debug)]
+pub(crate) enum ParameterErrorKind {
+    /// A provided buffer must be large enough to hold the image data. Where the buffer can be
+    /// allocated by the caller, it must ensure that it has a minimum size as hinted previously.
+    /// Even though the size is calculated from image data, this does counts as a parameter error
+    /// because they must react to a value produced by this library, which can have been subjected
+    /// to limits.
+    ImageBufferSize(usize, usize),
```
fintelia, line 539: How about named fields here:
```rust
ImageBufferSize { expected: usize, provided: usize },
```

src/common.rs
```diff
@@ -524,6 +524,46 @@ bitflags! {
...
 }
 
+#[derive(Debug)]
+pub struct ParameterError {
+    inner: ParameterErrorKind,
+}
+
+#[derive(Debug)]
+pub(crate) enum ParameterErrorKind {
+    /// A provided buffer must be large enough to hold the image data. Where the buffer can be
```
fintelia, line 534: This error can be triggered by a buffer that is either too large or too small, so we should probably start the comment with a sentence indicating that.

fintelia (approved):

PR #260, opened 2020-11-29
Add support for the sRGB chunk.

de-vri-es:
This PR adds support for the `sRGB` chunk.

A new type `SrgbRenderingIntent` is added, and `Info` now has a `srgb: Option<SrgbRenderingIntent>` field.

If the decoder encounters an `sRGB` chunk, it sets the `srgb` field along with the `source_gamma` and `source_chromaticities` field (using predefined values). Any future `gAMA` or `cHRM` chunks are ignored (and the data from previous ones is overwritten).

The encoder adds the `sRGB` chunk along with the pre-defined `gAMA` and `cHRM` chunks if the `srgb` field is set. Manually defined `source_gamma` and `source_chromaticities` are ignored in this case.

I believe this conforms to the specification: https://www.w3.org/TR/2003/REC-PNG-20031110/#11sRGB

/edit: CI failure looks like the same Windows/rustup issue, unrelated to the PR contents :(

### DIFF
--- a/src/chunk.rs
+++ b/src/chunk.rs
@@ -31,6 +31,8 @@ pub const pHYs: ChunkType = ChunkType([b'p', b'H', b'Y', b's']);
 pub const cHRM: ChunkType = ChunkType([b'c', b'H', b'R', b'M']);
 /// Source system's gamma value
 pub const gAMA: ChunkType = ChunkType([b'g', b'A', b'M', b'A']);
+/// sRGB color space chunk
+pub const sRGB: ChunkType = ChunkType([b's', b'R', b'G', b'B']);
 
 // -- Extension chunks --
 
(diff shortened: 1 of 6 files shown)


de-vri-es:
Note: Maybe `SrgbRenderingIntent` should be renamed to `IccRenderingIntent` or just `RenderingIntent`, as it is more general than the use for the sRGB chunk. It doesn't appear in other places in the PNG specification though.

HeroicKatora:
CI failure was addressed in #261.

de-vri-es:
Rebased, CI now passes :)

HeroicKatora (approved):

HeroicKatora:
Thanks!

de-vri-es:
Thanks for the review :)

PR #263, opened 2020-12-07
Fix the Avg filter for encoding

okaneco:
wrapping_add produces incorrect filter results for Average. The sum of the left pixel and above pixel should not overflow and wrap around. This change mimics the implementation used in `unfilter` for Avg in `filter.rs` where the byte values are converted to u16 for addition.

> The subtraction of the predicted value from the raw byte must be done modulo 256, so that both the inputs and outputs fit into bytes. However, the sum Raw(x-bpp)+Prior(x) must be formed without overflow (using at least nine-bit arithmetic). floor() indicates that the result of the division is rounded to the next lower integer if fractional; in other words, it is an integer division or right shift operation.

6.5. Filter type 3: Average  
https://www.w3.org/TR/PNG-Filters.html

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -198,8 +198,9 @@ pub(crate) fn filter(method: FilterType, bpp: BytesPerPixel, previous: &[u8], cu
         }
         Avg => {
             for i in (bpp..len).rev() {
-                current[i] =
-                    current[i].wrapping_sub(current[i - bpp].wrapping_add(previous[i]) / 2);
+                current[i] = current[i].wrapping_sub(
+                    ((u16::from(current[i - bpp]) + u16::from(previous[i])) / 2) as u8,
+                );
             }
 
             for i in 0..bpp {
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -269,4 +269,46 @@ mod test {
             }
         }
     }
+
+    #[test]
+    fn roundtrip_ascending_previous_line() {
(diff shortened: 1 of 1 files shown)


okaneco:
Current behavior

light-3

With this patch

light-fix-3

src/filter.rs
```diff
@@ -198,8 +198,9 @@ pub(crate) fn filter(method: FilterType, bpp: BytesPerPixel, previous: &[u8], cu
         }
         Avg => {
             for i in (bpp..len).rev() {
-                current[i] =
-                    current[i].wrapping_sub(current[i - bpp].wrapping_add(previous[i]) / 2);
+                current[i] = current[i].wrapping_sub(
+                    ((u16::from(current[i - bpp]) + u16::from(previous[i])) / 2) as u8,
+                );
```
HeroicKatora, line 203: There is a `roundtrip` test some lines down. It seems to have missed this due to using very uniform input data which does not trigger the overflow bug. Could you add a second set of inputs that triger it, i.e. linear ascending sequence or something like that?

HeroicKatora:
We should backport this to `v0.16`. It seems small and annoying enough to warrant another point-release.

src/filter.rs
okaneco: I've added a test which the current master fails on.  I can combine it into one test if you think that's better.

HeroicKatora (approved):

PR #266, opened 2020-12-08
Release 0.16.8

HeroicKatora:
A bug fix for one non-default filter type since `0.17` will likely still take a bit of time (aiming for around new year).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## 0.16.8
+
+* Fixed a bug that led to data corruption when encoding with Average filter.
+
 ## 0.16.7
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,7 @@
+## 0.16.8
+
+* Fixed a bug that lead to data corruption when encoding with Average filter.
```
fintelia, line 3: nit: lead -> led

fintelia (approved):

CHANGES.md
HeroicKatora: Thanks. I swear, I'm not doing that on purpose to check if you're reading them :wink:

PR #274, opened 2020-12-30
Improve documentation

ISSOtm:
[The documentation currently online](https://docs.rs/png/0.16.8/png/index.html) seems to largely assume familiarity with the PNG spec. Since I have minimal knowledge of the spec, this has given me trouble. I need a fairly low-level interface, since I am trying to port to Rust a program sensitive to whether a PNG is paletted or not.

This PR is intended to improve documentation, including to be more noob-friendly (as much as possible, anyway); I have made a few modifications based off my experience, and noted other possible changes below. If you have ideas, please tell me, and I will add them!

- [ ] A code comment mentions that some `ColorType`/`BitDepth` combinations are impossible. Maybe this should be pointed out in the docs?
- [ ] The C code block in `Transformations`'s doc does not have type for its constants, and it's not clear (to the reader) whether this code comes from this crate, or `libpng`. Further, what are they intended for?
- [ ] Maybe some more examples? [This is how I'm using the lib](https://github.com/ISSOtm/rsgbgfx/blob/52b9ccbc88f30a91f86085143212b701f0cca342/src/img/png.rs#L15), fwiw. How to use the various fields of `Info`, for example, I had to determine through trial and error.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1,19 +1,24 @@
 //! Common types shared between the encoder and decoder
 use std::{convert::TryFrom, fmt};
 
-/// Describes the layout of samples in a pixel
+/// Describes how a pixel is encoded.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(u8)]
 pub enum ColorType {
+    /// 1 grayscale sample.
     Grayscale = 0,
+    /// 1 red sample, 1 green sample, 1 blue sample.
     Rgb = 2,
+    /// 1 sample for the palette index.
     Indexed = 3,
+    /// 1 grayscale sample, then 1 alpha sample.
     GrayscaleAlpha = 4,
+    /// 1 red sample, 1 green sample, 1 blue sample, and finally, 1 alpha sample.
     Rgba = 6,
 }
 
 impl ColorType {
-    /// Returns the number of samples used per pixel of `ColorType`
+    /// Returns the number of samples used per pixel encoded in this way.
     pub fn samples(self) -> usize {
         self.samples_u8().into()
     }
@@ -71,7 +76,8 @@ impl ColorType {
     }
 }
 
-/// Bit depth of the PNG file
+/// Bit depth of the PNG file.
+/// Specifies the number of bits per sample.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(u8)]
 pub enum BitDepth {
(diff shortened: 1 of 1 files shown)


src/common.rs
```diff
@@ -474,12 +483,12 @@ impl Info {
...
     }
 
-    /// Returns the bits per pixel
+    /// Returns the number of bits per pixel.
     pub fn bits_per_pixel(&self) -> usize {
         self.color_type.samples() * self.bit_depth as usize
     }
 
-    /// Returns the bytes per pixel
+    /// Returns the number of bytes per pixel.
```
fintelia, line 491: This should probably say that the number is rounded up to the nearest byte. Though to be honest I'm not quite sure why we have this function at all. When would you actually want to do rounding when computing pixel size?

fintelia:
>  The C code block in Transformations's doc does not have type for its constants, and it's not clear (to the reader) whether this code comes from this crate, or libpng. Further, what are they intended for?

Those are the transformations from libpng which aren't yet supported by this crate, but which may be in the future. The syntax is actually for the bitflags macro (from the crate of the same name).

src/common.rs
HeroicKatora: The bytes per channel are rounded up. For 4bpp all channels are still in individual bytes.

HeroicKatora (approved):

HeroicKatora:
In general I do not think the library should go too far into the png specification. The documentation should be written with the right terminology and in a style that ensures one can quickly find the relevant portions and paragraphs but replicating it in different wording is, in my opinion, not a good approach. For example, the allowed color combinations are due to the specification and while maybe being slightly more obvious to duplicate the table from the specification, it also gives the appearance that the library does something that diverges from the specification or does not support all combinations. That can be highly confusing as well.

> Maybe some more examples? This is how I'm using the lib, fwiw. How to use the various fields of Info, for example, I had to determine through trial and error.

Sure, that seems good. Most of the methods lack dedicated examples so short snippets would be appreciated.

HeroicKatora:
Thanks! Feel free to open more PRs with additional improvements, as pointed out in your open questions.

ISSOtm:
Oh, I thought I'd add more documentation to this PR before merging it. :P

To avoid repeating the spec, and encourage users to read it, a solution may be to pepper the documentation with links to relevant spec paragraphs? Could also be useful as citation-like reference, again to provide guidance to the user. (For example, I thought paletted images couldn't have alpha, until someone pointed out the `tRNS` chunk to me.)

HeroicKatora:
Yes, link and section numbers would also be helpful.

PR #277, opened 2021-01-16
Add encode methods to common structs

Rimpampa:
Add `encode` methods to the structs that contain information present in PNG chunks.

Originated from [PR#255](https://github.com/image-rs/image-png/pull/255).


src/common.rs
```diff
@@ -343,6 +369,10 @@ impl ScaledFloat {
     pub fn into_value(self) -> f32 {
         Self::reverse(self.0) as f32
     }
+
+    pub fn encode<W: Write>(self, w: &mut W) -> encoder::Result<()> {
+        encoder::write_chunk(w, chunk::gAMA, &self.into_scaled().to_be_bytes())
+    }
```
HeroicKatora, line 375: `ScaledFloat` is used in different chunks, exposing `encode` such that it encodes it as a gama chunk could be confusing.
```suggestion
    pub(crate) fn encode_gama<W: Write>(self, w: &mut W) -> encoder::Result<()> {
        encoder::write_chunk(w, chunk::gAMA, &self.into_scaled().to_be_bytes())
    }
```

HeroicKatora (changes requested):
I just have one little nit.

src/encoder.rs
```diff
@@ -253,67 +253,11 @@ impl<W: Write> Writer<W> {
...
-            if let Some(g) = &self.info.source_gamma {
-                write_chunk(&mut self.w, chunk::gAMA, &g.into_scaled().to_be_bytes())?;
-            }
-
-            if let Some(c) = &self.info.source_chromaticities {
-                let enc = Self::chromaticities_to_be_bytes(&c);
-                write_chunk(&mut self.w, chunk::cHRM, &enc)?;
-            }
-        }
+        self.info.encode(&mut self.w)?;
```
HeroicKatora, line 256: Seems like you've accidentally dropped writing the `png` signature itself :joy:

Rimpampa: Yeah just noticed😅

HeroicKatora (approved):

PR #278, opened 2021-01-22
encoder: Fix off-by-one errors in ChunkWriter

### LINKED ISSUE #262: libpng warning: IDAT: Extra compressed data
I seem to be getting this warning on finishing writing a png with a `StreamWriter`:
```
libpng warning: IDAT: Extra compressed data
```
I can't seem to find any information about whether it is harmless or not. Doing the exact same operation, but not using a `StreamWriter`, does not produce a warning. What might this imply?

If it is truly harmless and unavoidable, it would be nice to have an option to suppress the warning.

scurest:
Fixes #262.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -439,7 +439,7 @@ impl<'a, W: Write> Write for ChunkWriter<'a, W> {
         let written = buf.read(&mut self.buffer[self.index..])?;
         self.index += written;
 
-        if self.index + 1 >= self.buffer.len() {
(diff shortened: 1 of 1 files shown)


HeroicKatora (commented):
How complicated would it be to generate a regression test for this, that we can add to the repository?

scurest:
I imagine you'd know better than me.

HeroicKatora (approved):
Never mind the question then, we can't currently detect the extra bytes when using our own decoder. So it's quite difficult.

PR #284, opened 2021-02-12
Add APNG Encoding

Rimpampa:
This is the new pull request, I went back with only `Writer` and `StreamWriter` and currently only the first one is able to encode APNG animations. I did some tests and it's working fine (didn't test frames smaller than the image).

Now for the `StreamEncoder` I was thinking that it could be done in two ways:
1) the `finish` method returns the `ChunkOutput` so that by reusing the `Writer` it's possible to encode multiple frames
2) Modify how the `ChunkWriter` works and make it store an entire chunk before flushing, and after flushing allow to change the `FrameControl`; then on the `StreamWriter` it would be possible to seamlessly encode every frame including the separate default image.

You can have a look at the second approach here [encoder.rs](https://github.com/Rimpampa/image-png/blob/encoder/src/encoder.rs).

The first one just needs to expose the `ChunkOutput` which shouldn't be much of a problem, considering also that switching from `AsMut` to `DerefMut` would make it even more silent. On the other hand implementing the second one would require a lot of changes that might not be what we want, but would make the public interface (imo) a lot nicer. I still don't know if limiting the buffer size to the maximum size of a chunk is a bad thing, where would be needed to bufferize more than 2GB of data? But regardless if we want in the future to implement the `PartialChunkData` thing you were talking about a solution like this will be needed because, as you already pointed out, it's impossible to know the chunk size in advance it follows that it must be the last thing to be written, but because it must preceed the chunk data, all the data must be bufferized before it gets written.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -139,37 +139,27 @@ impl<W: Write> Encoder<W> {
         }
     }
 
-    pub fn new_animated(
-        w: W,
-        width: u32,
-        height: u32,
-        num_frames: u32,
-        num_plays: u32,
-        sep_def_img: bool,
-    ) -> Encoder<W> {
+    pub fn set_animated(&mut self, num_frames: u32, num_plays: u32) {
         let actl = AnimationControl {
             num_frames,
             num_plays,
         };
         let fctl = FrameControl {
             sequence_number: 0,
-            width,
-            height,
+            width: self.info.width,
+            height: self.info.height,
             ..Default::default()
         };
-        let info = Info {
-            width,
-            height,
-            animation_control: Some(actl),
-            frame_control: Some(fctl),
-            ..Default::default()
-        };
-        Encoder {
-            w,
-            info,
-            filter: FilterType::default(),
-            adaptive_filter: AdaptiveFilterType::default(),
-            sep_def_img,
+        self.info.animation_control = Some(actl);
+        self.info.frame_control = Some(fctl);
+    }
+
+    pub fn set_sep_def_img(&mut self, sep_def_img: bool) -> Result<()> {
+        if self.info.animation_control.is_none() {
+            self.sep_def_img = sep_def_img;
+            Ok(())
+        } else {
+            Err(EncodingError::Format(FormatErrorKind::NotAnimated.into()))
         }
     }
 
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -439,7 +439,7 @@ impl<W: Write> Writer<W> {
                 let mut alldata = vec![0u8; 4 + MAX_fdAT_CHUNK_LEN as usize];
                 for chunk in zlib_encoded.chunks(MAX_fdAT_CHUNK_LEN as usize) {
                     alldata[..4].copy_from_slice(&fctl.sequence_number.to_be_bytes());
-                    alldata[4..].copy_from_slice(chunk);
+                    alldata[4..][..chunk.len()].copy_from_slice(chunk);
                     write_chunk(&mut self.w, chunk::fdAT, &alldata[..4 + chunk.len()])?;
                     fctl.sequence_number += 1;
                 }
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -426,7 +426,8 @@ impl<W: Write> Writer<W> {
                     self.write_chunk(chunk::IDAT, &chunk)?;
                 }
             } else {
-                let mut alldata = vec![0u8; 4 + MAX_fdAT_CHUNK_LEN as usize];
+                let buff_size = zlib_encoded.len().min(MAX_fdAT_CHUNK_LEN as usize);
+                let mut alldata = vec![0u8; 4 + buff_size];
                 for chunk in zlib_encoded.chunks(MAX_fdAT_CHUNK_LEN as usize) {
                     alldata[..4].copy_from_slice(&fctl.sequence_number.to_be_bytes());
                     alldata[4..][..chunk.len()].copy_from_slice(chunk);
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -365,8 +365,8 @@ impl<W: Write> Writer<W> {
 
     /// Writes the image data.
     pub fn write_image_data(&mut self, data: &[u8]) -> Result<()> {
-        const MAX_IDAT_CHUNK_LEN: u32 = (1u32 << 31) - 1;
-        const MAX_fdAT_CHUNK_LEN: u32 = u32::MAX - 4;
+        const MAX_IDAT_CHUNK_LEN: u32 = std::u32::MAX >> 1;
+        const MAX_fdAT_CHUNK_LEN: u32 = (std::u32::MAX >> 1) - 4;
 
         if self.info.color_type == ColorType::Indexed && self.info.palette.is_none() {
             return Err(EncodingError::Format(FormatErrorKind::NoPalette.into()));
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -289,7 +289,7 @@ pub struct Writer<W: Write> {
     filter: FilterType,
     adaptive_filter: AdaptiveFilterType,
     sep_def_img: bool,
-    written: u32,
+    written: u64,
 }
 
 const DEFAULT_BUFFER_LENGTH: usize = 4 * 1024;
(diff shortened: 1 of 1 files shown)


HeroicKatora (commented):
I'll take an in-depth look this weekend, my first impression is that it looks quite clean :)

src/encoder.rs
```diff
@@ -199,6 +262,37 @@ impl<W: Write> Encoder<W> {
     pub fn set_adaptive_filter(&mut self, adaptive_filter: AdaptiveFilterType) {
         self.adaptive_filter = adaptive_filter;
     }
+
+    /// Sets the dispose operation for each frame.
```
fintelia, line 266: Copy paste error?

Rimpampa: Yup, but there is also a some documentation missing so I won't worry much about it right now. Once we agree on an implementation I would fill in those gaps

src/encoder.rs
```diff
@@ -105,18 +123,56 @@ pub struct Encoder<W: Write> {
...
+    }
+
+    pub fn new_animated(
+        w: W,
+        width: u32,
+        height: u32,
+        num_frames: u32,
+        num_plays: u32,
+        sep_def_img: bool,
+    ) -> Encoder<W> {
```
HeroicKatora, line 152: This doesn't need to happen in the constructor, configuring `num_frames` etc. can be done with methods on `Encoder` similar to setting up color, compression, and default filters. I think this is a lot clearer than having 4 parameters with the same type, and a bool.

src/encoder.rs
```diff
@@ -297,12 +422,137 @@ impl<W: Write> Writer<W> {
...
+
+            if self.written == 0 {
+                for chunk in zlib_encoded.chunks(MAX_IDAT_CHUNK_LEN as usize) {
+                    self.write_chunk(chunk::IDAT, &chunk)?;
+                }
+            } else {
+                let mut alldata = vec![0u8; 4 + MAX_fdAT_CHUNK_LEN as usize];
+                for chunk in zlib_encoded.chunks(MAX_fdAT_CHUNK_LEN as usize) {
+                    alldata[..4].copy_from_slice(&fctl.sequence_number.to_be_bytes());
+                    alldata[4..][..chunk.len()].copy_from_slice(chunk);
```
HeroicKatora, line 442: That's a huge allocation even when it is not necessary. Selecting the minimum of the zlib_encoded's length and MAX_fdAT_CHUNK_LEN should do the trick and avoid this.

src/encoder.rs
```diff
@@ -297,12 +422,137 @@ impl<W: Write> Writer<W> {
...
+            fctl.height = height;
+            Ok(())
+        } else {
+            Err(EncodingError::Format(FormatErrorKind::NotAnimated.into()))
+        }
+    }
+
+    pub fn set_frame_position(&mut self, x: u32, y: u32) -> Result<()> {
+        if let Some(ref mut fctl) = self.info.frame_control {
+            if fctl.width + x > self.info.width || fctl.height + y > self.info.height {
```
HeroicKatora, line 505: See other comment about overflow.

HeroicKatora (changes requested):
That is far more approachable in implementation and use. There are a few implementation nits here and there but nice work overall!

src/encoder.rs
Rimpampa: Yup that's something I forgot to do, seems like I got confused by it at some point and added [..chunk.len()] where I should have done like you said

src/encoder.rs
```diff
@@ -297,12 +422,137 @@ impl<W: Write> Writer<W> {
...
+            fctl.delay_num = delay_num;
+            Ok(())
+        } else {
+            Err(EncodingError::Format(FormatErrorKind::NotAnimated.into()))
+        }
+    }
+
+    pub fn set_frame_dimesion(&mut self, width: u32, height: u32) -> Result<()> {
+        if let Some(ref mut fctl) = self.info.frame_control {
+            if width + fctl.x_offset > self.info.width || height + fctl.y_offset > self.info.height
```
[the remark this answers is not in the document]
Rimpampa: Ok, I didn't know `Ord for Option<T: Ord>` was a thing

src/encoder.rs
```diff
@@ -265,18 +368,34 @@ impl<W: Write> Writer<W> {
...
         }
 
-        let bpp = self.info.bpp_in_prediction();
-        let in_len = self.info.raw_row_length() - 1;
-        let prev = vec![0; in_len];
-        let mut prev = prev.as_slice();
-        let mut current = vec![0; in_len];
-        let data_size = in_len * self.info.height as usize;
+        let max = match self.info.animation_control {
+            Some(actl) if self.sep_def_img => actl.num_frames + 1,
```
Rimpampa, line 379: This could also lead to an overflow right?

Also considering that the sequence number and the number of frames are both 32 bits unsigned integers shouldn't it be impossible to encode an animation with more than u32::MAX / 2 frames? There is always at least a pair of chunks (fcTL and fdAT) for each frame so for each frame the sequence number is incremented at least twice

HeroicKatora: You're right, it could. If anything, simply catch it here by returning an error. However, having a look at the [reference patch for libpng](https://sourceforge.net/projects/libpng-apng/files/libpng16/1.6.37/libpng-1.6.37-apng.patch.gz/download)  it seems that it simply overflows by using C's unsigned integer arithmetic—so kind of like a TCP sequence number. That would also make sense.

Rimpampa: I think that returning an error would be better as the spcification states that the sequence numbers are unique and strictly sequencial (no gaps)

HeroicKatora: Okay, then add a check and error here if you want. Kind of awkward that the supposed C implementation doesn't actually implement the specification and I'm slightly skeptical that the mitigation of chunk reording warrants such a restriction but let's just be as spec compliant as possible.

src/encoder.rs
Rimpampa: Regarding this problem I think we could use the same `Limits` struct used for decoding, but this might be the subject of another PR

src/encoder.rs
Rimpampa: I decided to use wrapping_add as with the C implementation as there isn't really a good place to put the check, if you think it matters it could be solved in another PR, but we should think first at the `StreamWriter` implementation I think.
For the number of frames written I used an u64 so there is no need to check for overflows

HeroicKatora (commented):
You wanted to do a documentation pass once the implementation was done. I believe we are at that stage.

Rimpampa:
Nice, I'll start working on that as soon as possible. By the way, have you taken a look at the possible `StreamWriter` implementation? At this point I think it would be part of another PR so in the meanwhile I could block `into_stream_writer` for animation `Writer`s if you agree

src/encoder.rs
```diff
@@ -297,12 +412,141 @@ impl<W: Write> Writer<W> {
...
+        if let Some(ref mut fctl) = self.info.frame_control {
+            fctl.delay_den = delay_den;
+            fctl.delay_num = delay_num;
+            Ok(())
+        } else {
+            Err(EncodingError::Format(FormatErrorKind::NotAnimated.into()))
+        }
+    }
+
+    pub fn set_frame_dimesion(&mut self, width: u32, height: u32) -> Result<()> {
```
foresterre, line 480: possibly a type: `set_frame_dimesion` -> `set_frame_dimension`

src/encoder.rs
```diff
@@ -297,12 +412,141 @@ impl<W: Write> Writer<W> {
...
+            }
+            fctl.x_offset = x;
+            fctl.y_offset = y;
+            Ok(())
+        } else {
+            Err(EncodingError::Format(FormatErrorKind::NotAnimated.into()))
+        }
+    }
+
+    pub fn reset_frame_dimesion(&mut self) -> Result<()> {
```
foresterre, line 510: possibly a typo: `reset_frame_dimesion` -> `reset_frame_dimension`

HeroicKatora:
> By the way, have you taken a look at the possible StreamWriter implementation? At this point I think it would be part of another PR so in the meanwhile I could block into_stream_writer for animation Writers if you agree.

I do agree, it's fine if the stream is full after a single frame and adding the ability to write multiple frames with it might best be part of a separate PR.

Rimpampa:
That should be it for the documentation, most of it I've taken from the specification to avoid any mistakes.

BTW one thing I want to point out that I feel like I've not explained correctly is that the `StreamWriter` currently
is going to corrupt the png stream if used on an animation because it doesn't know it has to write fcTL and fdAT chunks.

HeroicKatora (approved):
LGTM. I can live with ChunkWriter/StreamWriter always writing IDAT for now. I'm not sure if it can even be considered 'corrupting' the data stream, at least not more than before. Correct me if I misunderstood but atm it will always write a separate default image though but also not increment the current frame number, which seems fine.

Rimpampa:
Yes you are right, it writes a separate default image. The corruption that I was referring to comes when you write some frames using the `Writer` and then call `(into_)stream_writer`.

PR #285, opened 2021-02-18
Add support for parsing iCCP chunk.

jrmuizel:


src/decoder/stream.rs
```diff
@@ -911,6 +912,43 @@ impl StreamingDecoder {
...
+                }
+            }
+
+            let mut profile = Vec::new();
+            let mut inflater = ZlibStream::new();
+            inflater.decompress(buf, &mut profile)?;
+            inflater.finish_compressed_chunks(&mut profile)?;
+
+            self.info.as_mut().unwrap().icc_profile = Some(profile);
+            Ok(Decoded::Nothing)
```
fintelia, line 948: Is there a way we can impose a maximum size for the ICC profile to prevent a malicious file from causing us to allocate an unreasonable amount of memory here?

HeroicKatora: PR would look good to me with this resolved.

jrmuizel: How is this problem handle elsewhere? Is there a way to cap the size that inflater will make the Vec?

jrmuizel: libpng uses PNG_USER_CHUNK_MALLOC_MAX which defaults to 8000000

HeroicKatora: The maximum should be capped at most at `Limits::bytes` from the field in `Decoder`. (It defaults to 64Mb but that's not a constant, it can be configured before decoding). Maybe it should be less because it's supposed to be somewhat of a total maximum.

jrmuizel: What about doing the actual capping? It seems like finish_compressed_chunks() will happily keep allocating as much as the the zlib stream wants.

HeroicKatora: As stated in the comment there, yes, we may want to fix that at some point. But that's less critical than the main decompression. Open a separate PR for it, if you can fix it?

HeroicKatora: Well, yes, that means you can't control the final size if everything is decoded in a single call to `decompress`. That's suboptimal indeed. To clarify, you can however call `decompress` with an empty slice so that each call appends at most `CHUNCK_BUFFER_SIZE` bytes of data to `profile`. Repeat until you're reasonably sure that there is no remaining input data and then call finish.. That's not optimal but should be quite good for a start?

HeroicKatora (approved):
Good enough for me, I'll make a note that the finish-phase of `ZlibDecoder` needs some way to check for bounds on the output vector size. We can merge with the one formatting nit fixed.

jrmuizel:
Done.

Issue #283, opened 2021-01-31
Build fails if `png-encoding` feature is disabled.


kettle11:
Here's the error:

```
/src/common.rs:2:20
  |
2 | use crate::{chunk, encoder};
  |                    ^^^^^^^
  |                    |
  |                    no `encoder` in the root
  |                    help: a similar name exists in the module: `decoder`
```

Issue #242, opened 2020-08-02
Encoding should use miniz_oxide as well


astei:
#218 brought some welcome performance improvements by changing decoding to use miniz_oxide. I think it'd be good if encoding were also done with miniz_oxide as well, to both improve performance and decrease binary size (important to me as I deploy this in a WebAssembly module).

scurest:
I made [a branch](https://github.com/image-rs/image-png/compare/master...scurest:miniz-deflate)'s that replaces deflate with miniz_oxide to see what effect this would have.

I only tested a little. Initial observations are:

* small compile time improvement
* no significant binary size improvement
* encode times appear to be comparable. It's somewhat difficult to compare since it depends on the compression settings. If I tweak the compression level to make my output PNG filesize close to master, the encode times are close too; miniz_oxide might actually be a little slower. (But that also might be because I've used it in a sub-optimal way....)

Craig-Macomber:
I can't see @scurest 's branch, but I made a similar one, this time replacing deflate with flate2 (which defaults to using miniz_oxide).

https://github.com/Craig-Macomber/image-png/tree/flate2 (I didn't know how to handle the huffman and rle compression levels, but otherwise the change was trivial)

This lead me to observe that flate2 depends on libc, even when using miniz_oxide, which is tracked by https://github.com/rust-lang/flate2-rs/issues/274 . I'll take a look at that.

I think if that gets fixed, my changes might be a win as far as number of dependencies and build time goes. Until then, maybe its progress toward taking more actively maintained and popular dependencies (flate2 seems to be well supported, and in this case mostly used miniz_oxide which we already depend on).

An interesting approach might be to direct all access to miniz_oxide through flate2, which would allow the user to select a compression back-end via flate2's features.

Craig-Macomber:
flate2 no longer depends on libc making using it to reduce dependencies here actually make sense. I updated my branch accordingly.

fintelia:
@Craig-Macomber could you make a PR with your changes?

Craig-Macomber:
> @Craig-Macomber could you make a PR with your changes?

Done. #350 350

yisibl:
Hi, after I upgraded to 0.17.6, the size of the generated PNG file is much bigger, is there any way to fix it?

[...]

PR #287, opened 2021-02-19
Make encoder a core feature

HeroicKatora:
Removes the `png-encoding` feature, always turning the functionality on.

Another goal here is to reduce the interface. In particular there was
already one new major release of `deflate` that we couldn't track without 
one of our own. The loss is that one can no longer precisely configure the 
level of compression which might be remediated with a new setter but the 
rough options should address most use cases. A semantic option (i.e. a
limit to time spent encoding) would be more actionable for a user anyways.

Since we no longer publicly depend on `deflate` this also prepares doing 
#242 later in a minor version bump.

Closes: #283

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,6 +10,7 @@ jobs:
       matrix:
         rust: ["1.34.2", stable, beta, nightly]
         os: [ubuntu-latest, windows-latest, macos-latest]
+        features: ["", "png-encoding"]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


CryZe:
Doesn't this unnecessarily increase compile times for everyone not intending to encode PNGs, which I'd say it's the most common scenario?

HeroicKatora:
While that technically true, it's not by a lot. It's also odd to have them split as a feature. Where there is a large enough difference in the size of implementation a separate crate is far more common. Also, at least in `png`, we can't fully separate the two. It would also complicate future ideas of adding stream-like editing, which the `png` specification was originally meant to allow for.

fintelia:
This all seems reasonable to me

PR #288, opened 2021-02-25
Accept Cow<'a, [u8]> for PLTE and tRNS data

AlexTMjugador:
Currently, a user of this crate, when encoding a PNG file with custom PLTE or tRNS data, is forced to pass that data as an owned Vec<u8>. However, the internals of this library do not really need an owned vector, just something that can be converted to a byte slice, and the requirement of a Vec may introduce unneeded copies and allocations in client code.

Fix the situation by accepting a `Cow<'a, [u8]>` instead, which will allow the library to clone the data to get ownership of it if needs it in the future.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -833,10 +833,7 @@ impl SubframeInfo {
     }
 }
 
-fn expand_paletted<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(
-    buffer: &mut [u8],
-    info: &Info<P, T, I>,
-) -> Result<(), DecodingError> {
+fn expand_paletted<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(buffer: &mut [u8], info: &Info<P, T, I>) -> Result<(), DecodingError> {
     if let Some(palette) = info.palette.as_ref() {
         if let BitDepth::Sixteen = info.bit_depth {
             // This should have been caught earlier but let's check again. Can't hurt.
@@ -852,8 +849,7 @@ fn expand_paletted<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(
             if let Some(ref trns) = info.trns {
                 utils::unpack_bits(buffer, 4, info.bit_depth as u8, |i, chunk| {
                     let (rgb, a) = (
-                        palette
-                            .as_ref()
+                        palette.as_ref()
                             .get(3 * i as usize..3 * i as usize + 3)
                             .unwrap_or(&black),
                         trns.as_ref().get(i as usize).unwrap_or(&0xFF),
@@ -865,8 +861,7 @@ fn expand_paletted<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(
                 });
             } else {
                 utils::unpack_bits(buffer, 3, info.bit_depth as u8, |i, chunk| {
-                    let rgb = palette
-                        .as_ref()
+                    let rgb = palette.as_ref()
                         .get(3 * i as usize..3 * i as usize + 3)
                         .unwrap_or(&black);
                     chunk[0] = rgb[0];
@@ -883,10 +878,7 @@ fn expand_paletted<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(
     }
 }
 
-fn expand_gray_u8<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(
-    buffer: &mut [u8],
-    info: &Info<P, T, I>,
-) {
+fn expand_gray_u8<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(buffer: &mut [u8], info: &Info<P, T, I>) {
     let rescale = true;
     let scaling_factor = if rescale {
         (255) / ((1u16 << info.bit_depth as u8) - 1) as u8
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1,8 +1,8 @@
 mod stream;
 mod zlib;
 
-use self::stream::{FormatErrorInner, CHUNCK_BUFFER_SIZE};
 pub use self::stream::{Decoded, DecodingError, StreamingDecoder};
+use self::stream::{FormatErrorInner, CHUNCK_BUFFER_SIZE};
 
 use std::io::{BufRead, BufReader, Read, Write};
 use std::mem;
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -833,7 +833,10 @@ impl SubframeInfo {
     }
 }
 
-fn expand_paletted<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(buffer: &mut [u8], info: &Info<P, T, I>) -> Result<(), DecodingError> {
+fn expand_paletted<P: AsRef<[u8]>, T: AsRef<[u8]>, I: AsRef<[u8]>>(
+    buffer: &mut [u8],
+    info: &Info<P, T, I>,
+) -> Result<(), DecodingError> {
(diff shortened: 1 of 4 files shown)


HeroicKatora (changes requested):
I'm quite uncomfortable with the amount of type parameters that this introduces. It adds quite a bit of code bloat from monomorphization in the worst case, and requires them to be specified explicitly. Would changing the fields' types to `Cow<'lt, [u8]>` work as well? This would merely introduce _one_ lifetime parameter that would default to `'static` (mostly equivalent to the current type) in many cases.

AlexTMjugador:
Thanks for the review! I also share your view that this is kind of uncomfortable, as evidenced by how the tests fail to compile now because of the required type parameters. I'll try your suggestion of using `Cow` and see how it works ASAP.

HeroicKatora:
And if this also turns out to still be quite the churn, then the fallback idea after that would be to allow writing these chunks separately, after having turned the IHDR into a `Writer`. This way, any extra type of lifetime parameter added would be local to a single method at a time—or it might be provided by a separate method.

But let's see if we can milk the `Cow<_>` idea first.

AlexTMjugador:
I've just implemented the suggested `Cow<'a, [u8]>` approach, which is indeed more ergonomic. However, there still is a matter that I think that is important to discuss.

https://github.com/image-rs/image-png/blob/7ed5169f56d3224a84f01a0309cddbef25844903/src/encoder.rs#L583-L586

In this test I've changed the original `encoder.stream_writer()` call to `encoder.into_stream_writer()`. This consumes the `Writer`, and although the caller is not able to use it anymore, this has no side effects for this test. However, the whole purpose of the original non-consuming call is to allow appending additional chunks after writing the image data, which sounds like a legitimate use case. Now, consider the new signature of that method:

https://github.com/image-rs/image-png/blob/7ed5169f56d3224a84f01a0309cddbef25844903/src/encoder.rs#L306-L316

Because it now takes `&'a mut self`, which is needed for the compiler to accept the code, the returned `StreamWriter` must outlive the `Writer`, which generates funky compiler errors when using the `stream_writer` method in that test case. I've tried changing scopes and lifetime parameters to no avail, and I feel that something may be fundamentally wrong with my code, but my google-fu has failed and I don't really know how I should tackle this.

Other than that, these changes:

1. Saved a clone in a test case: https://github.com/image-rs/image-png/blob/948e526b987f8a7ec025db9d88db6a047b56bdc9/src/encoder.rs#L622-L628
2. Removed this function, that was used internally even though it was not really necessary, and isn't mentioned in the docs: https://github.com/image-rs/image-png/blob/32c9d2c2230ce0cd0f0d1f26b76057a7fb94cc49/src/decoder/stream.rs#L1057-L1060

Which I think that are nice improvements.

src/decoder/mod.rs
```diff
@@ -203,13 +203,13 @@ impl<R: Read> Decoder<R> {
     }
 }
 
-struct ReadDecoder<R: Read> {
+struct ReadDecoder<'a, R: Read> {
     reader: BufReader<R>,
-    decoder: StreamingDecoder,
+    decoder: StreamingDecoder<'a>,
     at_eof: bool,
 }
```
HeroicKatora, line 210: The decoder doesn't need to have a lifetime parameter. It will _always_ allocate the chunks itself, i.e. the decoder always produces `Cow::Owned()`. Then the `Info` struct within it should use `'static` and not expose the parameter further.

src/decoder/stream.rs
```diff
@@ -321,13 +321,13 @@ impl From<DecodingError> for io::Error {
...
 /// PNG StreamingDecoder (low-level interface)
-pub struct StreamingDecoder {
+pub struct StreamingDecoder<'a> {
     state: Option<State>,
     current_chunk: ChunkState,
     /// The inflater state handling consecutive `IDAT` and `fdAT` chunks.
     inflater: ZlibStream,
     /// The complete image info read from all prior chunks.
-    info: Option<Info>,
+    pub(crate) info: Option<Info<'a>>,
```
HeroicKatora, line 330: Referring to this field here.

src/encoder.rs
```diff
@@ -199,9 +199,9 @@ impl<W: Write> Encoder<W> {
 }
 
 /// PNG writer
-pub struct Writer<W: Write> {
+pub struct Writer<'a, W: Write> {
     w: W,
-    info: Info,
+    info: Info<'a>,
```
HeroicKatora, line 204: Are the relevant fields actually used, from here one forward? If not then it would also be possible to remove the lifetime parameter here (thus addressing the concern of `StreamWriter`). We'd need to rebuild a new `Info<'static>` from the non-borrowed parts of the struct built in `Encoder`, and drop all borrowed parts.

src/decoder/mod.rs
AlexTMjugador: Indeed, that simplifies the signatures a bit. Thanks.

src/encoder.rs
AlexTMjugador: The `Writer` has a `init()` method (used by the `write_header` implementation of `Encoder`) that calls `encode` in `Info<'a>`. In turn, `encode` writes both PLTE and tRNS chunks if present. I'll think more about how can I tackle the `StreamWriter` concern with this information, but I'd say that `Writer` indirectly uses those fields.

AlexTMjugador:
Your review made me realize that I can lend the `Writer` struct another struct with only a subset of the `Info` parameters, and then pass on a reference to the real `Info` struct (which contains the actual PLTE and tRNS data) during the `init` call. This way the problem with lifetimes on Writer is entirely avoided, and in my opinion the resulting code is much more readable. As expected, this also fixes the problem with `encoder.stream_writer()` not working, so I reverted the affected test to its original version. Thanks!

HeroicKatora (commented):
Codewise this all seems clean, I'd like to have one look at benchmarks. I suspect that the optimizer is smart enought to avoid any overhead on the `to_info()` cast—or that they appear seldom enough so that it doesn't matter. But hard data is better than a suspicion.

AlexTMjugador:
I've just benchmarked the current master (commit `08fe119`, which had the role of the "old" version while benchmarking) against the latest commit of this PR (commit `47506e8`, the "new" version), on the same Debian Linux desktop PC with the same background tasks open. [This text file](https://github.com/image-rs/image-png/files/6065244/Commands.txt) shows the output of `uname -a` and `rustc -V` (I'm using Rust nightly fe1bf8e05 2021-02-23, 1.52.0) commands, and the relevant output of the `cargo bench` command I used a single time for each commit to test the performance (first on the "old" version, then on the "new" version). Criterion generated [this HTML report](https://github.com/image-rs/image-png/files/6065241/Criterion.benchmark.zip) with the results.

Based on this data, I'm inclined to conclude that these changes introduce extremely reduced performance regressions (in the order of less than 1%, tens of ms, on a 2011 Intel Core i3-2100 CPU) that, as explained in the Criterion report, are well within error margin and not likely to impact any real world usage, even on the worst case scenario.

HeroicKatora:
Thanks a lot!

Shnatsel:
I appreciate the thought of leaving the API open-ended enough to allow cloning in the future. However, I think this could be accomplished in the current code even without Cow, with a plain slice, and remove some of the complexity that way. Specifically my concern is that Cow is somewhat unergonomic.

Since the `image` API is not yet at 1.0, I am tempted to try the plain slice and see if that runs into any issues; this would improve API ergonomics a bit, and we could still backtrack on this if we need to.

AlexTMjugador:
I think that using a `Cow` internally makes sense, as its whole purpose is being open-ended enough to allow cloning if necessary, like you stated. Using a slice instead may be simpler now, but if the need to clone the data arises in the future, then we'll have to reimplement `Cow` functionality with methods like `copy_from_slice`, or use a `Cow` then, which does not seem so ergonomic.

However, I agree that offering an API that hides this away from the users of the library may be a good idea. I think that providing two variants of the methods to set the PLTE and tRNS data, one that takes borrowed data and other that takes owned data, would be pretty ergonomic for users, and under the hood we can simply instantiate a `Cow` struct with that as appropriate.

PR #290, opened 2021-03-07
Streaming Encoder for APNG

Rimpampa:
I already introduced this on [my last PR](https://github.com/image-rs/image-png/pull/284). I've added some comments on the code to explain what I did.

For some reasons I thought that my implementation of the `ChunkWriter` differed a lot from the current one, but in reality they both hold at most one full chunk, and in fact the old one didn't even check if the buffer size was greater than the maximum size of a chunk (even if it wouldn't matter that much as extremely rare cases need to bufferize 2GiB).

The way it works right now is that you write data and it automatically separates each frame, and all the parameters (selected with the `set_*` methods) of the animation are kept until the current frame is being written and applied on the following ones.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1332,8 +1332,10 @@ impl<'a, W: Write> Write for StreamWriter<'a, W> {
             return Ok(0);
         }
 
-        if matches!(self.writer, Wrapper::Chunk(_)) {
-            self.new_frame()?
+        match self.writer {
+            Wrapper::Chunk(_) => self.new_frame()?,
+            Wrapper::Zlib(_) => {}
+            Wrapper::None => unreachable!(),
         }
 
         if self.to_write == 0 {
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1312,15 +1312,15 @@ impl<'a, W: Write> StreamWriter<'a, W> {
             _ => unreachable!(),
         };
         wrt.flush()?;
-        wrt.writer.written += 1;
         if let Some(fctl) = self.fctl {
             wrt.set_fctl(fctl);
         }
-        wrt.write_header()?;
         let (scansize, size) = wrt.next_frame_info();
-        self.end = wrt.writer.written == wrt.writer.max_frames();
         self.line_len = scansize;
         self.to_write = size;
+        wrt.writer.written += 1;
+        wrt.write_header()?;
+        self.end = wrt.writer.written + 1 == wrt.writer.max_frames();
         self.writer = Wrapper::Zlib(ZlibEncoder::new(wrt, self.compression.to_options()));
         Ok(())
     }
@@ -1332,18 +1332,15 @@ impl<'a, W: Write> Write for StreamWriter<'a, W> {
             return Ok(0);
         }
 
-        match self.writer {
-            Wrapper::Chunk(_) => self.new_frame()?,
-            Wrapper::Zlib(_) => {}
-            Wrapper::None => unreachable!(),
-        }
-
-        if self.end {
-            let err = FormatErrorKind::EndReached.into();
-            return Err(EncodingError::Format(err).into());
+        if matches!(self.writer, Wrapper::Chunk(_)) {
+            self.new_frame()?
         }
 
         if self.to_write == 0 {
+            if self.end {
+                let err = FormatErrorKind::EndReached.into();
+                return Err(EncodingError::Format(err).into());
+            }
             self.writer = match self.writer.take() {
                 Wrapper::Zlib(wrt) => Wrapper::Chunk(wrt.finish()?),
                 _ => unreachable!("invalid writer state"),
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -492,6 +492,7 @@ impl<W: Write> Writer<W> {
     /// Writes the image data.
     pub fn write_image_data(&mut self, data: &[u8]) -> Result<()> {
         const MAX_IDAT_CHUNK_LEN: u32 = std::u32::MAX >> 1;
+        #[allow(non_upper_case_globals)]
         const MAX_fdAT_CHUNK_LEN: u32 = (std::u32::MAX >> 1) - 4;
 
         if self.info.color_type == ColorType::Indexed && !self.info.has_palette {
@@ -828,8 +829,6 @@ impl<'a, W: Write> DerefMut for ChunkOutput<'a, W> {
     }
 }
 
-/// < INTERNAL DOCUMENTATION >
-///
 /// This writer is used between the actual writer and the
 /// ZlibEncoder and has the job of packaging the compressed
 /// data into a PNG chunk, based on the image metadata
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -807,140 +808,580 @@ impl<W: Write> Drop for Writer<W> {
...
-impl<W: Write> Write for ChunkWriter<'_, W> {
-    fn write(&mut self, mut buf: &[u8]) -> io::Result<usize> {
-        let written = buf.read(&mut self.buffer[self.index..])?;
-        self.index += written;
+    /// NOTE: this bypasses the internal buffer so the flush method should be called before this
+    ///       in the case there is some data left in the buffer when this is called, it will panic
+    fn write_header(&mut self) -> Result<()> {
+        if self.index != 0 {
+            // return Err(todo!())
+            panic!("Called when not flushed") // TODO: add error code
```
HeroicKatora, line 912: TODO?

src/encoder.rs
```diff
@@ -807,140 +808,580 @@ impl<W: Write> Drop for Writer<W> {
...
+                self.curr_chunk = chunk::IDAT;
+            } else {
+                self.curr_chunk = chunk::fdAT;
+            }
+            if let Some(ref mut fctl) = wrt.info.frame_control {
+                fctl.encode(&mut wrt.w)?;
+                fctl.sequence_number += 1;
+            } else {
+                unreachable!()
+            }
```
HeroicKatora, line 929: This would be clarer by matching on `wrt.info.frame_control`.
```suggestion
        match wrt.info.frame_control {
            _ if wrt.sep_def_mg => self.curr_chunk = chunk::IDAT,
            None => self.curr_chunk = chunk::IDAT,
            Some(ref mut fctl) => {
                // …
            }
        }
```

src/encoder.rs
```diff
@@ -807,140 +808,580 @@ impl<W: Write> Drop for Writer<W> {
...
+    fn set_fctl(&mut self, f: FrameControl) {
+        if let Some(ref mut fctl) = self.writer.info.frame_control {
+            // ingnore the sequence number
+            *fctl = FrameControl {
+                sequence_number: fctl.sequence_number,
+                ..f
+            };
+        } else {
+            panic!("This function must be called on an animated PNG")
+        }
```
HeroicKatora, line 947: Rather odd when compared to the other setters that return an error?

src/encoder.rs
```diff
@@ -807,140 +808,580 @@ impl<W: Write> Drop for Writer<W> {
...
+    fn write(&mut self, mut data: &[u8]) -> io::Result<usize> {
+        if data.is_empty() {
+            return Ok(0);
+        }
+
+        // index == 0 means a chunk as been flushed out
+        if self.index == 0 {
+            let wrt = self.writer.deref_mut();
+            // ??? maybe use self.curr_chunk == chunk::fdAT ???
+            if !wrt.sep_def_img && wrt.info.frame_control.is_some() && wrt.written > 0 {
```
HeroicKatora, line 975: Or add a field `header_written: bool`?

src/encoder.rs
```diff
@@ -807,140 +808,580 @@ impl<W: Write> Drop for Writer<W> {
...
         Ok(())
     }
+
+    /// Flushes the buffered chunk, checks if it was the last frame,
+    /// writes the next frame header and gets the next frame scanline size
+    /// and image size.
+    fn new_frame(&mut self) -> Result<()> {
+        let mut wrt = match self.writer.take() {
+            Wrapper::Chunk(wrt) => wrt,
+            _ => unreachable!(),
```
HeroicKatora, line 1312: Can you add a short explanation of why this can't be reached? It would appear to me that an error in `Writer::write` for example leaves it in the state `Wrapper::None` but the instance still exists.

src/encoder.rs
```diff
@@ -807,140 +808,580 @@ impl<W: Write> Drop for Writer<W> {
...
+
+        if self.to_write == 0 {
+            if self.end {
+                let err = FormatErrorKind::EndReached.into();
+                return Err(EncodingError::Format(err).into());
+            }
+            self.writer = match self.writer.take() {
+                Wrapper::Zlib(wrt) => Wrapper::Chunk(wrt.finish()?),
+                _ => unreachable!("invalid writer state"),
+            };
```
HeroicKatora, line 1349: So if this errors in `wrt.finish()?`, doesn't it leave `self.writer` in the state `Wrapper::None`. Then I call `write` again and we enter the `unreachable` above.

HeroicKatora (changes requested):

src/encoder.rs
Rimpampa: it was more like a question if an error code should be added, as this code is only called internally and it can fail only if the implementation is wrong

src/encoder.rs
Rimpampa: This is like the other TODO, it's code called only inside the `StreamWriter` so it can fail only if we call it in the wrong situation

src/encoder.rs
Rimpampa: If you think that's better/nicer I will do it

src/encoder.rs
Rimpampa: Yes this should not `take()` but borrow instead, it should `take()` only right before constructing the `ZlibEncoder`

src/encoder.rs
Rimpampa: Yes you are totally right, I overlooked this line, but it seems to me that if this happens there is nothing to do, we wouldn't be able to continue because the `ChunkWriter` would get lost. Maybe I should add a variant to the `Wrapper` enum like `Unrecoverable` that would return an `EncodingError` every time `write` is called after that error. Maybe `None` can be used for that but it might be better to enforce the fact that `None` is used only as a temporary state

src/encoder.rs
HeroicKatora: If the implementation requires this internally then you can use `assert_eq!(self.index, 0)`.

src/encoder.rs
HeroicKatora: I see, then panicking is fine.

src/encoder.rs
```diff
@@ -90,6 +91,7 @@ impl fmt::Display for FormatError {
             EndReached => write!(fmt, "all the frames have been already written"),
             MissingFrames => write!(fmt, "there are still frames to be written"),
             MissingData(n) => write!(fmt, "there are still {} bytes to be written", n),
+            Unrecoverable => write!(fmt, "the writer is in an unrecoverable state"),
```
fintelia, line 94: Isn't the error more that the writer was _used_ after a previous operation put the writer into an unrecoverable state?

HeroicKatora: So something along the lines of:
> a previous error put the writer into an unrecoverable state

src/encoder.rs
Rimpampa: Now that I read this again, I don't understand how it would be implemented that flag, as the `if` over there is not checking if the header is written or not but rather it's checking if the chunk whose header has already been written is an fdAT in order to add the sequence number.

HeroicKatora (approved):
It's good enough for now, want to get ready with the next major release as well

PR #291, opened 2021-03-13
Tweak the Info struct

HeroicKatora:
It is now non-exhaustive which simplifies adding more information in the future. The borrowed option is also available for the icc chunk. The `InfoFilterType`, which has long been replaced for its encoding purpose, is removed.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -127,14 +127,9 @@ pub struct Encoder<'a, W: Write> {
 
 impl<'a, W: Write> Encoder<'a, W> {
     pub fn new(w: W, width: u32, height: u32) -> Encoder<'static, W> {
-        let info = Info {
-            width,
-            height,
-            ..Default::default()
-        };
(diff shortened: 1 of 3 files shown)


src/encoder.rs
```diff
@@ -415,16 +410,15 @@ impl PartialInfo {
...
-        Info {
-            width: self.width,
-            height: self.height,
-            bit_depth: self.bit_depth,
-            color_type: self.color_type,
-            frame_control: self.frame_control,
-            animation_control: self.animation_control,
-            compression: self.compression,
-            ..Default::default()
-        }
```
fintelia, line 427: Was this change actually necessary, or is it just for stylistic reasons?

HeroicKatora: Necessary, as the `..Default::default()` syntax is incompatible with private members.

fintelia (approved):

PR #295, opened 2021-05-06
fix fuzz targets

fintelia:
See also https://github.com/google/oss-fuzz/pull/5724


HeroicKatora (approved):

HeroicKatora:
@fintelia Thanks, I had a hard time focussing on actually doing it. I'm feeling a little burned out on the various image libraries at the moment. Can I ask you to donate slightly more attention next month, I might want to feel comfortable taking a bit of a break.

Issue #298, opened 2021-07-17
Writer needs an .into_inner() function


surferjeff:
I want to write a PNG to memory.  I wrote code like this:

```rs
    let mut encoder = png::Encoder::new(Cursor::new(Vec::new()), info.width, info.height);
    encoder.set_color(png::ColorType::RGB);
    encoder.set_depth(info.bit_depth);
    let mut writer = encoder.write_header()?;
    writer.write_image_data(&frame)?;
    let cursor = writer.into_inner();
    let bytes = cursor.into_inner();
```

I can't get the bytes back out without an `into_inner()` function.  The line  let `cursor = writer.into_inner();` fails to compile because there's no `.into_inner()` function.

fintelia:
Couldn't you do:
```rust
let mut bytes = Vec::new();
{
    let mut encoder = png::Encoder::new(Cursor::new(&mut bytes), info.width, info.height);
    encoder.set_color(png::ColorType::RGB);
    encoder.set_depth(info.bit_depth);
    let mut writer = encoder.write_header()?;
    writer.write_image_data(&frame)?;
}

**Comment by surferjeff** [comment]
Yes, that worked.  Thanks!

**Comment by ravenexp** [comment]
> Couldn't you do:
> 
> ```rust
> let mut bytes = Vec::new();
> {
>     let mut encoder = png::Encoder::new(Cursor::new(&mut bytes), info.width, info.height);
>     encoder.set_color(png::ColorType::RGB);
>     encoder.set_depth(info.bit_depth);
>     let mut writer = encoder.write_header()?;
>     writer.write_image_data(&frame)?;
> }
> ```

This workaround no longer works in `0.17`:

```
error[E0597]: `pngbuf` does not live long enough
  --> src/png.rs:31:52
   |
31 |         let mut encoder = Encoder::new(Cursor::new(&mut pngbuf), self.width, self.height);
   |                                        ------------^^^^^^^^^^^-
   |                                        |           |
   |                                        |           borrowed value does not live long enough
   |                                        argument requires that `pngbuf` is borrowed for `'static`
...
50 |     }
   |     - `pngbuf` dropped here while still borrowed
```

I believe this is because `Encoder::new()` now enforces a `'static` lifetime parameter for some reason:

https://github.com/image-rs/image-png/blob/c24fc93134ce82579595912230f9a3aed33bb1fc/src/encoder.rs#L173-L174

I think this issue should be reopened for version 0.17 and #299 may be revived as well.

[...]

PR #299, opened 2021-07-17
implement into_inner() for Writer

surferjeff:
Fixes https://github.com/image-rs/image-png/issues/298


surferjeff:
This is broken.  Investigating....

surferjeff:
False alarm.  It works as intended.

PR #300, opened 2021-08-14
Bump MSRV to 1.46 and release version 0.17.0

### LINKED ISSUE #272: Publish version including miniz_oxide update
Currently image has duplicate dependencies of both miniz_oxide 0.3 and 0.4 due to the png and the tiff crates. It appears it was updated in 4424245d09dac4be32e9c19690fcfe3cc1fbdfcc but this isn't included in the latest release. Can a new release be made with this commit?

fintelia:
Our bitflags dependency [decided to bump their minimum supported Rust version](https://github.com/bitflags/bitflags/issues/255), so we must do the same.

Fixes #272 (which is long, long overdue)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "png"
-version = "0.16.7"
+version = "0.17.0"
 license = "MIT OR Apache-2.0"
 
 description = "PNG decoding and encoding library in pure Rust"
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -8,7 +8,7 @@ jobs:
   build:
     strategy:
       matrix:
-        rust: ["1.34.2", stable, beta, nightly]
+        rust: ["1.46.0", stable, beta, nightly]
         os: [ubuntu-latest, windows-latest, macos-latest]
         features: [""]
     runs-on: ${{ matrix.os }}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,16 @@
+## 0.17.0
+
+* Increased MSRV to 1.46.0
+* Rework output info usage
+* Implement APNG encoding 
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "png"
-version = "0.16.7"
+version = "0.17.0"
 license = "MIT OR Apache-2.0"
 
 description = "PNG decoding and encoding library in pure Rust"
```
HeroicKatora, line 6: Since `bitflags` seems to preserve the right for bumps in new versions, should we restrict our dependency to `<= 1.4.0`?

HeroicKatora (approved):
Agreed that it's overdue and that the plan detailed in #213 was too extensive. Let's do the rest of those changes (in particular custom chunks) in the next version then.

Cargo.toml
fintelia: I think that would make us incompatible with any future crate that required bitflags 1.4.0+? (because 1.3 and 1.4 are semver compatible and Cargo doesn't allow multiple versions within the same compatibility range)

fintelia: That would be much worse than the current situation where only old compilers are broken, and it can be worked around by putting "=1.2.0" in a downstream Cargo.toml

HeroicKatora: True.. Okay, so we don't do anything but reactively advertise the MSRV. Kind of sucks but you're right, the only good way would be to influence cargo's dependency resolution, not introduce new bounds ourselves.

HeroicKatora:
Follow up to the MSRV discussion: `naga`, the Rust glsl compiler, had the same issue with their own MSRV and resolved it by making it incompatible with `1.3`: https://github.com/gfx-rs/naga/blob/v0.6/Cargo.toml#L18-L19

PR #320, opened 2022-01-26
Update dependencies

Expyron:


Cargo.toml
```diff
@@ -22,17 +22,15 @@ include = [
 [dependencies]
 bitflags = "1.0"
 crc32fast = "1.2.0"
-deflate = "0.9"
-[dependencies.miniz_oxide]
-version = "0.4.1"
-features = ["no_extern_crate_alloc"]
```
fintelia, line 28: What happened to this feature? Do we no longer need it / was it removed?

Expyron: It was changed to a no-op in miniz_oxide 0.4.2, and removed in 0.5.0.

HeroicKatora: It was only introduced for Rust `1.34` compatibilty, on our request actually. Since the major version bump and our bump of the MSRV it should no longer be necessary anyways.

fintelia (approved):

PR #324, opened 2022-02-04
Move color/bit depths combination check

MichaelMcDonnell:
There was a TODO comment in the `parse_ihdr` method to check if the color/bit depths combination is valid. It turned out there was already a check for this in the `read_info` method. It makes more sense for all of the checks to be in one place, so I moved the check into the `parse_ihdr` method.

I also renamed the fields of the `InvalidColorBitDepth` error enum value to match their names in the rest of the code. That way the field names do not need to be named when constructing the `InvalidColorBitDepth` value. The type in not exposed publicly so it is not an API breaking change.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -172,18 +172,6 @@ impl<R: Read> Decoder<R> {
         let mut reader = Reader::new(self.r, StreamingDecoder::new(), self.transform, self.limits);
         reader.init()?;
 
-        let color_type = reader.info().color_type;
-        let bit_depth = reader.info().bit_depth;
-        if color_type.is_combination_invalid(bit_depth) {
-            return Err(DecodingError::Format(
-                FormatErrorInner::InvalidColorBitDepth {
-                    color: color_type,
-                    depth: bit_depth,
-                }
-                .into(),
-            ));
-        }
-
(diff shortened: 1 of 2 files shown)


HeroicKatora (commented):
At some point we should find a way to control the stream state after such an error, or offer recovery. In particular some of the combinations make semantic sense (e.g. 4-bit rgb). In particular for `StreamWriter` it would be super useful if an png-cleaning/repeair tool was able to parse the rest of the stream at a chunk-level even in error cases.

This is also my only concern with performing validation beyond bit interpretation. But there's no strict need to worry about this right now.

MichaelMcDonnell:
Are you saying this change shouldn't be merged?

HeroicKatora:
This can be merged. I'm just trying to communicate that this reveals some holes in our interfaces, in particular there seem to be _some_ conflicts in responsibilities to me. That's just fyi, since you're seemingly working somewhat intricately with the implementation. If you see possibility to restructure at a higher level or with larger scope, feel free to expand on your thoughts and propose some architectural changes as necessary.

MichaelMcDonnell:
Ok thanks! 

I have been having some ideas about how to restructure the parser to be simpler and potentially be able to recover from bad chunks, but I'll need to do some experiments first before I can say anything concretely.

PR #326, opened 2022-02-04
Assign instead of `replace`

MichaelMcDonnell:
Use `swap` to swap the buffers instead of `replace`. It makes more sense and there is no return value that needs to be ignored.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -562,7 +562,7 @@ impl<R: Read> Reader<R> {
             (false, InterlaceInfo::Null)
         };
         // swap back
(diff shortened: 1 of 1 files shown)


fintelia:
As pointed out by the [#[must_use] note on mem::replace](https://doc.rust-lang.org/src/core/mem/mod.rs.html#831) we've been overthinking this:

> if you don't need the old value, you can just assign the new value directly

Could you implement that instead?

MichaelMcDonnell:
Doh! Thanks for pointing that out @fintelia! I've replaced the `swap` with assigning.

PR #327, opened 2022-02-04
Use byte string literal for `ChunkType`

MichaelMcDonnell:
I find the using a byte string literal easier to read than using arrays. The data still ends up in arrays, and you still get a compiler error if the byte string literal is longer than 4 bytes.


fintelia (approved):

MichaelMcDonnell:
Thanks!

PR #328, opened 2022-02-05
Fix transparency error handling

MichaelMcDonnell:
The original code created an error but never returned it. Also, it didn't use `OutsidePlteIdat` error variant so this also fixes an unused variant warning.

The code now checks that the transparency chunk is in the right place and returns an error message if it isn't. The transparency chunk is supposed to be after a PLTE chunk and before an IDAT chunk according to the [specification](https://www.w3.org/TR/PNG/#5ChunkOrdering).

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -154,7 +154,6 @@ pub(crate) enum FormatErrorInner {
         kind: ChunkType,
     },
     /// 4.3., some chunks must be between PLTE and IDAT.
-    // FIXME: why are we not using this?
     OutsidePlteIdat {
         kind: ChunkType,
     },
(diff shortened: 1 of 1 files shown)


src/decoder/stream.rs
```diff
@@ -856,10 +856,18 @@ impl StreamingDecoder {
...
+                // The transparency chunk must be after the palette chunk and
+                // before the data chunk.
+                if info.palette.is_none() {
+                    return Err(DecodingError::Format(
+                        FormatErrorInner::AfterPlte { kind: chunk::tRNS }.into(),
+                    ));
+                } else if self.have_idat {
+                    return Err(DecodingError::Format(
+                        FormatErrorInner::OutsidePlteIdat { kind: chunk::tRNS }.into(),
+                    ));
```
HeroicKatora, line 868: There's a `FIXME` in the variant documentation about this variant not being used and wondering where it _should_ be used. Can you remove it?

HeroicKatora (commented):
:+1:

src/decoder/stream.rs
MichaelMcDonnell: Good catch! I've amended the pull request to remove the `FIXME` in the variant documentation. Thanks!

HeroicKatora (approved):

Issue #329, opened 2022-02-06
File size is zero after calling Writer::finish


jforberg:
The size of the encoder's output file is still zero after calling finish.

The documentation suggests that finish() should finish writing out the file.

I added a repro case to your test suite, here: https://github.com/jforberg/image-png/tree/finish-bug-repro

fintelia:
It looks like the issue is that `Writer::finish` doesn't call flush on the wrapped writer. Since you are using a BufWriter, that means that the bytes are not pushed out to disk until the object is dropped

HeroicKatora:
This is a bug in `png`. We _could_ just ensure to call `flush()` on the underlying writer, and indeed this would allow us to report the error from the flush if any, but the bug goes a bit deeper than that. https://github.com/image-rs/image-png/blob/master/src/encoder.rs#L959-L960

This was introduced to stop the _inherent_ drop from also writing a second IEND chunk. However, it has the side effect of alos not calling other `Drop` implementations of fields. It could leak memory if called with `Cursor<Vec<u8>>` or similar resource-owning writers.

fintelia:
Yeah, we should probably get rid of the ManuallyDrop stuff in favor of a iend_written field that gets checked in the drop impl _and_ also flush the inner writer in case it is something like a `&mut BufWriter` which IIRC won't actually flush itself upon being dropped

PR #330, opened 2022-02-06
Finish drops correctly

HeroicKatora:
Reuses the `frames_written` as a termination flag, cleaning up and documenting related code. I'd really like to use destructuring here but Rust doesn't let us. The alternative of having a dedicated drop flag can also be discussed.

Closes: #329

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -2205,6 +2205,40 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn finish_drops_inner_writer() -> Result<()> {
+        struct NoWriter<'flag>(&'flag mut bool);
+
+        impl Write for NoWriter<'_> {
+            fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+                Ok(buf.len())
+            }
+            fn flush(&mut self) -> io::Result<()> {
+                Ok(())
+            }
+        }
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -704,19 +710,29 @@ impl<W: Write> Writer<W> {
             }
         }
 
-        self.written += 1;
+        self.image_written();
+
+        Ok(())
+    }
+
+    fn image_written(&mut self) {
```
fintelia, line 718: Perhaps call this `increment_frames_written`?

src/encoder.rs
```diff
@@ -452,10 +452,16 @@ impl<'a, W: Write> Encoder<'a, W> {
...
 pub struct Writer<W: Write> {
+    /// The underlying writer.
     w: W,
+    /// The local version of the `Info` struct.
     info: PartialInfo,
+    /// Global encoding options.
     options: Options,
-    written: u64,
+    /// The total number of image frames, counting all consecutive IDAT and fdAT chunks.
+    /// Note: we use `u64::MAX` as a flag to note when the IEND chunk was also added.
```
fintelia, line 462: Could we just add an extra bool field instead of using u64::MAX as a sentinel value?

fintelia (approved):

PR #333, opened 2022-02-17
Support ignore text chunk while decode

NoXF:
I find some png convert from jpeg may keep "EXIF" info into text chunk, but use the wrong way. This will make the whole parsing fail.

Most of the time, I don't use text chunks, so I add a ignore text chunk config, but not sure it is a common need.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1,8 +1,8 @@
 mod stream;
 mod zlib;
 
-pub use self::stream::{Decoded, DecodingError, StreamingDecoder};
 use self::stream::{DecodeConfig, FormatErrorInner, CHUNCK_BUFFER_SIZE};
+pub use self::stream::{Decoded, DecodingError, StreamingDecoder};
 
 use std::io::{BufRead, BufReader, Read, Write};
 use std::mem;
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -75,6 +75,8 @@ pub struct Decoder<R: Read> {
     transform: Transformations,
     /// Limits on resources the Decoder is allowed to use
     limits: Limits,
+    /// Ignore text chunk when decode
+    ignore_text_chunk: bool,
 }
 
 /// A row of data with interlace information attached.
@@ -137,6 +139,7 @@ impl<R: Read> Decoder<R> {
             r,
             transform: Transformations::IDENTITY,
             limits,
+            ignore_text_chunk: false,
         }
     }
 
@@ -167,9 +170,18 @@ impl<R: Read> Decoder<R> {
         self.limits = limits;
     }
 
+    /// Ignore parse text(text/ztxt/itxt) chunk while read meta data
+    pub fn ignore_text_chunk(&mut self) {
+        self.ignore_text_chunk = true;
+    }
+
     /// Reads all meta data until the first IDAT chunk
     pub fn read_info(self) -> Result<Reader<R>, DecodingError> {
-        let mut reader = Reader::new(self.r, StreamingDecoder::new(), self.transform, self.limits);
+        let mut decoder = StreamingDecoder::new();
+        if self.ignore_text_chunk {
+            decoder.ignore_text_chunk();
+        }
+        let mut reader = Reader::new(self.r, decoder, self.transform, self.limits);
         reader.init()?;
 
         // Check if the output buffer can be represented at all.
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -2,7 +2,7 @@ mod stream;
 mod zlib;
 
 pub use self::stream::{Decoded, DecodingError, StreamingDecoder};
-use self::stream::{FormatErrorInner, CHUNCK_BUFFER_SIZE};
+use self::stream::{DecodeConfig, FormatErrorInner, CHUNCK_BUFFER_SIZE};
 
 use std::io::{BufRead, BufReader, Read, Write};
 use std::mem;
@@ -75,8 +75,7 @@ pub struct Decoder<R: Read> {
     transform: Transformations,
     /// Limits on resources the Decoder is allowed to use
     limits: Limits,
-    /// Ignore text chunk when decode
-    ignore_text_chunk: bool,
+    decode_config: DecodeConfig,
 }
 
 /// A row of data with interlace information attached.
@@ -139,7 +138,7 @@ impl<R: Read> Decoder<R> {
             r,
             transform: Transformations::IDENTITY,
             limits,
-            ignore_text_chunk: false,
+            decode_config: DecodeConfig::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -167,9 +170,18 @@ impl<R: Read> Decoder<R> {
         self.limits = limits;
     }
 
+    /// Ignore parse text(text/ztxt/itxt) chunk while read meta data
+    pub fn ignore_text_chunk(&mut self) {
+        self.ignore_text_chunk = true;
+    }
```
HeroicKatora, line 176: The style for other options is `fn set_*(&mut self, Value)`. I don't particularly care about the `set` prefix—although the API guidelines recommend reserving pure names without prefix for getters if there are any–but would prefer taking a `bool` parameter here.

src/decoder/mod.rs
```diff
@@ -167,9 +170,18 @@ impl<R: Read> Decoder<R> {
...
+
     /// Reads all meta data until the first IDAT chunk
     pub fn read_info(self) -> Result<Reader<R>, DecodingError> {
-        let mut reader = Reader::new(self.r, StreamingDecoder::new(), self.transform, self.limits);
+        let mut decoder = StreamingDecoder::new();
+        if self.ignore_text_chunk {
+            decoder.ignore_text_chunk();
+        }
```
HeroicKatora, line 183: This would also remove some control flow here (in particular we don't depend on it defaulting to `false` in streaming).

HeroicKatora (changes requested):

HeroicKatora:
@NoXF also check if those chunks may be affected #337 / #338 . The decoder might not be at fault here.

NoXF:
> 

the images I tested was not encoded by image-png

src/decoder/mod.rs
```diff
@@ -187,6 +189,12 @@ impl<R: Read> Decoder<R> {
     pub fn set_transformations(&mut self, transform: Transformations) {
         self.transform = transform;
     }
+
+    /// Set the custom streaming decoder
+    pub fn set_streaming_decoder(&mut self, streaming_decoder: StreamingDecoder) {
+        self.streaming_decoder = streaming_decoder;
+    }
```
HeroicKatora, line 196: I see the benefits of this idea, it is a good decomposition in principle. However, it's may not be as simple as it appears and is certainly more complex than having its own `set_ignore_text_chunk`.

* Can you provide the usage with `set_ignore_text_chunk` as an example? This is a somewhat implicit way of performing the desired operation and I fear that some users may not understand exactly how to make use of it.

* If the stream decoder has been used before, is this intended to work correctly? What exactly _is_ correct? Should it be possible to decoder the header with an instance then pass it to `Decoder` and expect it to continue? Or should the `Decoder` always restart at a new `IHDR`?

HeroicKatora:
> the images I tested was not encoded by image-png

The linked issue is about incorrect errors during _decoding_ of images with `image-rs/png`. The specific images in question had been encoded by Photoshop, as far as I know, and it was also related to a specific text chunk that Photoshop apparently inserts.

src/decoder/mod.rs
NoXF: You are right, I looked at StreamingDecoder and realized that we should not use a low-level interface(StreamingDecoder) out of Decoder, it will confuse the other users. However, I think ignore_text_chunk should be a Decoder configuration that is passed to StreamingDecoder when inner StreamingDecoder is used. but in the current interface, for example, read_info, we lack to set StreamingDecoder to ignore_text_chunk.

this set_streaming_decoder may change to set_decoder_config and pass the setting to streaming_decoder. and I will give a commit soon.

> The linked issue is about incorrect errors during decoding of images with image-rs/png. The specific images in question had been encoded by Photoshop, as far as I know, and it was also related to a specific text chunk that Photoshop apparently inserts.

Yes, it may have come from Photoshop, I found the error image has some jpeg info written in the iTXT chunk. Looks like it did some matting.

HeroicKatora: In order to avoid duplication but also allow configuration `StreamingDecoder` in `read_info`, what are your thoughts about introducing the functionality like so:

```rust
/// Private, for now. Mostly because naming is hard and I'm unsure about the scope of its responsibilities.
/// Just want to clarify which fields the Decoder shares with StreamingDecoder.
#[derive(Clone, Copy)]
// Actually, Copy may be too restrictive in the future. So another reason not to expose this yet.
struct ChunkHandling {
    ignore_txt_chunk: bool,
    // In a separate PR, maybe we should move `limit: Limit` into this?
}

pub struct Decoder<R: Read> {
    …,
    chunks: ChunkHandling,
}

struct StreamingDecoder {
    …,
    pub(crate) chunks: ChunkHandling,
}

impl<R: Read> Decoder<R> {
    pub fn read_info(self) -> Result<Reader<R>, DecodingError> {
        let mut stream =  StreamingDecoder::new();
        stream.chunks  = self.chunks;
        let mut reader = Reader::new(self.r, stream, self.transform, self.limits);
        reader.init()?;
    }

   pub fn set_ignore_text_chunk(&mut self, ignore_text_chunk: bool) {
        self.chunks.ignore_text_chunk = ignore_text_chunk;
    }
}
```

This way the details are hidden (for now) while we iterate over the exact options and control that we want to allow the caller to customize in chunk handling.

NoXF: yes, that is basically how I did it. 

and this is the image I mentioned, you can look into it if you're interested.

* https://img.alicdn.com/i3/6000000005544/O1CN01lz0UEm1qpCWNP6YW8_!!6000000005544-2-mia.png

HeroicKatora: Your image seems to have been affected, and #337 fixed it. Locally,

```bash
$ cargo run --example pngcheck -- /tmp/O1CN01lz0UEm1qpCWNP6YW8_!!6000000005544-2-mia.png
OK: /tmp/O1CN01lz0UEm1qpCWNP6YW8_!!6000000005544-2-mia.png (2362x2362, 8-bit RGB+alpha, non-interlaced, 92.6%)
```

src/decoder/stream.rs
```diff
@@ -383,6 +383,18 @@ impl From<TextDecodingError> for DecodingError {
...
+pub struct DecodeConfig {
+    ignore_text_chunk: bool,
+}
+
+impl DecodeConfig {
+    pub fn set_ignore_text_chunk(&mut self, ignore_text_chunk: bool) {
+        self.ignore_text_chunk = ignore_text_chunk;
+    }
+}
+
```
HeroicKatora, line 397: I think the _user facing_ part was more consistent before. Then `Decoder` and `StreamingDecoder` have more similar interfaces.

```rust
pub(crate) struct DecodeConfig { … }

impl StreamingDecoder {
    pub fn set_ignore_text_chunk(&mut self, ignore_text_chunk: bool)  { … }
}
```

src/decoder/stream.rs
```diff
@@ -447,6 +461,11 @@ impl StreamingDecoder {
         self.info.as_ref()
     }
 
+    /// Set decode config
+    pub fn set_decode_config(&mut self, decode_config: DecodeConfig) {
```
HeroicKatora, line 465:
```suggestion
    pub(crate) fn set_decode_config(&mut self, decode_config: DecodeConfig) {
```

following the above comment.

src/decoder/mod.rs
```diff
@@ -187,6 +192,18 @@ impl<R: Read> Decoder<R> {
...
+    ///
+    /// eg.
+    /// ```
+    /// let mut decoder = png::Decoder::new(File::open("tests/pngsuite/basi0g01.png");
+    /// decoder.set_ignore_text_chunk(true);
+    /// assert!(decoder.read_info().is_ok());
+    /// ```
+    pub fn set_ignore_text_chunk(&mut self, ignore_text_chunk: bool) {
+        self.decode_config.set_ignore_text_chunk(ignore_text_chunk);
+    }
```
HeroicKatora, line 206: This looks good :+1:

src/decoder/mod.rs
```diff
@@ -187,6 +192,18 @@ impl<R: Read> Decoder<R> {
     pub fn set_transformations(&mut self, transform: Transformations) {
         self.transform = transform;
     }
+
+    /// Set decoder if ignore text chunk while parse
```
HeroicKatora, line 196:
```suggestion
    /// Set the decoder to ignore all text chunks while parsing.
```

HeroicKatora (commented):
One grammar nit that I missed.

HeroicKatora (approved):
Looking good, thanks for your work.

PR #334, opened 2022-02-18
Add duplicate chunk check

MichaelMcDonnell:
The duplicate chunk error variant was unused and resulted in a warning when compiling. Only some chunk types are allowed to appear multiple times according to the [specification](https://www.w3.org/TR/PNG/#5ChunkOrdering). I have added checks for that. This also fixes the unused variant warning when compiling.

The `parse_chunk` method is already checking if the info is `None` so it is safe to unwrap the info in the subsequent `parse_*` methods.

Ideally the parsing should be refactored at some point so that the parsing methods only receive what they need and not a reference to the whole mutable self. For example, the header, which creates the info, must appear right after the signature. That means that the info could be passed to the subsequent `parse_*` methods and the they would not need to unwrap it. That would, however, be a large complicated change.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -158,7 +158,6 @@ pub(crate) enum FormatErrorInner {
         kind: ChunkType,
     },
     /// 4.3., some chunks must be unique.
-    // FIXME: why are we not using this?
     DuplicateChunk {
         kind: ChunkType,
     },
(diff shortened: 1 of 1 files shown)


src/decoder/stream.rs
```diff
@@ -807,83 +800,90 @@ impl StreamingDecoder {
...
+                    }
 
-                info.trns = Some(Cow::Owned(vec));
-                Ok(Decoded::Nothing)
+                    info.trns = Some(Cow::Owned(vec));
+                    Ok(Decoded::Nothing)
+                }
+                c => Err(DecodingError::Format(
+                    FormatErrorInner::ColorWithBadTrns(c).into(),
+                )),
```
HeroicKatora, line 872: I found this more readable without extra indent.

HeroicKatora (commented):
The rework of receiving isn't even that appealing to me. There are enough chunks that either have dependencies or conflict on each other or can benefit from having more context information in any case (e.g. gAMA–cHRM–iCCP), various animation related chunks.

I do see value in somehow _statically_ clarifying that the `info` is available, though.

src/decoder/stream.rs
MichaelMcDonnell: Yes, it is much easier to read without the extra indent. I've fixed that.

HeroicKatora (approved):

MichaelMcDonnell:
Thanks!

PR #336, opened 2022-02-21
Forbid unsafe code again.

### LINKED ISSUE #335: Prefer panicking to unsafe during text chunks decoding
This code

https://github.com/image-rs/image-png/blob/592534409607b21bb84e521c9964e0f0f1436708/src/text_metadata.rs#L180..L188

doesn't have to be unsafe. Yes, it's perfectly safe, but I would prefer a panic and `#![forbid(unsafe_code)]` instead. It's will not affect performance in any way.

RazrFalcon:
Removes unnecessary use of unsafe code. No performance implications.

Closes #335

### DIFF
--- a/src/text_metadata.rs
+++ b/src/text_metadata.rs
@@ -177,11 +177,11 @@ fn encode_iso_8859_1_iter(text: &str) -> impl Iterator<Item = Result<u8, TextEnc
         .map(|c| u8::try_from(c as u32).map_err(|_| TextEncodingError::Unrepresentable))
 }
 
-#[allow(unsafe_code)]
 fn decode_ascii(text: &[u8]) -> Result<&str, TextDecodingError> {
     if text.is_ascii() {
-        // SAFETY: ASCII is a subset of UTF-8.
-        unsafe { Ok(std::str::from_utf8_unchecked(text)) }
+        // `from_utf8` cannot panic because we're already checked that `text` is ASCII-7.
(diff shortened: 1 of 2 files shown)


fintelia (approved):

src/text_metadata.rs
```diff
@@ -424,7 +414,9 @@ impl ITXtChunk {
             return Err(TextDecodingError::InvalidCompressionMethod);
         }
 
-        let language_tag = decode_ascii(language_tag_slice)?.to_owned();
+        let language_tag = std::str::from_utf8(language_tag_slice)
+            .map_err(|_| TextDecodingError::Unrepresentable)?
+            .to_string();
```
HeroicKatora, line 419: What if we _want_ this error on non-ascii-utf8? Afterall, such an image will _not_ decode the same with all other decoders and this is just a guess with regards to the text encoding. There are several other ascii-supersets that may be valid guesses here.

HeroicKatora:
I want to add, last time `Ascii` as a module in `core::str` was discussed (indeed, with exactly this method given as a possible use-case) it was determined to be unecessary…

RazrFalcon:
Then we have to use panic after all. Which is fine, since it's unreachable. 

It would be nice to have something like `std::str::from_ascii`.

HeroicKatora (approved):
Text decoding shouldn't be the main performance cost in any case, so double validation is not that costly. Let's just hope and wait for a std interface in the future.

PR #339, opened 2022-02-26
Bump version and metadata for 0.17.4

HeroicKatora:


HeroicKatora:
Feel free to review patch notes already, but let's do one round of dependency bumping. Both deflate and miniz_oxide have new major versions.

fintelia (approved):

Issue #340, opened 2022-02-27
DuplicateChunk error for image with sRGB chunk preceding gAMA chunk


DutChen18:
This image with an sRGB chunk preceding a gAMA chunk fails to decode in 0.17.4 but it decodes fine in 0.17.3.
https://user-images.githubusercontent.com/6209323/155889742-979e8698-2e87-4206-8930-a1a454c2dc10.png

```
$ pngcheck raw_strips/006e8720-00d7-425d-9382-7f1676970f5e.png 
ERROR: raw_strips/006e8720-00d7-425d-9382-7f1676970f5e.png
raw_strips/006e8720-00d7-425d-9382-7f1676970f5e.png: Chunk ChunkType { type: gAMA, critical: false, private: false, reserved: false, safecopy: false } must appear at most once.
```

I believe the breaking change to be from #334

The error originates from this check: https://github.com/image-rs/image-png/blob/bb55a6646b2493b1411dacb870faa5ef49003a3c/src/decoder/stream.rs#L959-L963
Because `source_gamma` is set in `parse_srgb` here: https://github.com/image-rs/image-png/blob/bb55a6646b2493b1411dacb870faa5ef49003a3c/src/decoder/stream.rs#L991-L995

ScranchNew:
I still encountered this error with version 0.24.8 with this image:

![Endless banner](https://github.com/image-rs/image-png/assets/35395613/ddca1264-1885-49a0-a582-afb8166259e5)

The Error shown was: `Decoding(DecodingError { format: Exact(Png), underlying: Some(Format(FormatError { inner: DuplicateChunk { kind: ChunkType { type: iCCP, critical: false, private: false, reserved: false, safecopy: false } } })) })`.

Running with version 0.23.14 currently solves the issue.

fintelia:
@ScranchNew Make sure you run `cargo update`. If you're still seeing this failure, please open a new issue with both the code you're using and the exact `png` crate version (0.23.14 and 0.24.8 are versions of the `image` crate which are compatible _multiple_ `png` crate versions).

ogoffart:
If one can still reproduce this issue with the last png version, the best would be to open the issue directly on the png crate repository and attach the exact image. https://github.com/image-rs/image-png

PR #341, opened 2022-03-01
Separate gAMA/cHRM chunk and contents

HeroicKatora:
Introduces two new fields to `Info`: `gama_chunk` and `chrm_chunk` which hold the _original_ contents of those chunks regardless of any overwriting/ignore by sRGB. They are also used to track if such a chunk was already parsed instead of reusing the shared fields. Since the struct is `#[non_exhaustive]` this is a backwards compatible change.

Closes: #340 (still needs a regression test)


HeroicKatora:
I'll already preparing a changelog and release PR directly afterwards, let's discuss the changelog item there.

tests/bugfixes/License.toml
```diff
@@ -0,0 +1,14 @@
...
+
+[[files]]
+file = "tests/bugfixes/gama-srgb-order-issue#304.png"
+SPDX-FileCopyrightText = "2013 zaimoku_woodpile <https://www.flickr.com/photos/11250735@N07/>"
+SPDX-License-Identifier = "CC-BY-2.0"
+Comment: """
+	Taken from: <https://www.flickr.com/photos/11250735@N07/8561945042>
+	with Modifications: Cropped, resized the image
+	Taken from: <https://raw.githubusercontent.com/slint-ui/slint/v0.1.6/examples/imagefilter/cat_preview_round.png>
+"""
```
fintelia, line 14: Is the syntax here correct? GitHub is highlighting it in red...

HeroicKatora: I would [assume so](https://github.com/Randy8080/reference/blob/main/toml.md#multiline-string) and my editor is also highlighting this properly.

HeroicKatora: Oops, that's why you're referring to. Yeah..

PR #350, opened 2022-08-05
Use Flate2 for encoding

Craig-Macomber:
Thus uses flat2 for encoding instead of the deflate crate.

Since flat2 defaults to using miniz_oxide, this shares the same underlying compression library as the decoding.

See issue #242

I'm not sure this is the right solution, and it does not full handle all the compression options (Huffman and Rle just use "fast" mode).


Cargo.toml
```diff
@@ -20,10 +20,10 @@ include = [
...
 
 [dependencies]
-bitflags = "1.0"
-crc32fast = "1.2.0"
-deflate = "1.0"
-miniz_oxide = "0.5.1"
+bitflags = "^1.0"
+crc32fast = "^1.2.0"
+flate2 = "^1.0"
+miniz_oxide = "^0.5.0"
```
fintelia, line 26: There's no need to add the `^` symbol here. Cargo treats `"1.0"` the same as `"^1.0"`

src/encoder.rs
```diff
@@ -2289,13 +2288,13 @@ mod tests {
...
-            Compression::Default => deflate::CompressionOptions::default(),
-            Compression::Fast => deflate::CompressionOptions::fast(),
-            Compression::Best => deflate::CompressionOptions::high(),
-            Compression::Huffman => deflate::CompressionOptions::huffman_only(),
-            Compression::Rle => deflate::CompressionOptions::rle(),
+            Compression::Default => flate2::Compression::default(),
+            Compression::Fast => flate2::Compression::fast(),
+            Compression::Best => flate2::Compression::best(),
+            Compression::Huffman => flate2::Compression::fast(), // TODO
+            Compression::Rle => flate2::Compression::fast(),     // TODO
```
fintelia, line 2297: Could you implement these two as `flate2::Compression::none()` instead? That way they'd still technically uphold the same guarantees as before

Cargo.toml
Craig-Macomber: I guess I've been doing too much npm recently. Cargo docs are actually pretty clear about it (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements)

```suggestion
bitflags = "1.0"
crc32fast = "1.2.0"
flate2 = "1.0"
miniz_oxide = "0.5.0"
```

src/encoder.rs
Craig-Macomber: Done. Also bumped crate version so I could reference the version in the deprecation notice.

fintelia (approved):

HeroicKatora (approved):
Looks okay. I don't particularly like the deprecation but it seems harmless enough. (Note: if we *had* a compression level 'plaintext'—or as `deflate` calls it's unimplemented option: `_ForceStored`—and this would have removed it, I would have vetoed; but alas, we do not). Definitely a maintenance win.

fintelia:
In fact, this change actually allows us to add a plaintext compression level because flate2 _does_ support it!

oyvindln:
> Looks okay. I don't particularly like the deprecation but it seems harmless enough. (Note: if we _had_ a compression level 'plaintext'—or as `deflate` calls it's unimplemented option: `_ForceStored`—and this would have removed it, I would have vetoed; but alas, we do not). Definitely a maintenance win.

If there is a need for RLE and huffman only it should be pretty easy to add support for it to flate2 as it's already supported by the back-ends I think. (RLE maybe via a window size 1 if there is no explicit option for it). Not sure where one would want to use the huffman only variant though, it doesn't have much purpose, RLE may have a slight use for compression speed I suppose.

HeroicKatora:
Don't feel compelled to do so based on this singular data point, I don't quite see their purpose either. No users *wants* RLE-only (unless they targeted non-compliant decoders for some reason?)—they want *speed*. If anything then the encoder should just switch to rle-only based on the speed setting. *If* it's actually so much faster.

fintelia:
The fastest approach is probably going to be [fpnge's strategy](https://www.lucaversari.it/FJXL_and_FPNGE.pdf) of using specially constructed Huffman tables so that the resulting codes are conducive to vectorized encoding. However, it does bake in some assumptions about the kind of filters that PNG uses, so it might not be suited to a generic compression library.

PR #351, opened 2022-08-06
Implement `Default` trait for `Compression`

jay3332:
Implements the Default trait for `common::Compression`, which just returns `Compression::Default`. 

I chose to stick with not using the new `#[default]` since that is a relatively new feature and would likely increase the MSRV


fintelia (approved):

fintelia:
Thanks!

PR #352, opened 2022-08-11
Remove goto! macro

fintelia:
Replacing all instances of the macro with the equivalent code expansion makes the function easier to read. Ironically, it results in _fewer_ lines of code, not even considering the lines to declare the macro itself.


HeroicKatora (approved):
:smile:

PR #353, opened 2022-08-11
Add Decoder::read_header_info method

fintelia:
Adds a method on Decoder to read the PNG header before starting the main decoding process. This is useful because decoding limits can still be set after the method has been called, which is required to properly implement the [`image::ImageDecoder`](https://docs.rs/image/latest/image/trait.ImageDecoder.html) trait in a way that doesn't [risk an OOM](https://github.com/image-rs/image/issues/1748).

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -69,8 +69,7 @@ impl Default for Limits {
 
 /// PNG Decoder
 pub struct Decoder<R: Read> {
-    /// Reader
-    r: R,
+    read_decoder: ReadDecoder<R>,
     /// Output transformations
     transform: Transformations,
     /// Limits on resources the Decoder is allowed to use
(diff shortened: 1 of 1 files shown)


src/decoder/mod.rs
```diff
@@ -169,18 +172,30 @@ impl<R: Read> Decoder<R> {
...
+    ///
+    /// Most image metadata will not be read until `read_info` is called, so those fields will be
+    /// None or empty.
+    pub fn read_header_info(&mut self) -> Result<&Info, DecodingError> {
+        while self.read_decoder.info().is_none() {
+            if self.read_decoder.decode_next(&mut Vec::new())?.is_none() {
+                return Err(DecodingError::Format(
+                    FormatErrorInner::MissingImageData.into(),
+                ));
+            }
```
HeroicKatora, line 185: The specific error variant seems improper here. But maybe that's just our current usage if there is no better one.

fintelia: Happy to switch it to something else. I was just matching the current behavior: https://github.com/image-rs/image-png/blob/536f76f3660fde6ba8be876c7bac86bae7298cb9/src/decoder/mod.rs#L385-L389

HeroicKatora: It seems more out-of-place now, especially if the method is being called explicitly. Maybe `UnexpectedEof`?

HeroicKatora (approved):

PR #362, opened 2022-11-22
Add corpus-bench example

fintelia:
Example output when run on the [QOI benchmark suite](https://qoiformat.org/benchmark/):

```
Directory                                     Ratio             Encode                    Decode
---------                                    -------     --------------------      --------------------
qoi_benchmark_suite/images/pngimg             20.62%      42 mps   0.16 GiB/s       69 mps   0.26 GiB/s
qoi_benchmark_suite/images/icon_512            7.79%      76 mps   0.28 GiB/s       96 mps   0.36 GiB/s
qoi_benchmark_suite/images/textures_photo     61.56%      27 mps   0.08 GiB/s       56 mps   0.16 GiB/s
qoi_benchmark_suite/images/photo_kodak        61.44%      24 mps   0.07 GiB/s       52 mps   0.15 GiB/s
qoi_benchmark_suite/images/screenshot_game    25.03%      43 mps   0.14 GiB/s       72 mps   0.23 GiB/s
qoi_benchmark_suite/images/icon_64            27.97%      32 mps   0.12 GiB/s       56 mps   0.21 GiB/s
qoi_benchmark_suite/images/photo_tecnick      55.88%      27 mps   0.08 GiB/s       52 mps   0.15 GiB/s
qoi_benchmark_suite/images/textures_pk        63.59%      32 mps   0.09 GiB/s       55 mps   0.16 GiB/s
qoi_benchmark_suite/images/photo_wikipedia    61.83%      30 mps   0.09 GiB/s       53 mps   0.15 GiB/s
qoi_benchmark_suite/images/textures_pk01      36.79%      27 mps   0.10 GiB/s       52 mps   0.20 GiB/s
qoi_benchmark_suite/images/textures_plants    24.33%      41 mps   0.15 GiB/s       66 mps   0.25 GiB/s
qoi_benchmark_suite/images/textures_pk02      43.02%      30 mps   0.11 GiB/s       48 mps   0.18 GiB/s
qoi_benchmark_suite/images/screenshot_web      9.14%      76 mps   0.29 GiB/s       98 mps   0.37 GiB/s

Total                                        28.472%      40 mps   0.14 GiB/s       66 mps   0.22 GiB/s
```

PR #363, opened 2022-11-22
Optimize Sub and Paeth filtering

fintelia:
This PR changes the `filter_internal` method to make it more conducive to auto-vectorization. Using the benchmark infrastructure from #362, it increases overall encoding performance by ~30%:

before...
```
Directory                                     Ratio             Encode
---------                                    -------     --------------------
Total                                        28.472%      40 mps   0.14 GiB/s
```

after...
```
Directory                                     Ratio             Encode
---------                                    -------     --------------------
Total                                        28.472%      60 mps   0.20 GiB/s
```

Also fixes a bug I noticed while making this change: `StreamWriter`'s `prev_buf` would contain filtered image data rather than unfiltered pixels (and adds a test to confirm).

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -676,8 +676,14 @@ impl<W: Write> Writer<W> {
         let adaptive_method = self.options.adaptive_filter;
 
         for line in data.chunks(in_len) {
-            current.copy_from_slice(line);
-            let filter_type = filter(filter_method, adaptive_method, bpp, prev, &mut current);
+            let filter_type = filter(
+                filter_method,
+                adaptive_method,
+                bpp,
+                prev,
+                &line,
+                &mut current,
+            );
             zlib.write_all(&[filter_type as u8])?;
             zlib.write_all(&current)?;
             prev = line;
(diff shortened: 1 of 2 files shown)


src/encoder.rs
```diff
@@ -1561,12 +1567,14 @@ impl<'a, W: Write> Write for StreamWriter<'a, W> {
         self.to_write -= written;
 
         if self.index == self.line_len {
+            let mut filtered = vec![0; self.curr_buf.len()];
```
HeroicKatora, line 1570: Can we reuse this line buffer like we do in the decoder? But ultimately this is only a performance concern, small FIXME note would be more than enough.

src/filter.rs
```diff
@@ -209,47 +207,90 @@ fn filter_internal(
     bpp: usize,
     len: usize,
     previous: &[u8],
-    current: &mut [u8],
+    current: &[u8],
+    output: &mut [u8],
 ) -> FilterType {
     use self::FilterType::*;
 
+    const CHUNK_SIZE: usize = 32;
```
HeroicKatora, line 215: Is there any particular reason for this size? Experimental data? So that it fits in a specific SIMD register length for a single-load access?

HeroicKatora (commented):
Looks good to me in terms of implementation, just could use some more comments to capture some information on how it was derived.

src/filter.rs
fintelia: This is based on benchmarking a couple possible sizes. There wasn't a ton of difference between 16, 32, and 64 byte sizes but this option was fastest. The fact that the best size is equal to two 256-bit SIMD registers is probably not a coincidence, however.

fintelia:
@HeroicKatora I added an explanation of how the alternative Paeth filter implementation was derived.

src/filter.rs
HeroicKatora: That might make a good comment. Relying on auto-vectorization in combination with experimental constants it can quickly become hard to distinguish between incidental and intentional constants.

HeroicKatora: That would make a good comment. Relying on auto-vectorization in combination with experimental constants it can quickly become hard to distinguish between incidental and intentional constants.

PR #364, opened 2022-11-22
Drastically improve encoding performance 

fintelia:
The goal of this PR is to achieve faster-than-QOI encoding performance. It adds a custom zlib implementation that leverages assumptions about the patterns in PNG data to get reasonable compression ratios at phenomenal speeds. Takes inspiration from [fpng](https://github.com/richgel999/fpng) and [fpnge](https://github.com/veluca93/fpnge).

Builds on #363 and improves performance by several _times_...
```
Directory                                     Ratio             Encode                    Decode
---------                                    -------     --------------------      --------------------
Total                                        28.282%     294 mps   0.99 GiB/s       53 mps   0.18 GiB/s
```

qoibench on the same machine...
```
# Grand total for qoi_benchmark_suite/images/
        decode ms   encode ms   decode mpps   encode mpps   size kb    rate
libpng:       5.3        69.6         87.49          6.67       398   24.2%
stbi:         4.6        39.7        100.63         11.70       561   34.2%
qoi:          1.5         2.0        308.68        227.88       463   28.2%

```

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -668,27 +668,55 @@ impl<W: Write> Writer<W> {
 
         let prev = vec![0; in_len];
         let mut prev = prev.as_slice();
-        let mut current = vec![0; in_len];
 
-        let mut zlib = ZlibEncoder::new(Vec::new(), self.info.compression.to_options());
         let bpp = self.info.bpp_in_prediction();
         let filter_method = self.options.filter;
         let adaptive_method = self.options.adaptive_filter;
 
-        for line in data.chunks(in_len) {
-            let filter_type = filter(
-                filter_method,
-                adaptive_method,
-                bpp,
-                prev,
-                &line,
-                &mut current,
-            );
-            zlib.write_all(&[filter_type as u8])?;
-            zlib.write_all(&current)?;
-            prev = line;
-        }
-        let zlib_encoded = zlib.finish()?;
+        let zlib_encoded = match self.info.compression {
+            Compression::Fast => {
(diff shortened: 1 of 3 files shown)


Shnatsel:
https://github.com/image-rs/image-png/pull/370 should come in handy for testing this.

Shnatsel (approved):
No issues on the roundtrip fuzzer after 10 million executions

src/encoder.rs
```diff
@@ -668,27 +668,67 @@ impl<W: Write> Writer<W> {
...
+                }
+
+                let compressed = compressor.finish()?.into_inner();
+                if compressed.len()
+                    > fdeflate::StoredOnlyCompressor::<()>::compressed_size((in_len + 1) * height)
+                {
+                    let mut compressor =
+                        fdeflate::StoredOnlyCompressor::new(std::io::Cursor::new(Vec::new()))?;
+                    for line in data.chunks(in_len) {
+                        compressor.write_data(&[0])?;
```
fintelia, line 703: This hard-codes writing unfiltered image data even if the caller requested a filter. Given that the data is being written uncompressed, doing filtering again would only add performance cost (both for encoding _and_ subsequent decoding), but seems worth mentioning

fintelia:
The core compression logic has also been fuzzed directly here: https://github.com/image-rs/fdeflate/blob/main/fuzz/fuzz_targets/deflate.rs

src/encoder.rs
Shnatsel: This sounds like it would make a good code comment - could you add it?

PR #365, opened 2022-11-23
Add CIFuzz GitHub action

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where image-png already runs (https://github.com/google/oss-fuzz/tree/master/projects/image-png). CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,35 @@
+name: CIFuzz
+on: [pull_request]
+permissions: {}
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   permissions:
+     security-events: write
+   steps:
+   - name: Build Fuzzers
+     id: build
+     uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+     with:
+       oss-fuzz-project-name: 'image-png'
+       language: rust
+   - name: Run Fuzzers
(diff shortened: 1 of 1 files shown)


HeroicKatora:
The 300 seconds is twice as long as our currently longest running CI task, not including setup time. Not a necessary reason to decide this PR either way but maybe there's a more fitting place to integrate the long running target? Is the feedback that it provides even necessarily actionable at the PR head? Most certainly it should run at certain pre-determined points between individual software releases.

fintelia (approved):

DavidKorczynski:
> The 300 seconds is twice as long as our currently longest running CI task, not including setup time.

I reduced it to 120 sec to match the existing CI better

> Most certainly it should run at certain pre-determined points between individual software releases.

It's being run on a continuous basis by https://github.com/google/oss-fuzz and this is an extension of it, in that CIFuzz is OSS-Fuzz's Github action https://google.github.io/oss-fuzz/getting-started/continuous-integration/

fintelia:
Could you rebase these changes? I think that should resolve the one CI failure.

DavidKorczynski:
Thank you for the patience @fintelia

PR #366, opened 2022-11-24
Find MSRV that fixes CI

fintelia:


fintelia:
Conclusion: rustc 1.54 is the earliest where CI passes

HeroicKatora (approved):
Fine, that's still a few below the one required at `image`.

PR #367, opened 2022-11-25
[Perforamnce] Feature to use Zlib-ng

reinismu:
For my workflow (Gstreamer encoding 50 fps in 4k using https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/tree/main/video/png) I needed a bit of performance. This more or less solved my speed issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ include = [
 [dependencies]
 bitflags = "1.0"
 crc32fast = "1.2.0"
-flate2 = "1.0"
+flate2 = { version = "1.0", features = ["zlib-ng"], default-features = false }
 miniz_oxide = "0.6.0"
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ include = [
 [dependencies]
 bitflags = "1.0"
 crc32fast = "1.2.0"
-flate2 = { version = "1.0", features = ["zlib-ng"], default-features = false }
+flate2 = { version = "1.0", default-features = false }
 miniz_oxide = "0.6.0"
 
 [dev-dependencies]
@@ -38,8 +38,11 @@ features = ["glutin"]
 default-features = false
 
 [features]
+default = ["rust_backend"]
 unstable = []
 benchmarks = []
+rust_backend = ["flate2/rust_backend"]
+zlib-ng = ["flate2/zlib-ng"]
 
 [[bench]]
 path = "benches/decoder.rs"


fintelia:
Could you take a look at #364 and see if that helps for your usecase? If we can get away with pure Rust dependencies we avoid some difficulties around cross compilation and conflicting features (which seems to be causing a few of the CI failures).

reinismu:
> Could you take a look at #364 and see if that helps for your usecase? If we can get away with pure Rust dependencies we avoid some difficulties around cross compilation and conflicting features (which seems to be causing a few of the CI failures).

That broke encoding completely in my case :/ 
I did test https://github.com/image-rs/image-png/pull/363 also, but it seemed to be worse with it. (Tho my measuring here is based on how fast memory fills up. I have a queue before processing)

fintelia:
Hm, weird that #364 broke encoding for you. It would be super helpful if you could share exactly what went wrong in that other PR thread given that we're moving towards merging it...

reinismu:
> Hm, weird that #364 broke encoding for you. It would be super helpful if you could share exactly what went wrong in that other PR thread given that we're moving towards merging it...

Got cold so can't access my work PC, but you should be able to reproduce my case. (Not sure how familiar you are with gstreamer, but wont hurt to learn it)

Example pipeline that records 4k images.
`gst-launch-1.0 videotestsrc is-live=false num-buffers=10 ! videoconvert ! video/x-raw, format=RGB, width=3840 , height= 2160 ! rspngenc compression-level=2 filter=4 ! multifilesink location="out/frame%d.png"`

* You will need to install gstreamer https://gstreamer.freedesktop.org/documentation/installing/on-linux.html?gi-language=c

* Next `git clone https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs` and `cargo build -p gst-plugin-png --release`. It should create `.so` lib for png plugin.

* Add its dir to gst plugin path ` export GST_PLUGIN_PATH=<path-to-release-folder>:$GST_PLUGIN_PATH`
* Try out previously mentioned pipeline. Make sure that it works and png files are not corrupted
* Now change https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/main/video/png/Cargo.toml `png` to your performance branch.
* Recompile
* Try pipeline again and inspect generated PNGs (They should be corrupted)

fintelia:
I tried the steps you suggested but the resulting images seem to be valid. However, an old version of my branch did used to produce subtly corrupt images. Did you by chance try it before my initial comment above?

When you get a chance, if you're able to share one of the corrupted PNGs that would be very helpful for debugging purposes. This is the first frame I get when running your command with compression-level=1 (my branch only swaps out the compression algorithm if mode "fast" is selected):

frame0

reinismu:
Ou ye `compression-level=1 ` in pipeline. I was using commit `975d4f873eb9374d0fa6e4c31e17e88d36857a47`. Just checked the latest changes and it works :)
Tho wouldn't this feature flag help as well? Or you want this project to be pure Rust?

reinismu:
One drawback I see in your branch:
```
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '2022-12-02T12-33_day-2_red-table-7_h34.mov':
  Metadata:
    major_brand     : qt  
    minor_version   : 537199360
    compatible_brands: qt  
    creation_time   : 2022-12-02T12:33:01.000000Z
  Duration: 00:00:26.44, start: 0.000000, bitrate: 1987370 kb/s
  Stream #0:0[0x1](und): Video: png (png  / 0x20676E70), rgba(pc, gbr/bt709/iec61966-2-1, progressive), 3840x2160, 1987368 kb/s, SAR 1:1 DAR 16:9, 50 fps, 50 tbr, 5k tbn (default)
    Metadata:
      creation_time   : 2022-12-02T12:33:01.000000Z
      handler_name    : VideoHandler
      vendor_id       : [0][0][0][0]
```

My branch
```
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from '2022-11-25T14-28_day-2_red-table-7_h34.mov':
  Metadata:
    major_brand     : qt  
    minor_version   : 537199360
    compatible_brands: qt  
    creation_time   : 2022-11-25T14:28:30.000000Z
  Duration: 00:01:27.72, start: 0.000000, bitrate: 643139 kb/s
  Stream #0:0[0x1](und): Video: png (png  / 0x20676E70), rgba(pc, gbr/bt709/iec61966-2-1, progressive), 3840x2160, 643137 kb/s, SAR 1:1 DAR 16:9, 50 fps, 50 tbr, 5k tbn (default)
    Metadata:
      creation_time   : 2022-11-25T14:28:30.000000Z
      handler_name    : VideoHandler
      vendor_id       : [0][0][0][0]
```

Bitrate increase to 1987370 kb/s from bitrate: 643139 kb/s (3x)

HeroicKatora:
The problem isn't the native dependency per-se, but rather the way these are integrated (as in this case). It's a compile time _choice_. But features in `Cargo.toml` are additive, that is, the final build configuration is the union of all features specified through the whole dependency tree. In theory, the `png` library shouldn't even be drawn into this configuration choice of `flate2`.

Unfortunately, the usual case does not implement features as additive and `flate2`'s rust backend is *mutually exclusive* with the `zlib-ng` backend. I don't think that this is something we can or want do about here.

In particular, also note that this is breaking because downstream can (and probably did) specify the dependency on `png`:

    [dependencies.png]
    default-features = false

Imagine we could detect the encoder choice at runtime, by some available method in `flate2`. Compare this to SIMD detection in jpeg, for instance. There's a feature that enables the _detection_ and the SIMD code portion, but if that feature is not enabled then it will simply always use the slow fallback. That's purely additive and it suffices if the final binary enables it, no matter how deep in the dependency tree all other consumers will benefit without them having to explicitly specify or change their code base.

The depth really matters since I'd strongly veto propagating the flag to `image` as well, but we'd have to with the design of this PR / the design of `flate2`'s feature flags.

fintelia:
I checked the `flate2` source code and contrary to what's implied by its documentation, I believe that its features actually are additive _provided that you enable at least one of them_. If you request multiple backends, it just picks one in order of precedence:

1. [most preferred] zlib-ng
2. cloudflare-zlib
3. zlib
4. miniz_oxide

What this means is that you should be able to use the zilb-ng backend without making any changes to the png crate by simply adding the following to your Cargo.toml:
```
[dependencies]
flate2 = { version = "1", features = ["zlib-ng"], default-features = false }
```

PR #370, opened 2022-12-24
Add an initial roundtrip fuzz target

Shnatsel:
Add a fuzz target that encodes some random data with random parameters into a PNG image, decodes it, and validates that the decoded result is the same as the original data fed into the encoder.

Doesn't randomize some important parameters yet, but it's a start.

### DIFF
--- a/fuzz/fuzz_targets/roundtrip.rs
+++ b/fuzz/fuzz_targets/roundtrip.rs
@@ -4,8 +4,8 @@ use png::FilterType;
 #[macro_use] extern crate libfuzzer_sys;
 extern crate png;
 
-fuzz_target!(|data: &[u8]| {
-    if let Some((raw, encoded)) = encode_png(data) {
+fuzz_target!(|data: (u8, u8, u8, Vec<u8>)| {
+    if let Some((raw, encoded)) = encode_png(data.0, data.1, data.2, &data.3) {
         let (_info, raw_decoded) = decode_png(&encoded);
         // raw_decoded can be padded with zeroes at the end, not sure if that's correct
         let raw_decoded = &raw_decoded[..raw.len()];
@@ -13,23 +13,14 @@ fuzz_target!(|data: &[u8]| {
     }
 });
 
-const METADATA_BYTES: usize = 3;
-
-fn encode_png(data: &[u8]) -> Option<(&[u8], Vec<u8>)> {
-    // we use the first few bytes of the fuzzer input
-    // to randomize the PNG encoding settings
-    // so we need at least that much to be present
-    if data.len() <= METADATA_BYTES {
-        return None;
-    }
-
-    // randomize height
-    let height = data[0] as u32; // max 255
+fn encode_png(height: u8, filter: u8, compression: u8, data: &[u8]) -> Option<(&[u8], Vec<u8>)> {
+    // Convert untyped bytes to the correct types and validate them
+    // height
+    let height = height as u32;
     if height == 0 { return None }
-    // randomize filter
-    let filter = FilterType::from_u8(data[1])?;
-    // randomize compression
-    let compression = data[2];
+    // filter
+    let filter = FilterType::from_u8(filter)?;
+    // compression
     let compression = match compression {
         0 => png::Compression::Default,
         1 => png::Compression::Fast,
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use png::FilterType;
+#[macro_use] extern crate libfuzzer_sys;
+extern crate png;
+
+fuzz_target!(|data: &[u8]| {
+    if let Some((raw, encoded)) = encode_png(data) {
+        let (_info, raw_decoded) = decode_png(&encoded);
+        // raw_decoded can be padded with zeroes at the end, not sure if that's correct
+        let raw_decoded = &raw_decoded[..raw.len()];
```
HeroicKatora, line 11: Yes, that's correct. The buffer is for a full frame width×height, but with APNG and animation control chunks we might have frames smaller than that initial size.

Shnatsel: I'm actually seeing the resulting decoded buffer being *larger* than the original input buffer.

That happens only with `image-png` - [`zune-png` always roundtrips the same buffer size](https://github.com/etemesi254/zune-image/pull/22) (although it doesn't handle APNG).

fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -0,0 +1,76 @@
+#![no_main]
+
+use png::FilterType;
+#[macro_use] extern crate libfuzzer_sys;
+extern crate png;
+
+fuzz_target!(|data: &[u8]| {
```
fintelia, line 7: It might be more straightforward to have this take a tuple of `(width, height, filter, compression, data)` rather than extracting those from the first bytes of a `&[u8]`.

Shnatsel: I would need to add `#[cfg_attr(fuzz), derive(Arbitrary)]` to these structs, and probably also a feature that enables the `arbitrary` dependency. If that's OK with you, I can do that.

fintelia: I was thinking of just making the argument type something like `(u16, u16, u8, u8, &[u8])` since IIRC the Arbitrary trait is already implemented for tuples.

Shnatsel: Done

fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -0,0 +1,71 @@
...
+    let bytes_per_pixel = color_type.samples() as u32; // safe to cast - actually u8 internally
+    let width = data.len() as u32 / height / bytes_per_pixel;
+    if width == 0 { return None }
+    let total_pixels = height.checked_mul(width)?;
+    let total_bytes = total_pixels.checked_mul(bytes_per_pixel)?;
+
+    // perform the PNG encoding
+    let data_to_encode = &data[..total_bytes as usize];
+    let mut output: Vec<u8> = Vec::new();
+    { // scoped so that we could return the resulting Vec at the end
```
fintelia, line 48: Is this scope needed? Now that Rust uses non-lexical lifetimes, I'd expect it to work without a new scope

Shnatsel: Not anymore. Removed

fintelia (approved):
This looks good to me. Is it ready to merge?

Shnatsel:
Yes. It could be expanded upon later, but I think it's useful as is already. Didn't find any issues though.

fuzz/fuzz_targets/roundtrip.rs
Shnatsel: Oh wait this **is** necessary. And now that I've removed it, it doesn't compile :facepalm: 

I thought my IDE ran `cargo check`, but it didn't because fuzzing is a separate folder that's not part of the workspace

HeroicKatora: The CI tests don't catch this either?

Shnatsel: Fuzzing isn't run on CI and fuzz/ folder is excluded from the workspace because it requires nightly, not to break compilation of the main crate on stable.

oss-fuzz might catch it later, if they run your fuzzers.

PR #371, opened 2022-12-24
Use iterators in Up and Avg filtering

okaneco:
Consolidate Avg remainder loop to main loop

In addition to removing some bounds checks, this seems to reduce the instruction count and enable more autovectorization.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -269,20 +269,19 @@ fn filter_internal(
             Sub
         }
         Up => {
-            for i in 0..len {
-                output[i] = current[i].wrapping_sub(previous[i]);
+            for ((out, cur), &prev) in output.iter_mut().zip(current).zip(previous) {
+                *out = cur.wrapping_sub(prev);
             }
             Up
         }
         Avg => {
-            for i in (bpp..len).rev() {
-                output[i] = current[i].wrapping_sub(
-                    ((u16::from(current[i - bpp]) + u16::from(previous[i])) / 2) as u8,
-                );
-            }
-
-            for i in 0..bpp {
-                output[i] = current[i].wrapping_sub(previous[i] / 2);
+            for (((out, cur), &cur_minus_bpp), &prev) in output
+                .iter_mut()
+                .zip(current.iter())
+                .zip(core::iter::repeat(&0_u8).take(bpp).chain(current.iter()))
+                .zip(previous.iter())
+            {
+                *out = cur.wrapping_sub(((u16::from(cur_minus_bpp) + u16::from(prev)) / 2) as u8);
             }
             Avg
         }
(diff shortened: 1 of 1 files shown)


fintelia:
The `repeat(0).take(bpp).chain(...)` approach is clever.  A large percentage of total decoding time is spent in this method, so even small optimizations are significant.

I was thinking of trying the Paeth approach of chunks_exact + remainder + final bpp as three separate loops for all the filters. Did you try something like that and see worse results for Up and Avg?

(I'd probably factor out the there loops into a generic method that takes a closure to call per-pixel. Which should produce good results assuming everything gets inlined together.)

okaneco:
Is `corpus-bench.rs` the recommended way to get a feel for encoding speed? I set it to the encoder types and compared between branches, `chunks_exact + remainder + final bpp` is faster than the previous commit.

I saw the work you did with the other filters and this seemed like low hanging fruit adjacent to it which could be improved as a stopgap until something better. I've had cases where chunks_exact + residual ended up being the same speed or slower in personal code so I was hesitant without a decent way of measuring. 

Another hot function for encoding, I noticed the `filter::sum_filter` doesn't get vectorized with the `saturating_add` for some reason. `unfilter` could probably use some similar love to its filters.

fintelia (approved):

fintelia:
There's certainly tons of room for improvement in the adaptive filtering. Right now it does 9 passes over the input row (5x sum_buffer + 4x filter) which is way more than necessary. I experimented with doing one pass to compute all the sums at once, followed by a filtering pass with whichever one was best. That produced good results but we'd want to do some refactoring at the same time to avoid the growing amount of code duplication.

fintelia:
Also agreed on `unfilter`. That function is the main bottleneck in decoding, but turns out to be trickier to vectorize because of dependencies between pixels. (Except for the Up filter which only depends on the previous row)

src/filter.rs
```diff
@@ -269,20 +269,56 @@ fn filter_internal(
...
+                .zip(cur_chunks.remainder())
+                .zip(cur_minus_bpp_chunks.remainder())
+                .zip(prev_chunks.remainder())
+            {
+                *out = cur.wrapping_sub(((u16::from(cur_minus_bpp) + u16::from(prev)) / 2) as u8);
             }
 
             for i in 0..bpp {
-                output[i] = current[i].wrapping_sub(previous[i] / 2);
+                output[i] = current[i].wrapping_sub((u16::from(previous[i]) / 2) as u8);
```
fintelia, line 321: This change isn't actually necessary since the casts are no-ops

fintelia: Well not no-ops, but they don't change the result at all

okaneco: Addressed in the latest commit.

PR #373, opened 2022-12-30
Faster decoding of fdeflate encoded PNGs

fintelia:
When a PNG has been encoded with fdeflate compression method added in #364, the decoder can make some assumptions that improve decoding performance. This PR adds a check on whether fdeflate compression was used and decompresses the image data accordingly.

### DIFF
--- a/src/decoder/zlib.rs
+++ b/src/decoder/zlib.rs
@@ -3,10 +3,64 @@ use super::{stream::FormatErrorInner, DecodingError, CHUNCK_BUFFER_SIZE};
 use miniz_oxide::inflate::core::{decompress, inflate_flags, DecompressorOxide};
 use miniz_oxide::inflate::TINFLStatus;
 
+enum Compressor {
+    FullZlib(DecompressorOxide),
+    FDeflate(fdeflate::Decompressor),
+}
+impl Compressor {
+    fn decompress(
+        &mut self,
+        input: &[u8],
+        output: &mut [u8],
+        output_position: usize,
+        end_of_input: bool,
+    ) -> (TINFLStatus, usize, usize) {
+        match self {
+            Compressor::FullZlib(decompressor) => decompress(
+                decompressor,
+                input,
+                output,
+                output_position,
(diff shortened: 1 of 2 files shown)


src/decoder/zlib.rs
```diff
@@ -3,10 +3,64 @@ use super::{stream::FormatErrorInner, DecodingError, CHUNCK_BUFFER_SIZE};
...
+                    }
+                    Ok((in_consumed, out_consumed)) => {
+                        assert!(in_consumed <= input.len());
+                        (TINFLStatus::HasMoreOutput, in_consumed, out_consumed)
+                    }
+                    Err(fdeflate::DecompressionError::NotFDeflate) => {
+                        *self = Compressor::FullZlib(DecompressorOxide::new());
+
+                        assert_eq!(output_position, 0);
+                        self.decompress(input, output, output_position, end_of_input)
```
HeroicKatora, line 46: This looks wrong, the `input` is not necessarily at the start of a complete and valid `zlib` encoded stream. It is passed and conumsed in chunks. If anything, we'd need `fdeflate` to create an 'equivalent' stream header so that `DecompressorOxide` can be primed with any data that may have been contained in the already consumed header. 

That could be quite complex, as it includes:

* The zlib header
* The chunk state with header
* The last 64kB of data as a literal output, for potential backreferences
* _Optionally_ some adler32 checksum initialization, somehow

Seems complex but actually somewhat doable.

fintelia: The `fdeflate` crate is actually able to guarantee that it never consumes even a single byte of the stream before being sure that it can decode the whole thing. It does so by rejecting any deflate stream with more than one block (i.e. where BFINAL isn't set in the first block header) or whose first block describes a huffman codes it cannot support. Not sure the best way to document that guarantee.

HeroicKatora: Huh, interesting indeed. Well, a good comment should suffice? That clears up the purpose of `assert_eq!` as well.

HeroicKatora (approved):
I let this stew for a while but it looks perfectly fine to me with the comments.

PR #374, opened 2022-12-31
Refactor filtering and optimize adaptive filtering

fintelia:
Also add a dependency on `multiversion` to take advantage of AVX2 for adaptive filtering when available.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ crc32fast = "1.2.0"
 fdeflate = "0.2.0"
 flate2 = "1.0"
 miniz_oxide = "0.6.0"
+multiversion = "0.7.1"
 
 [dev-dependencies]
 criterion = "0.3.1"
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -54,6 +54,15 @@ impl Default for AdaptiveFilterType {
     }
 }
 
+fn filter_sub(a: u8, _: u8, _: u8) -> u8 {
+    a
+}
+fn filter_up(_: u8, b: u8, _: u8) -> u8 {
+    b
+}
+fn filter_avg(a: u8, b: u8, _: u8) -> u8 {
+    ((u16::from(a) + u16::from(b)) / 2) as u8
+}
 fn filter_paeth(a: u8, b: u8, c: u8) -> u8 {
     // This is an optimized version of the paeth filter from the PNG specification, proposed by
     // Luca Versari for [FPNGE](https://www.lucaversari.it/FJXL_and_FPNGE.pdf). It operates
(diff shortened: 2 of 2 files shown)


okaneco:
I was optimizing `sum_buffer` and I'm able to get better performance on `corpus-bench` only by changing that function than this PR. AVX2 might be a different story but the current computer I have access to doesn't support those instructions.

A comparison of the bench numbers for encoding on my computer. Make sure to `set_adaptive_filter` on the encoder.
```bash
$ cargo run --release --example corpus-bench -- ./tests/benches/
master         76 mps   0.09 GiB/s
this PR        84 mps   0.10 GiB/s
sum_buffer    148 mps   0.17 GiB/s
```

<details><summary>Optimized `sum_buffer` code</summary>

The `saturating_add` calls get optimized if you use an accumulator for the inner chunk loops. Not sure if this approach could help this PR.

```rust
fn sum_buffer(buf: &[u8]) -> u64 {
    const CHUNK_SIZE: usize = 32;

    let mut buf_chunks = buf.chunks_exact(CHUNK_SIZE);
    let mut sum = 0_u64;

    for chunk in &mut buf_chunks {
        // At most, `acc` can be `32 * (i8::MIN as u8) = 32 * 128 = 4096`.
        let mut acc = 0;
        for &b in chunk {
            acc += u64::from((b as i8).unsigned_abs());
        }
        sum = sum.saturating_add(acc);
    }

    let mut acc = 0;
    for &b in buf_chunks.remainder() {
        acc += u64::from((b as i8).unsigned_abs());
    }

    sum.saturating_add(acc)
}
```
</details>

Another thing I tried was keeping a running sum of each row's weights to return from `filter_internal` but doing that for `Avg` and `Paeth` degraded performance compared to summing the buffer after it was filtered. Keeping a running sum for `Sub` and `Up` was about the same or very slightly better performance.

fintelia:
Interesting! I just tried your patch on my machine and also saw a clear benefit. Tweaking the `filter` method to skip considering the NoFilter case (which only serves to make compression worse!) and doing the filtering in-place both both give small additional gains in performance.
```
$ cargo run --example corpus-bench --release qoi_benchmark_suite
master                  83 mps   0.28 GiB/s
this PR (no AVX2)       86 mps   0.29 GiB/s
this PR (AVX2)         162 mps   0.55 GiB/s
sum_buffer             215 mps   0.73 GiB/s
sum_buffer + filter    232 mps   0.79 GiB/s
```

<details>
  <summary>Better filter() implementation</summary>

```rust
AdaptiveFilterType::Adaptive => {
    let mut min_sum: u64 = u64::MAX;
    let mut filter_choice = FilterType::NoFilter;
    for &filter in [Sub, Up, Avg, Paeth].iter() {
        filter_internal(filter, bpp, len, previous, current, output);
        let sum = sum_buffer(&output);
        if sum <= min_sum {
            min_sum = sum;
            filter_choice = filter;
        }
    }

    if filter_choice != Paeth {
        filter_internal(filter_choice, bpp, len, previous, current, output);
    }
    filter_choice
}
```
</details>

fintelia:
Closing in favor of #375

okaneco:
I also see an increase with the improved `filter` :+1:

PR #376, opened 2023-01-01
Also roundtrip-fuzz with various bit depths

Shnatsel:
Expands on #370

Also fixes build of fuzzers that #370 broke

### DIFF
--- a/fuzz/fuzz_targets/roundtrip.rs
+++ b/fuzz/fuzz_targets/roundtrip.rs
@@ -34,11 +34,9 @@ fn encode_png(width: u8, filter: u8, compression: u8, color_type: u8, bit_depth:
     };
 
     // infer the rest of the parameters
-    let bytes_per_row = raw_row_length_from_width(bit_depth, color_type, width);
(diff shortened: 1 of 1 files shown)


fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -66,4 +68,19 @@ fn decode_png(data: &[u8]) -> (png::OutputInfo, Vec<u8>) {
     let info = reader.next_frame(&mut img_data).unwrap();
 
     (info, img_data)
+}
+
+// copied from the `png` codebase because it's pub(crate)
+fn raw_row_length_from_width(depth: BitDepth, color: ColorType, width: u32) -> usize {
+    let samples = width as usize * color.samples();
+    1 + match depth {
```
fintelia, line 76: The `1 +` is to account for the filter byte that PNG rows contain before the start of actual image data. It cancels out the `- 1` above.

Shnatsel: Added comment explaining it. Thanks!

Shnatsel:
Anything else I should do before merging or is this good to go?

PR #378, opened 2023-01-05
Optimize Paeth decoding

okaneco:
Possible alternative to #377

I was experimenting with the Paeth filter and it seems like the current version works great for encoding but not for decoding for some reason. Using `filter_paeth_decode` from this PR gave a considerable speed increase but using it for encoding degrades encoding performance. Another slight speed increase came from removing the `chunks_exact` implementation and iterator juggling in `unfilter` for Paeth.

I saw ~30% increase in mps decoding speed across several folders. The current status of #377 lands me at 52 mps for decoding.
```bash
$ cargo run --release --example corpus-bench -- ./tests/benches/
                Decode
         --------------------
master    46 mps   0.06 GiB/s
f7ad67    57 mps   0.07 GiB/s
295130    61 mps   0.07 GiB/s
```

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -54,6 +54,27 @@ impl Default for AdaptiveFilterType {
     }
 }
 
+fn filter_paeth_decode(a: u8, b: u8, c: u8) -> u8 {
+    // Decoding seems to optimize better with this algorithm
+    let p = i16::from(a) + i16::from(b) - i16::from(c);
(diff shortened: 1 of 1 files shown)


fintelia:
I think the distinction is probably that the current `filter_paeth` works extremely well if the code can be autovectorized, but is slower if not.

fintelia (approved):

PR #381, opened 2023-01-15
Add support for palette in fuzz roundtrip

ogarokpeter:
Add dummy palette generation in roundtrip fuzzing.

### DIFF
--- a/fuzz/fuzz_targets/roundtrip.rs
+++ b/fuzz/fuzz_targets/roundtrip.rs
@@ -4,8 +4,8 @@ use png::{FilterType, ColorType, BitDepth};
 #[macro_use] extern crate libfuzzer_sys;
 extern crate png;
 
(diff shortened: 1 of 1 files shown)


fintelia:
I think it would be better to have the fuzzing target take in the whole palette as a `Vec<u8>` rather than generating based on a provided length.

fintelia:
Much of the benefit of fuzzing comes from passing all sorts of weird and malformed inputs to the program to make sure they're all handled sensibly. The more we handcraft logic to assemble valid inputs, the less likely we'll find cases that we forget to address

ogarokpeter:
Ok, agree with your point.

I've implemented fuzzing for palette.

ogarokpeter:
Hello again!

Is there anything else to be done or is it good to go?

fintelia:
Looks good to me

PR #382, opened 2023-01-19
Optimize Sub, Avg, and Paeth decoding with SIMD Within A Register

okaneco:
_SWAR: SIMD Within A Register
http://aggregate.org/SWAR/over.html
http://aggregate.org/MAGIC/#Average%20of%20Integers_

- Calculate average in `Avg` encoding filter with bitwise operations that don't widen to u16
- Use SWAR for `Sub` and `Avg` filters to filter one pixel at a time for 3, 4, 6, and 8 bpp images

The idea for SWAR is to treat regular integer types like `u32|u64` as a SIMD
register for arbitrary smaller-width bit lanes. We can calculate 3 bytes per
pixel and 4bpp pixels within a `u32`, 6bpp and 8bpp pixels within a `u64`.

It's not an n-times speedup based on the number of lanes because SWAR may
require some overhead of maintaining the lanes depending on the operation
(shifts, for instance). This is evidenced by masking operations during
calculation to correct or prevent lanes from overlapping. For the wrapping
addition operation here, 7-bit arithmetic is calculated to prevent a carry
digit affecting neighboring lanes. After the addition, a correcting XOR addition
is done to repair the most significant bit and arrive at the correct answer for
8-bit wrapping unsigned arithmetic.

Most PNGs seem to be 3 or 4bpp and SWAR seems to improve decoding time over the obvious for-each loop. There's a somewhat unavoidable (to me) level of ugliness and obfuscation in the implementation, but I've tried to leave enough comments and break up the commits.

---

*Edit (2023/01/21):*
This ended up not using bitwise operations for optimizing decoding but did use concepts from the initial commits to improve the `Sub`, `Avg`, and `Paeth` decoding filters.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -278,29 +278,37 @@ pub(crate) fn unfilter(
                 };
             }
 
-            avg_tail!(avg_tail_2, 2);
             avg_tail!(avg_tail_1, 1);
+            avg_tail!(avg_tail_2, 2);
 
             match tbpp {
-                BytesPerPixel::Eight => {
-                    assert!(len > 7);
-                    let mut lprev = u64::from_ne_bytes([
-                        current[0], current[1], current[2], current[3], current[4], current[5],
-                        current[6], current[7],
-                    ]);
-                    for (chunk, above) in current[8..]
-                        .chunks_exact_mut(8)
-                        .zip(previous[8..].chunks_exact(8))
+                BytesPerPixel::One => avg_tail_1(current, previous),
+                BytesPerPixel::Two => avg_tail_2(current, previous),
+                BytesPerPixel::Three => {
+                    assert!(len > 2);
+                    let mut lprev = u32::from_ne_bytes([current[0], current[1], current[2], 0]);
+                    for (chunk, above) in current[3..]
+                        .chunks_exact_mut(3)
+                        .zip(previous[3..].chunks_exact(3))
                     {
-                        let pprev = u64::from_ne_bytes([
-                            above[0], above[1], above[2], above[3], above[4], above[5], above[6],
-                            above[7],
-                        ]);
-                        let pcurrent = u64::from_ne_bytes([
-                            chunk[0], chunk[1], chunk[2], chunk[3], chunk[4], chunk[5], chunk[6],
-                            chunk[7],
-                        ]);
-                        let new_chunk = swar_add_u64(pcurrent, swar_avg_u64(pprev, lprev));
+                        let pprev = u32::from_ne_bytes([above[0], above[1], above[2], 0]);
+                        let pcurrent = u32::from_ne_bytes([chunk[0], chunk[1], chunk[2], 0]);
+                        let new_chunk = swar_add_u32(pcurrent, swar_avg_u32(pprev, lprev));
+                        chunk.copy_from_slice(&new_chunk.to_ne_bytes()[..3]);
+                        lprev = new_chunk;
+                    }
+                }
+                BytesPerPixel::Four => {
+                    assert!(len > 3);
+                    let mut lprev =
+                        u32::from_ne_bytes([current[0], current[1], current[2], current[3]]);
+                    for (chunk, above) in current[4..]
+                        .chunks_exact_mut(4)
+                        .zip(previous[4..].chunks_exact(4))
+                    {
+                        let pprev = u32::from_ne_bytes([above[0], above[1], above[2], above[3]]);
+                        let pcurrent = u32::from_ne_bytes([chunk[0], chunk[1], chunk[2], chunk[3]]);
+                        let new_chunk = swar_add_u32(pcurrent, swar_avg_u32(pprev, lprev));
                         chunk.copy_from_slice(&new_chunk.to_ne_bytes());
                         lprev = new_chunk;
                     }
@@ -326,37 +334,29 @@ pub(crate) fn unfilter(
                         lprev = new_chunk;
                     }
                 }
-                BytesPerPixel::Four => {
-                    assert!(len > 3);
-                    let mut lprev =
-                        u32::from_ne_bytes([current[0], current[1], current[2], current[3]]);
-                    for (chunk, above) in current[4..]
-                        .chunks_exact_mut(4)
-                        .zip(previous[4..].chunks_exact(4))
+                BytesPerPixel::Eight => {
+                    assert!(len > 7);
+                    let mut lprev = u64::from_ne_bytes([
+                        current[0], current[1], current[2], current[3], current[4], current[5],
+                        current[6], current[7],
+                    ]);
+                    for (chunk, above) in current[8..]
+                        .chunks_exact_mut(8)
+                        .zip(previous[8..].chunks_exact(8))
                     {
-                        let pprev = u32::from_ne_bytes([above[0], above[1], above[2], above[3]]);
-                        let pcurrent = u32::from_ne_bytes([chunk[0], chunk[1], chunk[2], chunk[3]]);
-                        let new_chunk = swar_add_u32(pcurrent, swar_avg_u32(pprev, lprev));
+                        let pprev = u64::from_ne_bytes([
+                            above[0], above[1], above[2], above[3], above[4], above[5], above[6],
+                            above[7],
+                        ]);
+                        let pcurrent = u64::from_ne_bytes([
+                            chunk[0], chunk[1], chunk[2], chunk[3], chunk[4], chunk[5], chunk[6],
+                            chunk[7],
+                        ]);
+                        let new_chunk = swar_add_u64(pcurrent, swar_avg_u64(pprev, lprev));
                         chunk.copy_from_slice(&new_chunk.to_ne_bytes());
                         lprev = new_chunk;
                     }
                 }
-                BytesPerPixel::Three => {
-                    assert!(len > 2);
-                    let mut lprev = u32::from_ne_bytes([current[0], current[1], current[2], 0]);
-                    for (chunk, above) in current[3..]
-                        .chunks_exact_mut(3)
-                        .zip(previous[3..].chunks_exact(3))
-                    {
-                        let pprev = u32::from_ne_bytes([above[0], above[1], above[2], 0]);
-                        let pcurrent = u32::from_ne_bytes([chunk[0], chunk[1], chunk[2], 0]);
-                        let new_chunk = swar_add_u32(pcurrent, swar_avg_u32(pprev, lprev));
-                        chunk.copy_from_slice(&new_chunk.to_ne_bytes()[..3]);
-                        lprev = new_chunk;
-                    }
-                }
-                BytesPerPixel::Two => avg_tail_2(current, previous),
-                BytesPerPixel::One => avg_tail_1(current, previous),
             }
 
             Ok(())
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -127,70 +127,72 @@ pub(crate) fn unfilter(
     let bpp = tbpp.into_usize();
     let len = current.len();
 
-    // SIMD Within a Register, see "Unpartitioned Operations With Correction Code"
-    // http://aggregate.org/SWAR/over.html
-    fn swar_add_u32(x: u32, y: u32) -> u32 {
-        // Do 7-bit addition so there's no carry over the most significant bit
-        let n = (x & 0x7f7f7f7f) + (y & 0x7f7f7f7f);
-        // 1-bit parity/XOR addition to fill in the missing MSB
-        n ^ (x ^ y) & 0x80808080
-    }
-
-    fn swar_add_u64(x: u64, y: u64) -> u64 {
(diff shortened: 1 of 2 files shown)


okaneco:
Some non-exhaustive numbers from corpus bench, comment out the Adaptive filter on L69 and then modify the `FilterType` to `Sub` and `Avg`.

I copied the plant folder and used `imagemagick` to convert the copied files to `-depth 16` since I don't know of a source for larger bit depth images. This is the `../qoi_benchmark_suite/images/textures_plants/8bpp` folder.

I saw consistent small improvements with the encoding time for Avg but it was folder dependent. A larger example is the `screenshot_web` folder.

<details>
<summary>Sub benches</summary>

```
Sub MASTER
Directory                                                       Encode                    Decode
---------                                                --------------------      --------------------
../qoi_benchmark_suite/images/textures_plants            160 mps   0.60 GiB/s       54 mps   0.20 GiB/s
../qoi_benchmark_suite/images/textures_plants/8bpp        68 mps   0.51 GiB/s       28 mps   0.21 GiB/s
../qoi_benchmark_suite/images/textures_pk02               79 mps   0.30 GiB/s       40 mps   0.15 GiB/s
../qoi_benchmark_suite/images/textures_pk01              104 mps   0.39 GiB/s       43 mps   0.16 GiB/s
../qoi_benchmark_suite/images/textures_pk                 99 mps   0.29 GiB/s       45 mps   0.13 GiB/s
../qoi_benchmark_suite/images/textures_photo              86 mps   0.24 GiB/s       54 mps   0.15 GiB/s
../qoi_benchmark_suite/images/screenshot_web             270 mps   1.01 GiB/s       65 mps   0.24 GiB/s

Sub SWAR
Directory                                                       Encode                    Decode
---------                                                --------------------      --------------------
../qoi_benchmark_suite/images/textures_plants            161 mps   0.60 GiB/s       64 mps   0.24 GiB/s
../qoi_benchmark_suite/images/textures_plants/8bpp        68 mps   0.51 GiB/s       34 mps   0.25 GiB/s
../qoi_benchmark_suite/images/textures_pk02               79 mps   0.30 GiB/s       45 mps   0.17 GiB/s
../qoi_benchmark_suite/images/textures_pk01              106 mps   0.40 GiB/s       49 mps   0.18 GiB/s
../qoi_benchmark_suite/images/textures_pk                102 mps   0.29 GiB/s       48 mps   0.14 GiB/s
../qoi_benchmark_suite/images/textures_photo              85 mps   0.24 GiB/s       58 mps   0.16 GiB/s
../qoi_benchmark_suite/images/screenshot_web             270 mps   1.01 GiB/s       78 mps   0.29 GiB/s
```
</details>

<details>
<summary>Avg benches</summary>

```
Avg MASTER
Directory                                                       Encode                    Decode
---------                                                --------------------      --------------------
../qoi_benchmark_suite/images/textures_plants            152 mps   0.57 GiB/s       52 mps   0.20 GiB/s
../qoi_benchmark_suite/images/textures_plants/8bpp        66 mps   0.49 GiB/s       29 mps   0.22 GiB/s
../qoi_benchmark_suite/images/textures_pk02               78 mps   0.29 GiB/s       39 mps   0.15 GiB/s
../qoi_benchmark_suite/images/textures_pk01               97 mps   0.36 GiB/s       41 mps   0.15 GiB/s
../qoi_benchmark_suite/images/textures_pk                103 mps   0.30 GiB/s       46 mps   0.13 GiB/s
../qoi_benchmark_suite/images/textures_photo              88 mps   0.25 GiB/s       53 mps   0.15 GiB/s
../qoi_benchmark_suite/images/screenshot_web             237 mps   0.89 GiB/s       61 mps   0.23 GiB/s

Avg SWAR
Directory                                                       Encode                    Decode
---------                                                --------------------      --------------------
../qoi_benchmark_suite/images/textures_plants            154 mps   0.57 GiB/s       60 mps   0.22 GiB/s
../qoi_benchmark_suite/images/textures_plants/8bpp        67 mps   0.50 GiB/s       32 mps   0.25 GiB/s
../qoi_benchmark_suite/images/textures_pk02               79 mps   0.30 GiB/s       44 mps   0.16 GiB/s
../qoi_benchmark_suite/images/textures_pk01               97 mps   0.36 GiB/s       46 mps   0.17 GiB/s
../qoi_benchmark_suite/images/textures_pk                104 mps   0.30 GiB/s       46 mps   0.13 GiB/s
../qoi_benchmark_suite/images/textures_photo              88 mps   0.25 GiB/s       54 mps   0.15 GiB/s
../qoi_benchmark_suite/images/screenshot_web             243 mps   0.91 GiB/s       71 mps   0.27 GiB/s
```
</details>

---

An enhancement for the `cargo bench` could be an image of each filter type and then a 3/4 and 6/8 bpp to track changes more specifically by only running certain benches with the regex-filter for what you're tuning. It's hard to have the patience for letting `criterion` run when the corpus bench is easier to iterate on.

src/filter.rs
```diff
@@ -132,12 +132,100 @@ pub(crate) fn unfilter(
...
+
     match filter {
         NoFilter => Ok(()),
         Sub => {
             let current = &mut current[..len];
-            for i in bpp..len {
-                current[i] = current[i].wrapping_add(current[i - bpp]);
+            if bpp > len {
+                return Err("Filtering failed: bytes per pixel is greater than length of row");
+            }
```
okaneco, line 169: This is a new error path that makes it possible for `Sub` to return an error that would have been `Ok` before.

I'm not sure of the strategy for handling `bpp > len` because `Avg` and `Paeth` already return it. Should it be hoisted out of the match and apply to all filters, or is this fine?

src/filter.rs
```diff
@@ -132,12 +132,100 @@ pub(crate) fn unfilter(
...
+
+            match tbpp {
+                BytesPerPixel::Three => {
+                    assert!(len > 2);
```
okaneco, line 173: I couldn't get the compiler to realize indexing was in bounds without adding these asserts. Information doesn't seem to stick from `tbpp.into_usize()`.

src/filter.rs
fintelia: Honestly, unfilter probably shouldn't return a Result in the first place. Assertions are fine for code clarity or to inform the optimizer, but the method is only called from a single place with arguments that should have already been validated:

https://github.com/image-rs/image-png/blob/56e3ea6000b825c90e02e5df743bb0479a686294/src/decoder/mod.rs#L777

Like staring at that line, it seems pretty silly to have `require_length` and all the code around validating `previous` and `current` have the same length, when they are both guaranteed to be exactly `rowlen-1` bytes long!

src/filter.rs
```diff
@@ -190,18 +278,83 @@ pub(crate) fn unfilter(
...
-            avg_tail!(avg_tail_3, 3);
             avg_tail!(avg_tail_2, 2);
             avg_tail!(avg_tail_1, 1);
 
             match tbpp {
-                BytesPerPixel::Eight => avg_tail_8(current, previous),
-                BytesPerPixel::Six => avg_tail_6(current, previous),
-                BytesPerPixel::Four => avg_tail_4(current, previous),
-                BytesPerPixel::Three => avg_tail_3(current, previous),
+                BytesPerPixel::Eight => {
```
fintelia, line 285: Could you reverse the order of these so it matches `Sub`? We might as well make them consistent if the code is changing anyway

fintelia:
I did some performance experiments. On my machine, iterating with chunks but computing the filters component-wise gives a similar/better speedup.
```rust
// Example of avg unfilter:

assert!(len > 3);
let mut lprev = [current[0], current[1], current[2], current[3]];
for (chunk, above) in current[4..]
    .chunks_exact_mut(4)
    .zip(previous[4..].chunks_exact(4))
{
    let new_chunk = [
        chunk[0].wrapping_add(((above[0] as u16 + lprev[0] as u16) / 2) as u8),
        chunk[1].wrapping_add(((above[1] as u16 + lprev[1] as u16) / 2) as u8),
        chunk[2].wrapping_add(((above[2] as u16 + lprev[2] as u16) / 2) as u8),
        chunk[3].wrapping_add(((above[3] as u16 + lprev[3] as u16) / 2) as u8),
    ];
    *TryInto::<&mut [u8; 4]>::try_into(chunk).unwrap() = new_chunk;
    lprev = new_chunk;
}
```

src/filter.rs
okaneco: Sure, I've reordered the match arms.

src/filter.rs
okaneco: Oh, I kept losing track of the caller for `unfilter` with all the context switches between `decoder::mod` and `decoder::filter` so I wasn't sure what data `unfilter` was getting.

The validation of `require_length` is completely unnecessary then. I wasn't sure if the `bpp > len` check was anywhere else but it seems like it should definitely live at least one layer above this function in `next_raw_interlaced_row`.

I've made it so the function no longer returns a result and the error check is done in the caller function.

okaneco:
I'll give that a try. Do you have a preference for where to import TryInto?

okaneco:
`TryInto` seems like a better way to handle the shuffling of `to|from_ne_bytes`.

I'm seeing an improvement too. Is it the same for `Sub` or just for `Avg`?

fintelia:
The results were similar with `Sub` and `Paeth`. I also tried adding assertions in the loop body which produced nicer code but no difference in performance:
```rust
assert_eq!(chunk.len(), 4);
assert_eq!(above.len(), 4);
assert_eq!(lprev.len(), 4); // no-op since type is [u8;4]
```
With that change the compiler actually outputs quite reasonable SIMD code:
```
pxor   %xmm0,%xmm0             // set xmm0 = 0
punpcklbw %xmm0,%xmm3          // expand xmm3 from 4*u8 -> 4*u16
movd   (%rdx,%rcx,4),%xmm1     // load chunk into xmm1
movd   (%rsi,%rcx,4),%xmm2     // load above into xmm2
punpcklbw %xmm0,%xmm2          // expand xmm2 from 4*u8 -> 4*u16
paddw  %xmm3,%xmm2             // add chunk and above
psrlw  $0x1,%xmm2              // divide by 2
packuswb %xmm2,%xmm2           // pack back 4*u16 -> 4*u8
paddb  %xmm1,%xmm2             // wrapping add chunk and average
movd   %xmm2,(%rdx,%rcx,4)     // store result
mov    %rdi,%rcx
movdqa %xmm2,%xmm3             // set lprev = chunk
cmp    %rdi,%rax
je     1c964 <_ZN3png6filter8unfilter17h2f777b205fb3a0a2E+0xe64>

File: src/filter.rs [hunk 4]
```diff
@@ -190,48 +242,208 @@ pub(crate) fn unfilter(
...
+                            chunk[0]
+                                .wrapping_add(filter_paeth_decode(a_bpp[0], b_bpp[0], c_bpp[0])),
+                            chunk[1]
+                                .wrapping_add(filter_paeth_decode(a_bpp[1], b_bpp[1], c_bpp[1])),
+                            chunk[2]
+                                .wrapping_add(filter_paeth_decode(a_bpp[2], b_bpp[2], c_bpp[2])),
+                        ];
+                        *TryInto::<&mut [u8; 3]>::try_into(chunk).unwrap() = new_chunk;
+                        a_bpp = new_chunk;
+                        c_bpp = TryInto::<[u8; 3]>::try_into(b_bpp).unwrap();
```
fintelia, line 352: I think the type inference should work if you do `c_bpp = b_bpp.try_into().unwrap()`.

okaneco:
Awesome and surprising discoveries with all this :laughing:

I never expected the performance to improve that much with such simple changes. I guess "thinking in SWAR" can sometimes help the compiler optimize even further when dealing with these small byte arrays.

okaneco:
I refactored the code to remove all the assertions and slicing. There's no noticeable performance impact but it's simpler code with no manual indexing. This also meant `len` and `bpp` weren't necessary for every branch so they could be moved to the two arms where they're used. A diff of the last two commits, ~400 line difference in assembly output and 44 less lines of code https://rust.godbolt.org/z/3n9Y863Mr

For `Avg`, I removed the macro because expanding the `One` and `Two` arms occupied fewer lines than the macro. 

For `Paeth`, expanding `One` and `Two` dropped the instruction count a bit and unified the logic so it seemed worth doing. This got rid of the iterators and some manual indexing only used in that arm.

For `Sub`, expanding `One` and `Two` seemed to just add more overhead so I left them in a combined match arm.

The current format of all the repeated arms seems like a candidate for a macro but I'm not sure if it's possible to expand repeated array indexing in a clean way. I'm not a macro expert though, I mostly avoid writing them unless they're very simple.

HeroicKatora (commented):
I'm very content with the code itself. Can you add comments for each of the changed tricks, to paint a picture of how they worked before and why the current choice is faster. That can be, in essence, the comments you explained in the PR. Having them in code hopefully helps anyone trying to improve further avoid trying that prior state of code for instance.

okaneco:
I've added an explanation commit.

It's out of scope for this PR but now I'm curious if this technique could be used in `filter` with unrolled components and `[0_u8; N]` arrays with some constant power of 2 to get any more speed out of encoding.

fintelia (approved):

PR #386, opened 2023-03-02
Add option to ignore Adler32 checksum in ZlibStream, set ignore as default

### LINKED ISSUE #254: Don't verify adler32 checksum by default
For lightly compressed images adler32 checksum calculation [accounts for 50% of zlib decompression time](https://github.com/Frommi/miniz_oxide/issues/76), and the efforts to optimize adler32 implementation itself are largely exhausted.

adler32 allows verification of the final uncompressed bitstream; however, the input file integrity is already verified by crc32 in the PNG format. This makes adler32 verification redundant, since the only things it can detect are:

 1. A bug in the zlib decompression implementation
 1. File corruption so subtle that the outer crc32 missed it

Both of these scenarios are so rare that they're not worth the extra runtime overhead in the default configuration.

Disclaimer: my understanding is based on the summary provided [here](https://stackoverflow.com/questions/33914640/adler-32-checksums-in-a-png-file), I'm not an expert on the PNG format.

### LINKED ISSUE #347: Option to disable CRC check
I need to disable the CRC check for PNG (and any other image) files.  My use case is I'm using `image-rs` to load images (PNG, JPEG, BMP, other) to calculate image fuzzy hashes that can be used to identify variations on images commonly used in phishing/malware campaigns.  I don't need the CRC or the rest of the image to be technically correct. Many applications make a best effort to present images to users, even if they're obviously malformed.  I need to be able to do the same or else a bad actor need only tweak something like the CRC slightly and I won't be able to process the file. 

I found that there is an option to disable the CRC check for `image-png` through the rust config, but it requires enabling the `fuzzing` configuration: https://github.com/image-rs/image-png/blob/master/src/decoder/stream.rs#L24-L28 

It is tempting to use the `fuzzing` configuration for `image-png` in my application, but I'm concerned that the `fuzzing` configuration may be used for other purposes as well in the future and I don't want any surprises.

I guess my question is -- can we get a (documented) configuration option or some other function to disable the CRC check?

okaneco:
Add `ignore_adler32` flag to `ZlibStream`
Disable computing and checking the Adler-32 checksum by default
Reset `ZlibStream` where possible instead of instantiating new structs
Add `ignore_crc` flag to DecodeOptions

I'm not sure what to do about the user interface for this. I added options to the `StreamingDecoder`. The coupling of setting the option is clunky to sync with the multiple levels of structs from `ZlibStream` upwards.

I see a 3.5%-8% improvement on the benchmarks.

Closes #254 and closes #347

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -384,6 +384,7 @@ impl From<TextDecodingError> for DecodingError {
 #[derive(Clone)]
 pub struct DecodeOptions {
     ignore_adler32: bool,
+    ignore_crc: bool,
     ignore_text_chunk: bool,
 }
 
@@ -391,6 +392,7 @@ impl Default for DecodeOptions {
     fn default() -> Self {
         Self {
             ignore_adler32: true,
+            ignore_crc: false,
             ignore_text_chunk: false,
         }
     }
@@ -402,6 +404,20 @@ impl DecodeOptions {
     /// Defaults to `true`.
     pub fn set_ignore_adler32(&mut self, ignore_adler32: bool) {
         self.ignore_adler32 = ignore_adler32;
+    }
+
+    /// When set, the decoder will not compute and verify the CRC code.
+    ///
+    /// Defaults to `false`.
+    pub fn set_ignore_crc(&mut self, ignore_crc: bool) {
+        self.ignore_crc = ignore_crc;
+    }
+
+    /// Flag to ignore computing and verifying the Adler-32 checksum and CRC
+    /// code.
+    pub fn set_ignore_checksums(&mut self, ignore_checksums: bool) {
+        self.ignore_adler32 = ignore_checksums;
+        self.ignore_crc = ignore_checksums;
     }
 
     /// Ignore text chunks while decoding.
@@ -516,6 +532,14 @@ impl StreamingDecoder {
         self.inflater.set_ignore_adler32(ignore_adler32)
     }
 
+    /// Set whether to compute and verify the Adler-32 checksum during
+    /// decompression. Return `true` if the flag was successfully set.
+    ///
+    /// The decoder defaults to `false`.
+    pub fn set_ignore_crc(&mut self, ignore_crc: bool) {
+        self.decode_options.set_ignore_crc(ignore_crc)
+    }
+
     /// Low level StreamingDecoder interface.
     ///
     /// Allows to stream partial data to the encoder. Returns a tuple containing the bytes that have
@@ -593,8 +617,10 @@ impl StreamingDecoder {
                             return Ok((0, Decoded::ImageDataFlushed));
                         }
                         self.current_chunk.type_ = type_str;
-                        self.current_chunk.crc.reset();
-                        self.current_chunk.crc.update(&type_str.0);
+                        if !self.decode_options.ignore_crc {
+                            self.current_chunk.crc.reset();
+                            self.current_chunk.crc.update(&type_str.0);
+                        }
                         self.current_chunk.remaining = length;
                         self.apng_seq_handled = false;
                         self.current_chunk.raw_bytes.clear();
@@ -602,8 +628,16 @@ impl StreamingDecoder {
                         Ok((1, Decoded::ChunkBegin(length, type_str)))
                     }
                     Crc(type_str) => {
-                        let sum = self.current_chunk.crc.clone().finalize();
-                        if CHECKSUM_DISABLED || val == sum {
+                        // If ignore_crc is set, do not calculate CRC. We set
+                        // sum=val so that it short-circuits to true in the next
+                        // if-statement block
+                        let sum = if self.decode_options.ignore_crc {
+                            val
+                        } else {
+                            self.current_chunk.crc.clone().finalize()
+                        };
+
+                        if val == sum || CHECKSUM_DISABLED {
                             self.state = Some(State::U32(U32Value::Length));
                             if type_str == IEND {
                                 Ok((1, Decoded::ImageEnd))
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -380,13 +380,33 @@ impl From<TextDecodingError> for DecodingError {
     }
 }
 
-/// Decode config
-#[derive(Default)]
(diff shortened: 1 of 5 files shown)


src/decoder/zlib.rs
```diff
@@ -98,17 +112,34 @@ impl ZlibStream {
...
             out_buffer: vec![0; 2 * CHUNCK_BUFFER_SIZE],
             out_pos: 0,
+            ignore_adler32: true,
         }
     }
 
     pub(crate) fn reset(&mut self) {
         self.started = false;
         self.in_buffer.clear();
+        self.in_pos = 0;
```
okaneco, line 122: This seemed like an oversight but I'm not sure since nothing fails with this reset or not. I'm not clear on what the documentation means:
```rust
    /// The logical start of the `in_buffer`.
    in_pos: usize
```

L172 is as follows, should this be `<=`?
```rust
        if self.in_buffer.len() == self.in_pos {
            self.in_buffer.clear();
            self.in_pos = 0;
        }
```

src/decoder/stream.rs
```diff
@@ -467,6 +471,20 @@ impl StreamingDecoder {
...
+    }
+
+    /// Set whether to compute and verify the Adler-32 checksum during
+    /// decompression. The decoder defaults to `true`.
+    ///
+    /// This flag cannot be modified after decompression has started until the
+    /// [StreamingDecoder] is reset.
+    pub fn set_ignore_adler32_checksum(&mut self, flag: bool) {
+        self.inflater.set_ignore_adler32(flag);
+    }
```
HeroicKatora, line 486: Regarding API, that's fine but there should be similar setters for `Decoder`, in its `decoder: DecodeConfig` attribute. In a more forward looking note, in `gif` we switched the structus so that the reader is supplied after the `Decoder` is configured. (The struct has been renamed to [`DecodeOptions`](https://docs.rs/gif/latest/gif/struct.DecodeOptions.html) afterwards). In that case, the api could be 

* `StreamingDecoder::set_options(&mut self, _: &DecodeOptions)`.

We should probably switch to a similar API in the next major version.

src/decoder/zlib.rs
HeroicKatora: Huh. That seems plausible but I can't really say either. I don't think the reset code path is tested as exhaustive as it could be.

fintelia:
The `miniz_oxide` crate can optionally use `simd-adler32`. Would you mind benchmarking with that enabled and see how performance compares?

okaneco:
`decode/kodim07.png` is 5% faster than `simd-adler32`, `decode/Transparency.png` is 8% faster. All the other decode benches are ~1% faster with the checksum disabled than the `simd` feature.

okaneco:
On the corpus bench, the numbers are about equal.

Should this PR include upgrading to `miniz_oxide 0.7` and enabling the `simd-adler32` feature? The MSRV is `1.50`.

src/decoder/stream.rs
okaneco: I noticed that struct but it's `pub(crate)`, should that be made public? And should I rename it to match the `gif` API?

HeroicKatora:
I'd be fine with it.

oyvindln:
The only reason it's not enabled by default in miniz_oxide is that uses unsafe, which it has to to use SIMD directly.

fintelia:
I think the unsafe is justified given that it is a self contained crate and there's an enormous performance difference.

It occurred to me that the reason corpus-bench didn't show a difference is because it defaults to fast compression, which is then decoded using `fdeflate` (which unconditionally checks the adler32). If you run with `--speed default` the decode path should fallback to miniz_oxide instead

fintelia:
It occurs to me that PNGs also have CRC32 checksums. Is there any reason you'd want to check one and not the other? It may make more sense to have them both controlled by the same toggle

okaneco:
First pass at the API for this. The `DecodeOptions` struct was duplicated in the `Decoder` and `StreamingDecoder` so I removed it from `Decoder`.

Handling the Adler32 flag is a bit awkward because it lives on the `ZlibStream`. I currently have it configured so that the Adler flag can't be modified after the decompression has started.

fintelia:
I did some benchmarking on my machine. For our benchmark suite, I saw no difference except for Transparency.png which is a weird case (it is a ~3000 _byte_ file with a 122x compression ratio. Decoding takes like 80 microseconds). On `corpus-bench --speed default`:
```
main branch:       109 mps  0.370 GiB/s
this PR:           135 mps  0.457 GiB/s
miniz 0.7 + simd:  135 mps  0.457 GiB/s
```

I'm in favor of merging this because it is a commonly requested feature and doesn't degrade performance, but for most users I don't think there will be much difference compared to just using the simd adler32 implementation.

okaneco:
I removed a FIXME I left while refactoring code.

I noticed that it might make sense to relocate `Limits` to the `DecodeOptions` struct and remove it from the `Reader`. A branch of what that looks like is here, seemed semi-out of scope and in-scope for this PR. It removes a todo/fixme in `reserve_current_chunk` and `parse_iccp` but I'm not sure if it's wanted.
https://github.com/okaneco/image-png/commit/30e452887dd5eea55027ee42cb44b8e09cc1c50e

More work would need to be done to graft `Limits` onto the ZlibStream and account for limit exceeded errors there.

fintelia (approved):

Shnatsel:
> Also, PNGs have CRC32 checksums in addition to adler32. Is there any reason you'd want to check one and not the other? It may make more sense to have them both controlled by the same toggle

Actually yes. You only ever need one checksum to check file integrity - it's crc32 that covers the entire file. The Adler32 checksum is just there as dead weight because it was part of the underlying file format and PNG just reused it.

Adler32 covers the decompressed Zlib data before filtering etc. So it can be used to check that the Zlib decoder is functioning correctly, but nothing more. So it should be disabled always, while crc32 should be optional.

@fintelia this is why verifying adler32 unconditionally is not a good default for `fdeflate`

fintelia:
Well, a bad adler32 also means that the PNG is technically corrupt. But the main reason I haven't got around to making the checksum optional for fdeflate is because I haven't be able to measure any perf difference from doing so

PR #388, opened 2023-03-05
Small code cleanup

fintelia:
A careful read of the match statement reveals that `next_interlaced_row` always returns same thing as it got from `next_interlaced_row_impl`.


fintelia:
I'm working on a larger refactor of this code. Closing this to avoid conflicts

PR #389, opened 2023-03-09
Refactor Reader implementation

fintelia:
This inlines a few methods that are only called in one place and otherwise tries to make the code cleaner / easier to understand. The current complexity makes it difficult to make optimizations and improvements

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -196,7 +196,19 @@ impl<R: Read> Decoder<R> {
             .decoder
             .set_decode_config(self.decode_config);
 
-        let mut reader = Reader::new(self.read_decoder, self.transform, self.limits);
+        let mut reader = Reader {
+            decoder: self.read_decoder,
+            bpp: BytesPerPixel::One,
+            subframe: SubframeInfo::not_yet_init(),
+            fctl_read: 0,
+            next_frame: SubframeIdx::Initial,
+            prev: Vec::new(),
+            current: Vec::new(),
+            scan_start: 0,
+            transform: self.transform,
+            processed: Vec::new(),
+            limits: self.limits,
+        };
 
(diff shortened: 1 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -760,74 +707,55 @@ impl<R: Read> Reader<R> {
...
+                    if !self.current.is_empty() {
+                        return Err(DecodingError::Format(
+                            FormatErrorInner::UnexpectedEndOfChunk.into(),
+                        ));
+                    } else {
                         return Err(DecodingError::Format(
-                            FormatErrorInner::UnknownFilterMethod(row[0]).into(),
+                            FormatErrorInner::NoMoreImageData.into(),
                         ));
                     }
```
okaneco, line 739: Maybe more clear with the arms reversed.

```rust
return if self.current.is_empty() {
    Err(DecodingError::Format(
        FormatErrorInner::NoMoreImageData.into(),
    ));
} else {
    Err(DecodingError::Format(
        FormatErrorInner::UnexpectedEndOfChunk.into(),
    ));
};
```

src/decoder/mod.rs
```diff
@@ -556,42 +543,36 @@ impl<R: Read> Reader<R> {
...
 
     /// Fetch the next interlaced row and filter it according to our own transformations.
     fn next_interlaced_row_impl(&mut self) -> Result<Option<InterlacedRow>, DecodingError> {
         use crate::common::ColorType::*;
         let transform = self.transform;
 
+        let (rowlen, interlace) = match self.next_pass() {
+            Some((rowlen, interlace)) => (rowlen, interlace),
+            None => return Ok(None),
+        };
```
okaneco, line 557: After 1.65, I think this can be a `let-else`.

HeroicKatora (approved):

PR #391, opened 2023-03-18
Release 0.17.8-rc and call for testing

fintelia:
This prepares a preview release to enable broader testing before publishing.

There's been a lot of phenomenal performance improvements to this crate since the last release, including a complete rewrite of the zlib compression/decompression logic and major changes filtering and unfiltering. Further, this was all possible without any semver breaking changes. This introduces risks that we've introduced bugs or regressions which would get immediately picked up by downstream users once released.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,18 @@
 ## Unreleased
 
+## 0.17.8
+
+* Increased MSRV to 1.57.0.
+* Substantially optimized encoding and decoding:
+  - Autovectorize filtering and unfiltering.
+  - Make the "fast" compression preset use fdeflate.
+  - Switch decompression to always use fdeflate.
+  - Updated to miniz_oxide 0.7.
+  - Added an option to ignore checksums.
+* Added corpus-bench example which measures the compression ratio and time to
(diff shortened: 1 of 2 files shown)


Shnatsel:
Just so you know, I'm about to do a test run comparing the decoding results between 0.17.7 and 0.17.8 on my corpus of ~30,000 real-world PNGs. I'll let you know about the results in a day or so and file issues on the bug tracker.

Shnatsel:
I scraped another 58000 images and ran the test on them too because I could. No regressions on the total corpus of ~80,000 images. I filed #392 but that's actually a fix, not a regression.

So this is good to go!

Shnatsel:
I can help write the announcement too. I've been doing them for zune, and [did one for the PNG crate earlier](https://redd.it/hgrum2). If you want me to write this one too, just let me know and I'll do it.

fintelia:
@Shnatsel it would be awesome if you could write a release announcement for v0.17.8 once it is out

Shnatsel:
Sure thing!

It looks ready to ship to me, stability-wise. Is there anything else you want to tackle before release?

PR #394, opened 2023-04-03
Release 0.17.8

fintelia:


Shnatsel:
How's this for the release announcement? https://hackmd.io/@xmxSgN6MSImJPcY4MOKfjA/H1C5TYO-2

fintelia:
@Shnatsel  That looks quite good! A few minor points:
* It might be worth mentioning that the main `image` crate uses `png` for all PNG handling (as I think most users depend on this one only indirectly through `image`)
* I actually measured a 75% improvement in decoding performance since 0.17.7 on the QOI benchmark suite. But it'll depend heavily on the individual image file.
* A bit pedantic, but fdeflate is a zlib implementation, not gzip. The two are closely related wrapper formats for deflate streams but zlib is the one used by PNGs.

I'll give @HeroicKatora a bit more time to offer feedback, but otherwise will go ahead and publish the release tomorrow.

Shnatsel:
I went by `cargo bench` results in the `image-png` repo. One image improved by 75% but I assumed it has hit the fast path that let it go through `fdeflate`. But if you're sure that's not the case, I'll go ahead and adjust the language accordingly.

Shnatsel:
I see the release is up. Do you want to post the announcement, or should I?

Keep in mind [the best times to post on /r/rust](https://social-rise.com/subreddit-analysis/rust)

fintelia:
Could you make the post?

Shnatsel:
Sure, will do tomorrow.

Shnatsel:
Done: https://redd.it/12ks0ka

PR #395, opened 2023-04-04
Add alpha transformation

### LINKED ISSUE #239: Request: transform to add an alpha channel.
I'd like a transform that adds an alpha channel in the output, if the image doesn't have one.

Essentially I want to have it be able to make the output always be RGBA8, regardless of whatever the file happens to have. Currently you can turn u16 to u8, expand sub-8-bit samples to full bytes, and expand paletted images, but there seems to be no option to force in an alpha channel.

daxpedda:
This adds a new transformation, called `ALPHA`, that adds an alpha channel if missing.

After familiarizing myself with the PNG spec, I realized that color types with alpha aren't supported below 8 bits anyway. So this made the whole thing much simpler, because it can rely on `EXPAND`. So `Transformations::ALPHA` implies `Transformations::EXPAND`.

Which also means it takes tRNS info into account, made sense to me.

Fixes #239.


src/common.rs
```diff
@@ -734,6 +734,8 @@ bitflags! {
         /// less than 8-bit depth to 8-bit depth; and expand tRNS chunks
         /// to alpha channels.
         const EXPAND              = 0x0010; // read only */
+        /// Expand paletted images to include an alpha channel.
+        const ALPHA              = 0x10000; // read only */
```
daxpedda, line 738: I simply increased the bitflag to the next one after `SCALE_16`, which was the last one in the comment above.
Not entirely sure about the name either.

tests/check_testimages.rs
```diff
@@ -143,6 +144,38 @@ fn render_images_identity() {
     });
 }
 
+#[test]
+fn render_images_alpha() {
```
daxpedda, line 148: Happy to add more tests checking for `EXPAND` + `ALPHA`.

src/decoder/mod.rs
```diff
@@ -647,18 +647,26 @@ impl<R: Read> Reader<R> {
...
-                info.color_type
-            };
+                };
+            }
+            if t.contains(Transformations::ALPHA) {
+                color_type = match info.color_type {
+                    Grayscale => GrayscaleAlpha,
+                    Rgb => Rgba,
+                    // TODO
+                    Indexed => Indexed,
```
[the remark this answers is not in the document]
daxpedda: Now that it relies on `EXPAND`, this isn't really relevant anymore.

tests/check_testimages.rs
daxpedda: Now that `ALPHA` relies on `EXPAND`, this isn't really relevant anymore.

daxpedda:
@fintelia friendly ping.

fintelia:
Thanks for the ping. I've been meaning to finish up #396 which refactors the transformations and thus conflicts with this PR. Hopefully in the next few days I'll be able to refresh my memory on both these PRs and figure out the best way to resolve the conflict

fintelia:
Sorry for the delay. That other PR is now merged, so this one can move forward. The transforms are now applied [here](https://github.com/image-rs/image-png/blob/5d927837296572fefc66fe20857e3867a061050f/src/decoder/mod.rs#L602). They read from the pixel data in `row` and write directly into the `output_buffer`.

The easiest approach would probably be to run as a second pass after the other transforms (which works because the input data should take less space in the output buffer than the final output). But you could also try folding it into some/all of the other transforms if it doesn't add too much complexity.

daxpedda:
> But you could also try folding it into some/all of the other transforms if it doesn't add too much complexity.

I think that is what I did, which is what I had before too.

Thanks for the ping! Should be ready to go now.

src/utils.rs
```diff
@@ -38,41 +38,67 @@ where
...
         }
-        output[channels] = if input == trns { 0 } else { 0xFF };
+        output[channels] = if let Some(trns) = trns {
+            if input == trns {
+                0
+            } else {
+                0xFF
+            }
+        } else {
+            0
```
fintelia, line 100: Shouldn't this case be `0xFF`? I'd expect that adding an alpha channel would cause all pixels to be opaque, rather than making all pixels transparent

daxpedda: Strange, how did the tests work so far then?
But yeah, you are right, I had that in my old revision before the rebase as well and was wondering why I did that.

daxpedda: Fixed.

I'm gonna be honest, I didn't dig too deep into this, for some reason my tests work fine with both values. Let me know if you want me to dig into this further.

src/utils.rs
```diff
@@ -38,41 +38,67 @@ where
...
-        output[channels] = if input == trns { 0 } else { 0xFF };
+        output[channels] = if let Some(trns) = trns {
+            if input == trns {
+                0
+            } else {
+                0xFF
+            }
+        } else {
+            0xFF
+        };
```
fintelia, line 101: Would this work as:
```suggestion
output[channels] = if Some(input) == trns { 0 } else { 0xFF };
```

daxpedda: No, that would just create an unused `input` variable.
```suggestion
        output[channels] = match trns {
            Some(trns) if input == trns => 0,
            _ => 0xFF,
        };
```
Should I apply that?

daxpedda: No, actually you are right!
On it.

daxpedda: Done.

fintelia (approved):
I'm convinced that the logic in `next_interlaced_row_impl` could be simplified, but I'm OK with merging this now and leaving that for a followup

PR #400, opened 2023-05-27
Bump `bitflags` to v2

daxpedda:
Changelog: https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#200.

This is a breaking change. The API that is implemented for `Transformations` by the macro is now different.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ include = [
 ]
 
 [dependencies]
-bitflags = "1.0"
+bitflags = "2.0"
 crc32fast = "1.2.0"
 fdeflate = "0.3.0"
 flate2 = "1.0"
--- a/src/common.rs
+++ b/src/common.rs
@@ -725,6 +725,7 @@ bitflags! {
     const SCALE_16            = 0x8000; // read only
     ```
     "]
+    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
     pub struct Transformations: u32 {
         /// No transformation
         const IDENTITY            = 0x0000; // read and write */


fintelia:
Thanks for this PR. Since it is a breaking change, it'll have to wait for the next major release.

fintelia:
Uh, not any immediate plans. There's a number of API changes I'd like to make at the same time, but it'll probably be a while before I have a chance to fully design what those should be

nyurik:
@fintelia thx for your work on this. Is there a reason why you want to bundle all breaking changes together? Version numbers are "cheap" - it's ok to have several breaking versions as they become ready for release -- in my opinion, rapid iteration is better than stalled dependencies... Or am I missing something?  Thx!

fintelia:
No, we do tend to be overly hesitant to do breaking changes.

Part of it is not getting a ton of user feedback beyond things like download counts which show  some users being very slow to update (we still get a significant number of downloads for the `0.16.x` series?!). But another big aspect is [limited maintainer bandwidth](https://github.com/image-rs/organization/issues/15) that makes it easy to focus on issues/changes that can be handled immediately and let PRs with breaking changes drift to the end of the TODO list

anforowicz:
Any updates? :-)

(One other breaking change that I'd like to see merged is https://github.com/image-rs/image-png/pull/421 - maybe it's time to pull the trigger on both of them?  I haven't looked closely at other PRs that may be waiting for their turn because they introduce breaking changes - e.g. I don't have a strong opinion whether https://github.com/image-rs/image-png/pull/488 is ready or not.)

hydra:
Nearly a year has passed now, any chance of a release with this or #516 merged?

nyurik:
thanks!!!

PR #405, opened 2023-08-08
Refactor `utils::unpack_bits` for palette expanded images

okaneco:
Reuse the previous row in `unpack_bits` calculation instead of expanding the palette within the buffer
Use `chunks_exact` and iterate from start to end of the buffer instead of back to front in-place

---

I found this code easier to reason about and might enable someone to optimize this further in the future.
I didn't see a performance difference between the two versions.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -619,12 +619,10 @@ impl<R: Read> Reader<R> {
         };
         match (color_type, trns) {
             (ColorType::Indexed, _) if expand => {
-                output_buffer[..row.len()].copy_from_slice(row);
-                expand_paletted(output_buffer, info, trns)?;
+                expand_paletted(row, output_buffer, info, trns)?;
             }
             (ColorType::Grayscale | ColorType::GrayscaleAlpha, _) if bit_depth < 8 && expand => {
-                output_buffer[..row.len()].copy_from_slice(row);
-                expand_gray_u8(output_buffer, info, trns)
+                expand_gray_u8(row, output_buffer, info, trns)
             }
             (ColorType::Grayscale | ColorType::Rgb, Some(trns)) if expand => {
                 let channels = color_type.samples();
@@ -811,6 +809,7 @@ impl SubframeInfo {
 }
 
 fn expand_paletted(
+    row: &[u8],
     buffer: &mut [u8],
     info: &Info,
     trns: Option<Option<&[u8]>>,
@@ -842,7 +841,7 @@ fn expand_paletted(
                     &[]
                 };
 
-                utils::unpack_bits(buffer, 4, info.bit_depth as u8, |i, chunk| {
+                utils::unpack_bits(row, buffer, 4, info.bit_depth as u8, |i, chunk| {
                     let (rgb, a) = (
                         palette
                             .get(3 * i as usize..3 * i as usize + 3)
@@ -855,7 +854,7 @@ fn expand_paletted(
                     chunk[3] = a;
                 });
             } else {
-                utils::unpack_bits(buffer, 3, info.bit_depth as u8, |i, chunk| {
+                utils::unpack_bits(row, buffer, 3, info.bit_depth as u8, |i, chunk| {
                     let rgb = palette
                         .get(3 * i as usize..3 * i as usize + 3)
                         .unwrap_or(&black);
@@ -873,7 +872,7 @@ fn expand_paletted(
     }
 }
 
-fn expand_gray_u8(buffer: &mut [u8], info: &Info, trns: Option<Option<&[u8]>>) {
+fn expand_gray_u8(row: &[u8], buffer: &mut [u8], info: &Info, trns: Option<Option<&[u8]>>) {
     let rescale = true;
     let scaling_factor = if rescale {
         (255) / ((1u16 << info.bit_depth as u8) - 1) as u8
@@ -881,7 +880,7 @@ fn expand_gray_u8(buffer: &mut [u8], info: &Info, trns: Option<Option<&[u8]>>) {
         1
     };
     if let Some(trns) = trns {
-        utils::unpack_bits(buffer, 2, info.bit_depth as u8, |pixel, chunk| {
+        utils::unpack_bits(row, buffer, 2, info.bit_depth as u8, |pixel, chunk| {
             chunk[1] = if let Some(trns) = trns {
                 if pixel == trns[0] {
                     0
@@ -894,7 +893,7 @@ fn expand_gray_u8(buffer: &mut [u8], info: &Info, trns: Option<Option<&[u8]>>) {
             chunk[0] = pixel * scaling_factor
         })
     } else {
-        utils::unpack_bits(buffer, 1, info.bit_depth as u8, |val, chunk| {
+        utils::unpack_bits(row, buffer, 1, info.bit_depth as u8, |val, chunk| {
             chunk[0] = val * scaling_factor
         })
     }
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -3,15 +3,16 @@ use std::iter::StepBy;
 use std::ops::Range;
 
 #[inline(always)]
-pub fn unpack_bits<F>(row: &[u8], buf: &mut [u8], channels: usize, bit_depth: u8, func: F)
+pub fn unpack_bits<F>(input: &[u8], output: &mut [u8], channels: usize, bit_depth: u8, func: F)
 where
     F: Fn(u8, &mut [u8]),
 {
-    let mask = ((1u16 << bit_depth) - 1) as u8;
+    // Only [1, 2, 4, 8] are valid bit depths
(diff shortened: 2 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -620,11 +620,11 @@ impl<R: Read> Reader<R> {
         match (color_type, trns) {
             (ColorType::Indexed, _) if expand => {
                 output_buffer[..row.len()].copy_from_slice(row);
-                expand_paletted(output_buffer, info, trns)?;
+                expand_paletted(row, output_buffer, info, trns)?;
             }
             (ColorType::Grayscale | ColorType::GrayscaleAlpha, _) if bit_depth < 8 && expand => {
                 output_buffer[..row.len()].copy_from_slice(row);
-                expand_gray_u8(output_buffer, info, trns)
+                expand_gray_u8(row, output_buffer, info, trns)
```
okaneco, line 627: I was able to delete these `copy_from_slice`s and all tests passed, but I'm not sure if that's always correct. I didn't see a performance difference in benchmarks so I left them alone.

It seemed like it could be a bug if `buf.len() < channels` since `unpack_bits` wouldn't expand any data. `buf` would be left with whatever was in the scratch buffer instead of the previous row.

I'm not sure if that's a possible condition to be in by the time the code gets here or if it would have errored out earlier.

fintelia: The intention when I added those `copy_from_slice` calls is that they'd be removed once `expand_{paletted, gray_8}` were refactored (the other cases below already don't use them). It is worth double checking they're not hiding any bugs / size mismatches, but I don't think we should keep them just in case

okaneco: Pushed a commit to remove them.

okaneco: `next_raw_interlaced_row` makes sure that we have a full row and returns an error if not, so I think that makes it okay to remove these.

okaneco:
We could do something like this to special case `BitDepth::Eight` since it doesn't need to calculate the mask and shift.

```rust
if bit_depth == 8 {
    for (&curr, chunk) in row.iter().zip(&mut buf_chunks) {
        func(curr, chunk);
    }
} else {
    let mask = ((1u16 << bit_depth) - 1) as u8;

    for &curr in row.iter() {
        let mut shift = 8 - bit_depth as i32;

        while shift >= 0 {
            if let Some(chunk) = buf_chunks.next() {
                let pixel = (curr >> shift) & mask;
                func(pixel, chunk);
            } else {
                return;
            }

            shift -= bit_depth as i32;
        }
    }
}
```
--- 

I also tried this for the 8-bit depth cases for 3 and 4 channels and played with the constant size. It seemed slightly faster on the benchmark but not enough to show up green on criterion. The 3 byte palette lookup for each pixel seems to make it hard to find low hanging fruit for improvement.

```rust
const CHUNK_SIZE: usize = 32;

let mut buf_chunks = buf.chunks_exact_mut(channels * CHUNK_SIZE);
let mut row_chunks = row.chunks_exact(CHUNK_SIZE);

for (buf_chunk, row_chunk) in (&mut buf_chunks).zip(&mut row_chunks) {
    row_chunk
        .iter()
        .zip(buf_chunk.chunks_exact_mut(channels))
        .for_each(|(&curr, chunk)| func(curr, chunk));
}

(buf_chunks.into_remainder().chunks_exact_mut(channels))
    .zip(row_chunks.remainder().iter())
    .for_each(|(chunk, &curr)| func(curr, chunk));
```

I put `speed_bench_palette.png` into the `tests/benches/` folder and used `cargo bench -- speed_bench`
https://github.com/etemesi254/zune-image/tree/dev/test-images/png/benchmarks

For testing, `cargo test -- render` will run the tests that are most likely to fail without running the whole suite.

src/utils.rs
```diff
@@ -1,40 +1,38 @@
 //! Utility functions
-use std::iter::{repeat, StepBy};
+use std::iter::StepBy;
 use std::ops::Range;
 
 #[inline(always)]
-pub fn unpack_bits<F>(buf: &mut [u8], channels: usize, bit_depth: u8, func: F)
+pub fn unpack_bits<F>(row: &[u8], buf: &mut [u8], channels: usize, bit_depth: u8, func: F)
```
fintelia, line 6: This signature should probably be `unpack_bits<F>(input: &[u8], output: &mut [u8], ...)` or similar

src/utils.rs
```diff
@@ -1,40 +1,38 @@
...
 where
     F: Fn(u8, &mut [u8]),
 {
```
fintelia, line 9: Could you also add an assertion that the `bit_depth` value is reasonable?

src/utils.rs
```diff
@@ -1,40 +1,38 @@
...
+
+    for &curr in row.iter() {
+        let mut shift = 8 - bit_depth as i32;
+
+        while shift >= 0 {
+            if let Some(chunk) = buf_chunks.next() {
+                let pixel = (curr >> shift) & mask;
+                func(pixel, chunk);
+            } else {
+                return;
```
fintelia, line 31: Slightly nervous about silently returning here. If we expect that `row` and `buf` will always have matching lengths (which I think we do?), then we should `unwrap` / `expect` instead of returning

fintelia (commented):
Added a few comments, but generally looks good to me!

fintelia:
I'd say special casing bit_depth = 8 does make sense, but is fine for a followup if you'd prefer

src/utils.rs
okaneco: That makes sense.

In order to change this, I had to swap from iterating over the row in the outer loop to iterating over the chunks. When the shift cycle resets within the loop, we request the next `row` value and if it doesn't exist then it's an error.

okaneco:
Does it make sense to relocate `src/utils.rs` to the `decoder` module in `src/decoder`? The functions insde of it are only used in `decoder`.

With that change, `unpack_bits` would be able to return errors instead of having to panic or assert. The assertion could isntead return `FormatErrorInner::InvalidBitDepth` but I'm not sure there's a proper error for the `input` buffer being out of data.

fintelia:
Moving the method to `src/decoder` seems reasonable.

However, I'd prefer not to make the method return errors. The crate should already be validating any user-provided input buffers and the image bit depth in other places, so if either of those things are wrong in `unpack_bits` it means there's a bug in our code. By triggering a panic, we get a backtrace any time things go wrong and (if we're lucky) can detect the issue from fuzzing rather than waiting for user bug reports about incorrectly decoded images.

This is admittedly a painful tradeoff: Having the code panic means we learn about bugs faster and often they're a bit easier to fix, but at the same time means they potentially have a higher user impact until they're fixed (which can sometimes take quite a while)

src/utils.rs
fintelia: How about asserting that `input.len() * (8/bit_depth) * channels == output.len()` at the start of the method, and then the iteration order doesn't need to be swapped? Or is the problem that the check would fail because some of the call sites actually pass an output buffer that's larger than required?

okaneco: Yes, the checks fail. The input and output buffer sizes seem to vary enough that I can't easily derive an assertion equality that would work.

okaneco:
Thanks for explaining the philosophy of errors vs panics for this situation. I saw a (possibly) redundant check nearby in `expand_paletted` and thought perhaps errors might've been wanted as well further down. But it also makes sense that you want to panic in such an internal function or discover it with fuzzing coverage because it indicates a bug.
https://github.com/image-rs/image-png/blob/7642f0f5a595b7b0efeb773c7279ba735f59bc0e/src/decoder/mod.rs#L819-L828

src/utils.rs
fintelia: That's weird that the checks fail, give that we intentionally set sizes for the two buffers:
https://github.com/image-rs/image-png/blob/7642f0f5a595b7b0efeb773c7279ba735f59bc0e/src/decoder/mod.rs#L582

https://github.com/image-rs/image-png/blob/7642f0f5a595b7b0efeb773c7279ba735f59bc0e/src/decoder/mod.rs#L600

okaneco: `rowlen` and `output_line_size` can be different values.

`rowlen` comes from `Reader::next_pass` and is equivalent to `self.subframe.rowlen`.

`output_line_size` is the result of `Reader::output_line_size` which calls into `ColorType::raw_row_length_from_width` with the width argument being `self.subframe.width` for non-Adam7 images.
https://github.com/image-rs/image-png/blob/7642f0f5a595b7b0efeb773c7279ba735f59bc0e/src/common.rs#L57-L69

---

Writing out a longer answer of speculation, I think I came across a solution.

We want the input buffer to have at least as many shifts as there are output entries. Then, the input row will always have enough lookups for the output buffer chunks.

```rust
let input_max_entries = input.len() * (8 / bit_depth as usize);
let output_entries = output.len() / channels;
assert!(input_max_entries >= output_entries);
```

PR #410, opened 2023-08-31
Ignore subsequent iCCP chunks to match `libpng` behavior.

anforowicz:
PTAL?

This fixes https://github.com/image-rs/image/issues/1825

Question: once/if this gets merged, should I also add a separate test to the `image` crate?  Having `image`-level test coverage seems desirable if it ever decides to switch to a different PNG decoder (e.g. see https://github.com/image-rs/image/issues/1852).


fintelia:
Don't worry about adding the same test to `image`. If we do switch to a different decoding backend (which is now much less likely given significant perf improvements `image-png` since that issue was created), we'll certainly want to double check all the test cases from this repository. But adding every test case from every backend to the main image crate would needlessly bloat those tests, and it would largely be wasted because the relevant code was already checked.

PR #413, opened 2023-09-21
Scaffolding for direct benchmarking of `crate::filter::unfilter`.

anforowicz:
Can you PTAL?  The new benchmarks from this PR should help with work on performance improvements of `fn unfilter`.

The real motivation for this PR is that it is a prelude to a series of commits at https://github.com/image-rs/image-png/compare/bf2c26b5d9d99973d98a850084ea62af32b403e4...1c78a3fa50912240dd05cbcfd192793c23687cde that results in a significant decoding speed-up (measured on Intel).  Special thanks to @veluca93 for help with various SIMD questions that I had.  Results that I got from running the benchmarks:
- `decode/kodim02.png`: -26.6% time
- `decode/kodim17.png`: -25.7% time
- `decode/kodim07.png`: -27.1% time
- `decode/kodim23.png`: -21.8% time
- `unfilter/filter=Sub/bpp=3`: -71.5% time
- `unfilter/filter=Sub/bpp=6`: -86.1% time
- `unfilter/filter=Avg/bpp=3`: -71.8% time
- `unfilter/filter=Avg/bpp=6`: -70.6% time
- `unfilter/filter=Paeth/bpp=3`: -51.4% time
- `unfilter/filter=Paeth/bpp=6`: -62.9% time
- AFAICT other `decode/...` and `unfilter/...` benchmarks are unchanged

I note that the speed-up above depends on `std::simd` which hasn't yet been stabilized, so I thought that it would be best to have a separate PR for just the benchmarks.  (i.e. the new benchmarks seem desirable on their own, even if we decide not to pursue the `std::simd`-based improvements)

/cc @calebzulawski as FYI, since the improvements are possible thanks to `std::simd` and https://github.com/rust-lang/rust/issues/86656.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -594,15 +594,7 @@ impl Info<'_> {
     /// has the consequence that the number of possible values is rather small. To make this fact
     /// more obvious in the type system and the optimizer we use an explicit enum here.
     pub(crate) fn bpp_in_prediction(&self) -> BytesPerPixel {
-        match self.bytes_per_pixel() {
-            1 => BytesPerPixel::One,
-            2 => BytesPerPixel::Two,
-            3 => BytesPerPixel::Three,
-            4 => BytesPerPixel::Four,
-            6 => BytesPerPixel::Six,   // Only rgb×16bit
-            8 => BytesPerPixel::Eight, // Only rgba×16bit
-            _ => unreachable!("Not a possible byte rounded pixel width"),
-        }
+        BytesPerPixel::for_prediction(self.bytes_per_pixel())
     }
 
     /// Returns the number of bytes needed for one deinterlaced image.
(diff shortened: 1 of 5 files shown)


calebzulawski:
That looks great! Feel free to file any issues or let me know if you see any cases where performance is worse than you'd expect.

anforowicz:
FWIW, I assume that [the `mips_cross` CI failure](https://github.com/image-rs/image-png/actions/runs/6267592231/job/17020970355) is unrelated to this PR.  Please shout if you think this is not the case.

calebzulawski:
I don't think it's related (but we ran into similar).  I believe mips was demoted to tier 3, so might not have up to date toolchains

fintelia:
Yeah, the mips failure is unrelated. In the main image crate we switched to powerpc for big endian testing (https://github.com/image-rs/image/pull/1987), so probably just need to do the same thing here.

I should hopefully be able to take a look at the rest of the PR later tonight

src/common.rs
```diff
@@ -695,6 +687,18 @@ impl Info<'_> {
 }
 
 impl BytesPerPixel {
+    pub(crate) fn for_prediction(bpp: usize) -> Self {
```
fintelia, line 690: Could you call this `from_usize` or similar?

fintelia (approved):

fintelia:
Really excited to see the progress to enable using SIMD from _safe_ Rust code! Avoiding, or ideally forbidding, unsafe code is one of the big differentiators that we can provide over the C libraries that are generally used for image decoding. But that's often meant that we either have to forgo potential performance improvements or tolerate some unsafe in order to use SIMD

src/common.rs
anforowicz: Done.  I assume that you plan to squash the last 2 small commits (`cargo fmt` + this rename) when merging?

Issue #368, opened 2022-12-18
Bad encoding of image after image::resize


DLFW:
Hi!
Not sure if I do something stupid or – in case it's a bug – if it's on the  [image](https://github.com/image-rs/image) crate.

However, when I read an image with `image::open`, then resize the image, and then save it with image-png, the resulting file is visually corrupted. When I save it with the `image` crate, it's fine.

## Example

Original:
![test](https://user-images.githubusercontent.com/22507197/208302706-f3c9ccb4-1d7e-4cbb-bcf4-b24c572570fb.png)

Resized and saved with `image-png`:
![test-output-png-encoder](https://user-images.githubusercontent.com/22507197/208302812-5e93aa43-8d12-4f39-8784-74225c995455.png)

Resized and saved with `image::save` is fine:
![test-output-image](https://user-images.githubusercontent.com/22507197/208302733-1f441f1a-f415-4bc5-92d1-b6917802f8ff.png)

## Remarkable

* The problem seems to only occur with 16 bit per color PNG files (at least I could always reproduce it with them and not with 8bpc)
* When I don't scale the image, it works fine
* Intrestingly, when I scale with the [Nearest filter](https://docs.rs/image/0.24.5/image/imageops/enum.FilterType.html), it works. All other filters and the thumbnail operation don't.
* Using image-png 0.17.7

## Minimal example to reproduce

([As project in my GitLab](https://gitlab.com/DLF/rustpngtest/-/tree/example1))

16-bit RGB PNG must exists in `/tmp/test.png`.

```rust
fn main() {
    // Get input with image crate from rgb16 png file
    let in_img = image::open("/tmp/test.png").expect("/tmp/test.png not found");

    assert!(in_img.color() == image::ColorType::Rgb16);

    let out_img = in_img.resize(220, 220, image::imageops::FilterType::Triangle);

    assert!(out_img.color() == image::ColorType::Rgb16);

    // Write it with image - works
    out_img
        .save("/tmp/test-output-image.png")
        .expect("Cannot create output file");

    // Write it with image-png - fails
    let file = std::fs::File::create("/tmp/test-output-png-encoder.png")
        .expect("Cannot create output file");
    let ref mut writer = std::io::BufWriter::new(file);

    let mut encoder = png::Encoder::new(writer, out_img.width(), out_img.height());
    encoder.set_color(png::ColorType::Rgb);
    encoder.set_depth(png::BitDepth::Sixteen);

[...]

PR #414, opened 2023-09-25
Using `std::simd` to speed-up `unfilter` for `Paeth` for bpp=3 and bpp=6

anforowicz:
PTAL?

I hope that we can land this series of 4 commits please.  I understand that relying on unstable language/standard-library features means additional maintenance burden, but:
- There is a precedent for using the `unstable` feature of the `png` crate in the past: 6c0b8fa2a041ddf32e7acfdb5d5877187863ea29 (i.e. this PR doesn't introduce the `unstable` feature)
- The risk of having to deal with breaking changes (which an unstable feature can introduce at any time) only affects library clients that explicitly opt into using the `unstable` feature.

Do you think it might be desirable to cover the nightly + `unstable`-feature-enabled configuration on CI of the `png` crate?  FWIW Chromium uses the nightly compiler (rolled into Chromium toolchain every 1-2 weeks) and can also serve as a canary for detecting breakages (if/once we start depending on the `png` crate - this is still a work-in-progress and we are still evaluating Rust performance against the C/C++ implementation).

Note that I have tried to extend the `std::simd` approach to other bpp cases, but it failed to produce measurable improvements - see: https://github.com/image-rs/image-png/commit/36b541b6b4cf4eec5e1f8f21e544a324c7442bbd

Can you please provide your feedback on whether I should also try to make additional changes to simplify existing `unfilter` code?  (@marshallpierce has pointed out that two versions of the code at https://godbolt.org/z/5MvssMncb compile to the same auto-vectorized code.)  On one hand, simpler code seems nice (easier to read, less wrapping to fit 100 columns).  OTOH, the simplification can be seen as an unnecessary change (and therefore unnecessary risk as auto-vectorization is a bit magical and difficult to test).

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -10,8 +10,7 @@ use crate::common::BytesPerPixel;
 #[cfg(feature = "unstable")]
 mod simd {
     use std::simd::{
-        i16x4, i16x8, u8x4, u8x8, LaneCount, Simd, SimdInt, SimdOrd, SimdPartialEq, SimdUint,
-        SupportedLaneCount,
+        u8x4, u8x8, LaneCount, Simd, SimdInt, SimdOrd, SimdPartialEq, SimdUint, SupportedLaneCount,
     };
 
     /// This is an equivalent of the `PaethPredictor` function from
@@ -54,6 +53,40 @@ mod simd {
             .select(a, smallest.simd_eq(pb).select(b, c))
     }
 
+    /// Memory of previous pixels (as needed to unfilter `FilterType::Paeth`).
+    /// See also https://www.w3.org/TR/png/#filter-byte-positions
+    #[derive(Default)]
+    struct PaethState<const N: usize>
+    where
+        LaneCount<N>: SupportedLaneCount,
+    {
+        /// Previous pixel in the previous row.
+        c: Simd<i16, N>,
+
+        /// Previous pixel in the current row.
+        a: Simd<i16, N>,
+    }
+
+    /// Mutates `x` as needed to unfilter `FilterType::Paeth`.
+    ///
+    /// `b` is the current pixel in the previous row.  `x` is the current pixel in the current row.
+    /// See also https://www.w3.org/TR/png/#filter-byte-positions
+    fn paeth_step<const N: usize>(state: &mut PaethState<N>, b: Simd<u8, N>, x: &mut Simd<u8, N>)
+    where
+        LaneCount<N>: SupportedLaneCount,
+    {
+        // Storing the inputs.
+        let b = b.cast::<i16>();
+
+        // Calculating the new value of the current pixel.
+        let predictor = paeth_predictor(state.a, b, state.c);
+        *x += predictor.cast::<u8>();
+
+        // Preparing for the next step.
+        state.c = b;
+        state.a = x.cast::<i16>();
+    }
+
     fn load3(src: &[u8]) -> u8x4 {
         u8x4::from_array([src[0], src[1], src[2], 0])
     }
@@ -67,28 +100,12 @@ mod simd {
         debug_assert_eq!(prev_row.len(), curr_row.len());
         debug_assert_eq!(prev_row.len() % 3, 0);
 
-        // Paeth tries to predict pixel x using the pixel to the left of it, a,
-        // and two pixels from the previous row, b and c:
-        //
-        //     prev_row: c b
-        //     curr_row: a x
-        //
-        // The first pixel has no left context, and so uses an Up filter, p = b.
-        // This works naturally with our main loop's p = a+b-c if we force a and c
-        // to zero.
-        let mut a = i16x4::default();
-        let mut c = i16x4::default();
-
+        let mut state = PaethState::<4>::default();
         for (prev, curr) in prev_row.chunks_exact(3).zip(curr_row.chunks_exact_mut(3)) {
-            let b = load3(prev).cast::<i16>();
+            let b = load3(prev);
             let mut x = load3(curr);
-
-            let predictor = paeth_predictor(a, b, c);
-            x += predictor.cast::<u8>();
+            paeth_step(&mut state, b, &mut x);
             store3(x, curr);
-
-            c = b;
-            a = x.cast::<i16>();
         }
     }
 
@@ -105,28 +122,12 @@ mod simd {
         debug_assert_eq!(prev_row.len(), curr_row.len());
         debug_assert_eq!(prev_row.len() % 6, 0);
 
-        // Paeth tries to predict pixel x using the pixel to the left of it, a,
-        // and two pixels from the previous row, b and c:
-        //
-        //     prev_row: c b
-        //     curr_row: a x
-        //
-        // The first pixel has no left context, and so uses an Up filter, p = b.
-        // This works naturally with our main loop's p = a+b-c if we force a and c
-        // to zero.
-        let mut a = i16x8::default();
-        let mut c = i16x8::default();
-
+        let mut state = PaethState::<8>::default();
         for (prev, curr) in prev_row.chunks_exact(6).zip(curr_row.chunks_exact_mut(6)) {
-            let b = load6(prev).cast::<i16>();
+            let b = load6(prev);
             let mut x = load6(curr);
-
-            let predictor = paeth_predictor(a, b, c);
-            x += predictor.cast::<u8>();
+            paeth_step(&mut state, b, &mut x);
             store6(x, curr);
-
-            c = b;
-            a = x.cast::<i16>();
         }
     }
 }
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -2,6 +2,87 @@ use core::convert::TryInto;
 
 use crate::common::BytesPerPixel;
 
+/// SIMD helpers for `fn unfilter`
+///
+/// TODO(https://github.com/rust-lang/rust/issues/86656): Stop gating this module behind the
+/// "unstable" feature of the `png` crate.  This should be possible once the "portable_simd"
+/// feature of Rust gets stabilized.
+#[cfg(feature = "unstable")]
+mod simd {
+    use std::simd::{i16x4, u8x4, SimdInt, SimdOrd, SimdPartialEq, SimdUint};
+
+    /// This is an equivalent of the `PaethPredictor` function from
+    /// [the spec](http://www.libpng.org/pub/png/spec/1.2/PNG-Filters.html#Filter-type-4-Paeth)
+    /// except that it simultaenously calculates the predictor for all SIMD lanes.
+    /// Mapping between parameter names and pixel positions can be found in
+    /// [a diagram here](https://www.w3.org/TR/png/#filter-byte-positions).
+    ///
+    /// Examples of how different pixel types may be represented as multiple SIMD lanes:
+    /// - RGBA => 4 lanes of `i16x4` contain R, G, B, A
+    /// - RGB  => 4 lanes of `i16x4` contain R, G, B, and a ignored 4th value
+    ///
+    /// The SIMD algorithm below is based on [`libpng`](https://github.com/glennrp/libpng/blob/f8e5fa92b0e37ab597616f554bee254157998227/intel/filter_sse2_intrinsics.c#L261-L280).
+    fn paeth_predictor(a: i16x4, b: i16x4, c: i16x4) -> i16x4 {
+        let pa = b - c; // (p-a) == (a+b-c - a) == (b-c)
+        let pb = a - c; // (p-b) == (a+b-c - b) == (a-c)
+        let pc = pa + pb; // (p-c) == (a+b-c - c) == (a+b-c-c) == (b-c)+(a-c)
+
+        let pa = pa.abs();
+        let pb = pb.abs();
+        let pc = pc.abs();
+
+        let smallest = pc.simd_min(pa.simd_min(pb));
+
+        // Paeth algorithm breaks ties favoring a over b over c, so we execute the following
+        // lane-wise selection:
+        //
+        //     if smalest == pa
+        //         then select a
+        //         else select (if smallest == pb then select b else select c)
+        smallest
+            .simd_eq(pa)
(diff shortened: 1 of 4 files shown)


okaneco:
Regarding the last paragraph on simplification, it appears that the code on the left didn't autovectorize until Rust 1.72. The code on the right is autovectorized from Rust 1.65 onward.
https://godbolt.org/z/MKe4qE4sc (both at rustc 1.71)

fintelia:
I just did some benchmarking of these changes on my CPU (a Ryzen 5600X). Overall, I do see substantial improvement in the Paeth unfiltering time (**-21%** for 3bpp and **-38%** for 6bpp) which is particularly impactful because it tends to be a bottleneck. 

However, another thing I've discovered is that there are severe performance bug in `rustc 1.72.0` (https://github.com/rust-lang/rust/issues/115212) which was resolved in https://github.com/rust-lang/rust/pull/115236 and backported to `rustc 1.72.1`. Using a non-buggy compiler, the performance changes for Avg and Sub were quite mixed. Sub/bpp=6 did get 10% faster, but it already ran at nearly 12 GB/s so that's getting into diminishing returns.

anforowicz:
Thanks for taking a look and pointing out that the impact of the changes is different on the latest nightly.  I've re-tested with `rustc 1.74.0-nightly (5ae769f06 2023-09-26)` and I see the following changes in the new `unfilter` benchmark:
- Significant Paeth improvements (time -21% for bpp=3 and -37% for bpp=6)
- Small Avg improvement for bpp=3 (time -5%) and a big regression for bpp=6 (time +30%)
- Small Sub improvement for bpp=6 (time -9%) and a big regression for bpp=3 (time +80%)

I've tried rerunning the benchmarks with `--sample-size=1000 --measurement-time=50` (i.e. collecting 10 times more data) and the Avg/bpp=3 gains disappeared while Sub/bpp=6 gains persisted.

So, I am not sure how to proceed here:
- Option 1: Redo/simplify this PR to just focus on Paeth (I think I'll just rebase + force push instead of fixing forward)
- Option 2: Use old code for Avg and Sub/bpp=3, but still include the Sub/bpp=6 changes

WDYT?  I think I'd lean toward Option 1, because bpp=6 seems less common than bpp=3 (I don't have hard data to back that up though) and because the gains for Sub/bpp=6 were relatively small (with default Criterion settings I measured -9%;  when collecting x10 samples three times I got: -5.2%, -5.6%, -4.6%).

fintelia:
I'd say go with option 1. It isn't just that sub filtering with bpp=6 is pretty rare. Another big factor is that because decoding time is split between decompression (which happens at roughly 1 GB/s) and unfiltering, Amdahl's law means that speeding up an already very fast unfiltering method will have only a very small impact on the total decoding time.

anforowicz:
Okay - I switched to option 1 (`std::simd`-ification of only the `Paeth` `unfilter` for bpp=3 and bpp=6).  PTAL?

FWIW some failures were reported at https://github.com/image-rs/image-png/actions/runs/6330253240:
- `mips_cross` seems like an unrelated, known issue (based on https://github.com/image-rs/image-png/pull/413#issuecomment-1730479173)
- `test_all` failure *is* introduced by this PR.  It seems that the CI needs to be told to only test the `unstable` crate feature when using the unstable/nightly `rustc`.

src/filter.rs
```diff
@@ -2,6 +2,170 @@ use core::convert::TryInto;
...
+
+            // We can speculate that writing 4 bytes might be more efficient (just as with using
+            // `u8x4::from_slice` above), but we can't use that here, because we can't clobber the
+            // first byte of the next pixel in the `curr_row`.
+            store3(x, curr_row);
+
+            prev_row = &prev_row[3..];
+            curr_row = &mut curr_row[3..];
+        }
+        // Can't use `u8x4::from_slice` for the last `[u8;3]`.
```
okaneco, line 122:
```suggestion
        }
        assert!(prev_row.len() >= 3 && curr_row.len() >= 3);
        // Can't use `u8x4::from_slice` for the last `[u8;3]`.
```

(Not a reviewer, just interested in this work 👍)

It looks like adding an assert after the loop removes some bounds checks, L127-130 on the left side in diff view. I wouldn't expect a change in performance though.
https://rust.godbolt.org/z/WdGP8fqxG

---
Similarly, the compiler should know that `prev_row` and `curr_row` are the same length so it probably doesn't need to be added to the conditional check in the `while` loop above. It removes all the other panic handling code in Compiler Explorer but that's missing the context of the full library code so I wouldn't expect it to affect performance, it might just add a superfluous check.
https://rust.godbolt.org/z/avzr68esK
```rust
    while prev_row.len() >= 4 && curr_row.len() >= 4 {
```

anforowicz: Thanks!  I have trouble reading and following the disassembly, so I really appreciate your feedback here.

(BTW: Have you seen the comparison of C++ vs Rust disasembly posted in https://github.com/image-rs/image-png/discussions/416: https://godbolt.org/z/j8fK39oEE ?)

okaneco:
> Thanks! I have trouble reading and following the disassembly, so I really appreciate your feedback here.
> 
> (BTW: Have you seen the comparison of C++ vs Rust disasembly posted in #416: https://godbolt.org/z/j8fK39oEE ?)

I haven't looked too closely at the algorithm differences, the intrinsics look like they should produce roughly the same code. More on that under the divider.

One noticeable difference in the Rust version is the generation of constants at the top. They looked like masks, so I tried to see what part of the code was making that. If the load function is changed, then a majority of them disappear which may help with performance (220 -> 162 lines, 4 less movs). There's one more label of data that we might be able to remove. I haven't tested or ran this code but it looks equivalent.
https://rust.godbolt.org/z/YzWMbdsco

```rust
fn load3(src: &[u8]) -> u8x4 {
    let mut temp = [0; 4];
    temp[..3].copy_from_slice(&src[..3]);
    u8x4::from_slice(&temp)
}
```

---

The manual `abs` produces the same assembly output in Rust. https://rust.godbolt.org/z/36jTz7e5d

```rust
fn i16x4_abs(mut a: i16x4) -> i16x4 {
    let zeros = i16x4::default();
    let is_negative = a.simd_lt(zeros).to_int().cast::<u16>();
    let xored = a.cast::<u16>() ^ is_negative;
    (xored - is_negative).cast::<i16>()
}
```

I wrote out the manual select, but I haven't tested or benchmarked this code. It uses one more packed compare than the `select`. https://rust.godbolt.org/z/cbYK5xchf

```rust
fn if_then_else(c: i16x4, t: i16x4, e: i16x4) -> i16x4 {
    (c & t) | (c & !e)
}
//     if_then_else(
//         smallest.simd_eq(pa).to_int(),
//         a,
//         if_then_else(smallest.simd_eq(pb).to_int(), b, c),
//     )
```

When looking at comparisons in godbolt, I make use of the maximize and minimize on the diff view and ctrl+f to get an idea of the differences in instructions. I also use the code preview in the scroll bar to get an idea of instruction count.

anforowicz:
Thanks @okaneco!  I've tried implementing your proposed changes in https://github.com/image-rs/image-png/commit/97d4f63b85c77e3b8f29b44ac6f03bd028ea5b1d and when measuring them (AMD EPYC 7B12, rustc 1.74.0-nightly (5ae769f06 2023-09-26)) I didn't see an improvement:

```
$ taskset --cpu-list 4-7 nice -n -19 target/release/deps/unfilter-a5b9c1f23e358474 --bench --baseline=simd1 Paeth/bpp=[36]
...
unfilter/filter=Paeth/bpp=3
                        time:   [18.635 µs 18.658 µs 18.677 µs]
                        thrpt:  [627.44 MiB/s 628.09 MiB/s 628.84 MiB/s]
                 change:
                        time:   [-0.8646% -0.5573% -0.2935%] (p = 0.00 < 0.05)
                        thrpt:  [+0.2944% +0.5604% +0.8721%]
Found 30 outliers among 100 measurements (30.00%)
  15 (15.00%) low severe
  8 (8.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe

unfilter/filter=Paeth/bpp=6
                        time:   [18.867 µs 18.875 µs 18.883 µs]
                        thrpt:  [1.2121 GiB/s 1.2126 GiB/s 1.2131 GiB/s]
                 change:
                        time:   [-0.3220% -0.1147% +0.0969%] (p = 0.30 > 0.05)
                        thrpt:  [-0.0968% +0.1148% +0.3230%]
Found 18 outliers among 100 measurements (18.00%)
  3 (3.00%) low severe
  1 (1.00%) low mild
  8 (8.00%) high mild
  6 (6.00%) high severe


lukasza2 in /usr/local/google/home/lukasza/src/github/image-png
$ taskset --cpu-list 4-7 nice -n -19 target/release/deps/unfilter-a5b9c1f23e358474 --bench --baseline=simd1 Paeth/bpp=[36]
...
unfilter/filter=Paeth/bpp=3
                        time:   [18.696 µs 18.699 µs 18.702 µs]
                        thrpt:  [626.59 MiB/s 626.71 MiB/s 626.81 MiB/s]
                 change:
                        time:   [+0.0355% +0.0812% +0.1249%] (p = 0.00 < 0.05)
                        thrpt:  [-0.1248% -0.0812% -0.0354%]
  4 (4.00%) high mild
  4 (4.00%) high severe

unfilter/filter=Paeth/bpp=6
                        time:   [18.874 µs 18.879 µs 18.885 µs]
                        thrpt:  [1.2120 GiB/s 1.2124 GiB/s 1.2127 GiB/s]
                 change:
                        time:   [-0.1953% -0.0306% +0.1590%] (p = 0.74 > 0.05)
                        thrpt:  [-0.1587% +0.0306% +0.1957%]
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
```

So, I think it's okay to stick with the current, simpler code in the PR.

okaneco:
Thanks for trying that out. I prefer the simpler code as well, and we're probably at the limit of improvement for this area of code without greater architectural changes.

I noticed when adding `if_then_else` to the full unfilter program that it produced 2 more instructions. So even though the extra data was removed, the change to `load3` is probably perf neutral-to-slightly-negative.

https://rust.godbolt.org/z/6hfzGYxqh (updated `load3` on the left, `if_then_else` on right)

---

The other difference I could easily spot was that our casting contains unnecessary instructions. I wrongfully assumed casting would saturate when narrowing, but it's an `as` cast:

> This follows the semantics of Rust’s `as` conversion for casting integers (wrapping to other integer types, and saturating to float types).
_https://doc.rust-lang.org/stable/std/simd/trait.SimdInt.html#tymethod.cast_

Since it follows `as`-cast semantics, the conversion looks like `(i16 & 255) as u8`. This results in the code having that label of 8 `.short 255` and calling `pand` to clip to `u8` before calling `packus` and `padd`. This probably isn't a huge bottleneck but it's unnecessary in our case (and seemingly unavoidable while using `Simd` casts) since we know the values are originally bytes.

In the C++ code, the addition can be done in place by reinterpreting the `__m128i`, previously treated as `i16`, as a vector of `u8` with `_mm_add_epi8`. Then to store, they call the saturating pack instrinsic which narrows from `i16` to `u8`. It'd be nice if we could safely reinterpret the `Simd` structs as if we were using intrinsics (such as going from `i16x4` to `u8x8` to perform wrapping addition), but I'm not aware of a way.

```c++
      // L100-102
      /* Note `_epi8`: we need addition to wrap modulo 255. */
      d = _mm_add_epi8(d, nearest);
      store3(row, _mm_packus_epi16(d,d));
```
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=packus&techs=SSE_ALL&ig_expand=4903

anforowicz:
> In the C++ code, the addition can be done in place by reinterpreting the `__m128i`, previously treated as `i16`, as a vector of `u8` with `_mm_add_epi8`. Then to store, they call the saturating pack instrinsic which narrows from `i16` to `u8`. It'd be nice if we could safely reinterpret the `Simd` structs as if we were using intrinsics (such as going from `i16x4` to `u8x8` to perform wrapping addition), but I'm not aware of a way.

I guess that in the long-term this may be solved by https://github.com/rust-lang/project-safe-transmute and in the short-term maybe `bytemuck` (or `std::simd`) can declare that `Simd` is `AnyBitPattern + NoUninit`?

okaneco:
Yes, I believe this issue is relevant.
https://github.com/rust-lang/portable-simd/issues/116

A `to_bits`/`from_bits` would be nice as mentioned in the discussion. The PR `#368` in that repo might be making `to_bytes` features public but I'm not sure.

src/filter.rs
```diff
@@ -2,6 +2,171 @@ use core::convert::TryInto;
...
+
+    fn load6(src: &[u8]) -> u8x8 {
+        u8x8::from_array([src[0], src[1], src[2], src[3], src[4], src[5], 0, 0])
+    }
+
+    fn store6(src: u8x8, dest: &mut [u8]) {
+        dest[0..6].copy_from_slice(&src.to_array()[0..6])
+    }
+
+    /// Undoes `FilterType::Paeth` for `BytesPerPixel::Three`.
```
fintelia, line 138: nit: This should say `BytesPerPixel::Six`

fintelia:
Other than the CI issue of enabling the "unstable" feature with the stable compiler, is this ready to merge?

src/filter.rs
anforowicz: Done

anforowicz:
> Other than the CI issue of enabling the "unstable" feature with the stable compiler, is this ready to merge?

Yes, I think so.

anforowicz:
Quick disclaimer: this PR only helps for RGB8 and RGB16 images which (according to [the notes here](https://github.com/image-rs/image-png/discussions/416#discussioncomment-7436871)) account for only 3.9% of PNG images found in top 500 websites.  Nevertheless, it's probably still desirable to merge these changes IMO.

anforowicz:
BTW: sorry for being slow with this PR - I didn't realize that I have the power to change the CI configuration through the PR itself .  I... err... don't have much experience with github workflows... :-/

PR #420, opened 2023-10-31
New PNG inputs for benchmarking non-compressed, non-filtered data.

anforowicz:
PTAL?

The new test images are somewhat artificial, but they help to see the performance of the bulk of the PNG decoding code.  Runtime overhead of this code is relatively small when compared to the overhead of decompression (mostly in the `fdeflate` crate) and unfiltering (in `src/filter.rs`), but performance of this code is still important (e.g. as discussed in [another discussion thread](https://github.com/image-rs/image-png/discussions/416#discussioncomment-7436871), optimizing this code can yield 10+% of gains for a more realistic test corpus).  And I think that these new test images will be a useful tool for evaluating performance of this code (which may be more difficult to confidently measure using other test inputs).

I am adding 2 separate images, because I think that:

* The 128x128 image will be useful for evaluating performance of copying the image data out of a PNG file
* The 8x8 image will be useful for evaluating constant costs (e.g. when trying to optimize handling of small images - favicons, emojis, etc.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ flate2 = "1.0"
 miniz_oxide = { version = "0.7.1", features = ["simd"] }
 
 [dev-dependencies]
+byteorder = "1.5.0"
 clap = { version = "3.0", features = ["derive"] }
 criterion = "0.3.1"
 getopts = "0.2.14"
--- a/benches/decoder.rs
+++ b/benches/decoder.rs
@@ -3,6 +3,8 @@ use std::fs;
 use criterion::{criterion_group, criterion_main, Criterion, Throughput};
 use png::Decoder;
 
+mod png_generator;
+
 fn load_all(c: &mut Criterion) {
     for entry in fs::read_dir("tests/benches/").unwrap().flatten() {
         match entry.path().extension() {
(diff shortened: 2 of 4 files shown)


tests/benches/noncompressed/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+fn write_noncompressed_deflate_block(w: &mut impl Write, data: &[u8], last: bool) {
+    if last {
+        w.write_u8(1).unwrap();
+    } else {
+        w.write_u8(0).unwrap();
+    }
+
+    assert!(data.len() < 65535);
```
fintelia, line 46: nit: Should be `<=`

tests/benches/noncompressed/src/main.rs
```diff
@@ -0,0 +1,108 @@
...
+            .take(width as usize)
+            .flatten()
+            .collect::<Vec<_>>()
+    };
+
+    let mut zlib_data = Vec::new();
+    zlib_data.write_u8(0x78).unwrap(); // CM + CINFO
+    zlib_data.write_u8(0x01).unwrap(); // FLG
+
+    let mut chunks = image_pixels.as_slice().chunks_exact(8192);
```
fintelia, line 79: Is there a particular reason to prefer 8KB blocks rather than the max size? If not, fdeflate has a [`StoredOnlyCompressor`](https://docs.rs/fdeflate/latest/fdeflate/struct.StoredOnlyCompressor.html) that could replace this manual deflate stream construction

fintelia (commented):
I wonder if it would be better to add a new benchmark that constructed these uncompressed files on the fly. That would enable us to have a larger number of variations without having to check in all the (uncompressed!) files into git.

In particular, it might be interesting to try breaking up the files into different sized iDAT chunks (8KB vs 64KB vs unlimited), and to try some much larger files (some decoders have a performance cliff when the image no longer fits in the L2 cache).

tests/benches/noncompressed/src/main.rs
anforowicz: Thanks!  TIL :-).  I should have checked `fdeflate` before writing this code (I've struggled for quite a bit to produce a working zlib stream).  I'll make the switch in a few minutes.

anforowicz:
> I wonder if it would be better to add a new benchmark that constructed these uncompressed files on the fly. That would enable us to have a larger number of variations without having to check in all the (uncompressed!) files into git.

Sounds good.  I don't have a strong reason to work with actual files (rather than generating inputs at test runtime), but let me list some weak reasons below:

* Consistency with the existing benchmarks
* Being able to use the PNGs in other benchmarks (e.g. in benchmarks that also cover C++/Rust FFI, or in benchmarks of other decoders like `libpng` or `wuffs`)

tests/benches/noncompressed/src/main.rs
anforowicz: Thanks for flagging.  Done.

tests/benches/noncompressed/src/main.rs
anforowicz: Done

anforowicz:
@fintelia - what are the next steps for this PR?  Are there any actions I should take from my side?  (once this gets merged, I plan to rebase and submit the next follow-up I have lined up here: https://github.com/image-rs/image-png/compare/master...anforowicz:image-png:idat-straight-to-zlibstream)

PR #422, opened 2023-11-03
Reduce copying and allocations

fintelia:
This eliminate Reader::prev and adds special handling of unfiltering for the first row.

It has minimal impact overall on the QOI corpus. However, if limited to QOI corpus images under 64KB, the average decoding speed goes from ~~244 MP/s to 436 MP/s~~ on my machine. _Edit: I think I messed up the benchmarks but I still see a modest perf improvement_ 

cc: @anforowicz

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -210,9 +210,9 @@ impl<R: Read> Decoder<R> {
             subframe: SubframeInfo::not_yet_init(),
             fctl_read: 0,
             next_frame: SubframeIdx::Initial,
-            prev: Vec::new(),
-            current: Vec::new(),
-            scan_start: 0,
+            data_stream: Vec::new(),
+            prev_start: 0,
+            current_start: 0,
             transform: self.transform,
             scratch_buffer: Vec::new(),
             limits: self.limits,
@@ -347,12 +347,12 @@ pub struct Reader<R: Read> {
     /// control chunk. The IDAT image _may_ have such a chunk applying to it.
     fctl_read: u32,
     next_frame: SubframeIdx,
-    /// Previous raw line
-    prev: Vec<u8>,
-    /// Current raw line
-    current: Vec<u8>,
-    /// Start index of the current scan line.
-    scan_start: usize,
+    /// Vec containing the uncompressed image data currently being processed.
+    data_stream: Vec<u8>,
+    /// Index in `data_stream` where the previous row starts.
+    prev_start: usize,
+    /// Index in `data_stream` where the current row starts.
+    current_start: usize,
     /// Output transformations
     transform: Transformations,
     /// This buffer is only used so that `next_row` and `next_interlaced_row` can return reference
@@ -444,8 +444,7 @@ impl<R: Read> Reader<R> {
             return Err(DecodingError::LimitsExceeded);
         }
 
-        self.prev.clear();
-        self.prev.resize(self.subframe.rowlen, 0);
+        self.prev_start = self.current_start;
 
         Ok(())
     }
@@ -504,8 +503,9 @@ impl<R: Read> Reader<R> {
             line_size: self.output_line_size(self.subframe.width),
         };
 
-        self.current.clear();
-        self.scan_start = 0;
+        self.data_stream.clear();
+        self.current_start = 0;
+        self.prev_start = 0;
         let width = self.info().width;
         if self.info().interlaced {
             while let Some(InterlacedRow {
@@ -597,7 +597,8 @@ impl<R: Read> Reader<R> {
         output_buffer: &mut [u8],
     ) -> Result<(), DecodingError> {
         self.next_raw_interlaced_row(rowlen)?;
-        let row = &self.prev[1..rowlen];
+        assert_eq!(self.current_start - self.prev_start, rowlen - 1);
+        let row = &self.data_stream[self.prev_start..self.current_start];
 
         // Apply transformations and write resulting data to buffer.
         let (color_type, bit_depth, trns) = {
@@ -706,8 +707,7 @@ impl<R: Read> Reader<R> {
                 let (pass, line, width) = adam7.next()?;
                 let rowlen = self.info().raw_row_length_from_width(width);
                 if last_pass != pass {
-                    self.prev.clear();
-                    self.prev.resize(rowlen, 0u8);
+                    self.prev_start = self.current_start;
                 }
                 Some((rowlen, InterlaceInfo::Adam7 { pass, line, width }))
             }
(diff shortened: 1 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -756,17 +757,21 @@ impl<R: Read> Reader<R> {
         }
 
         // Get a reference to the current row and point scan_start to the next one.
```
anforowicz, line 759: nit: Probably need to remove the remaining reference to `scan_start` here

src/decoder/mod.rs
```diff
@@ -723,27 +723,28 @@ impl<R: Read> Reader<R> {
...
                     FormatErrorInner::NoMoreImageData.into(),
                 ));
             }
 
             // Clear the current buffer before appending more data.
-            if self.scan_start > 0 {
-                self.current.drain(..self.scan_start).for_each(drop);
-                self.scan_start = 0;
+            if self.prev_start > 0 {
+                self.data_stream.drain(..self.prev_start).for_each(drop);
```
anforowicz, line 735: The `drain` / `for_each(drop)` approach seems complex - I wonder if the compiler is able to optimize this as well as a more direct move - e.g. something like:

```
            self.data_stream.copy_within(self.prev_start.., 0);
            self.data_stream.truncate(self.data_stream.len() - self.current_start);
```

I can't really read assembly well, but the `drain` approach seems to at least result in longer assembly: https://godbolt.org/z/z48q1xej5

OTOH, maybe this feedback doesn't belong to this PR and we should fix this in a separate PR?  (e.g. `drain`-based compaction is also used in `zlib.rs`)

src/decoder/mod.rs
```diff
@@ -444,8 +444,7 @@ impl<R: Read> Reader<R> {
             return Err(DecodingError::LimitsExceeded);
         }
 
-        self.prev.clear();
-        self.prev.resize(self.subframe.rowlen, 0);
+        self.prev_start = self.current_start;
```
anforowicz, line 447: optional, feel free to ignore: There is minimal code duplication here (e.g. new line 710 looks the same) + I wonder if the code would be more readable if this single line was hoisted into a well-named separate function (maybe `reset_prev_row`)?

src/decoder/mod.rs
```diff
@@ -347,12 +347,12 @@ pub struct Reader<R: Read> {
...
-    prev: Vec<u8>,
-    /// Current raw line
-    current: Vec<u8>,
-    /// Start index of the current scan line.
-    scan_start: usize,
+    /// Vec containing the uncompressed image data currently being processed.
+    data_stream: Vec<u8>,
+    /// Index in `data_stream` where the previous row starts.
+    prev_start: usize,
+    /// Index in `data_stream` where the current row starts.
```
anforowicz, line 354: I wonder if it would be useful to point out that

* 1) `prev_start` points at pixel data - `rowlen - 1` worth of bytes, but 2) `current_start` points filter kind byte + pixel data (potentially incomplete - at most `rowlen` worth of bytes).
* `prev_start == current_start` indicates that there is no previous row

src/filter.rs
```diff
@@ -452,6 +461,82 @@ pub(crate) fn unfilter(
                 *curr = curr.wrapping_add(above);
             }
         }
+        Avg if previous.is_empty() => match tbpp {
```
anforowicz, line 464: I wonder if `roundtrip_ascending_previous_line` and `roundtrip` tests (below in the same source file) could be duplicated and/or tweaked to explicitly cover round-tripping when there is no previous row?

anforowicz (commented):
Nice - thanks!

Some notes:

* There is a choice here between 1) special-casing handling of empty `previous` in `fn unfilter` (what this PR does) and 2) having some zeros at the beginning of `data_stream`.  The latter would result in smaller code size + wouldn't require a branch at the beginning of `unfilter`.  But the difference between 1 and 2 is probably small, and the current PR is an improvement, so probably no point in overthinking this aspect.

* There is also a choice between 1) having a single `data_stream` buffer that gets compacted from time to time (what this PR does) vs 2) having two buffers and avoiding compaction.  The latter can be implemented by 2a) always having `rowlen` of bytes in `prev` and `curr` and swapping them (implemented [here](https://github.com/image-rs/image-png/commit/7aaf20773f2e8c3e0ee590151314961b7d819d61)) or 2b) having 2 longer buffers (implemented [here](https://github.com/image-rs/image-png/compare/f5021fbb0c5c05c5184a7bd6f453ed1bfcf1bc82...29fba626d8785d8a7c3bf14416218855da00552f)).  I think it may be worth comparing all 3 approaches (e.g. testing/benching them using the `noncompressed` test inputs) to see how big the cost of compaction is.

anforowicz:
> it may be worth comparing all 3 approaches

Not sure if the data below helps or just stirs the pot unnecessarily :-) : 

* `prev-row-copy-avoidance--double-long-buffer` (see [here](https://github.com/image-rs/image-png/compare/2d7efdda3bdf577d1b377ae0aac46d083cdf222d%5E...c87c88775505efbdd3b07d00cb226e017281f30c))

    - `noncompressed-8x8.png`: change in time:
      `[-3.4699% -3.2960% -3.1177%] (p = 0.00 < 0.05)`
    - `noncompressed-128x128.png`: change in time:
      `[-1.7766% -1.3460% -0.9357%] (p = 0.00 < 0.05)`

* `prev-row-copy-avoidance--fintelia` (see [here](https://github.com/image-rs/image-png/compare/2d7efdda3bdf577d1b377ae0aac46d083cdf222d%5E...ffbf4d7887de7832c570abf1eaf3d54a63eb2e40))

    - `noncompressed-8x8.png`: change in time:
      `[-5.5093% -5.3152% -5.0900%] (p = 0.00 < 0.05)`
    - `noncompressed-128x128.png`: change in time:
      `[-5.2180% -4.8080% -4.3969%] (p = 0.00 < 0.05)`

* `prev-row-copy-avoidance--swap-prev-and-curr` (see [here](https://github.com/image-rs/image-png/compare/2d7efdda3bdf577d1b377ae0aac46d083cdf222d%5E...afa4a1d44311f1691f5763400b8607ef6decce19))

    - `noncompressed-8x8.png`: change in time:
      `[+16.659% +16.859% +17.074%] (p = 0.00 < 0.05)`
    - `noncompressed-128x128.png`: change in time:
      `[-17.995% -17.647% -17.318%] (p = 0.00 < 0.05)`

I am very surprised by the wild changes from `prev-row-copy-avoidance--swap-prev-and-curr`.

I still worry about the cost of compaction (i.e. the calls to `drain` in `Reader::next_raw_interlaced_row` and `ZlibStream::transfer_finished_data`), but maybe we should just merge this PR and move on.

anforowicz:
@fintelia, I just wanted to double-check that you've seen *all* of my comments above.  (The performance comparison is probably safe to ignore, but I was wondering about my comments/suggestions about unit test coverage and about tweaking code comments that refer to the deleted/renamed `scan_start` field.)

fintelia:
Somehow didn't see those comments or your changes on #420. Not exactly sure what happened, but I'll address them when I have a chance

PR #425, opened 2023-11-14
Upgrade fdeflate and remove ZlibStream::in_buffer

fintelia:
fdeflate 0.3.1 guarantees forward progress, so there's no need to buffer input data to the zlib stream.


anforowicz:
This is great!  Thanks!

PR #428, opened 2023-11-15
Faster u32 parsing

anforowicz:
This series of 5 commits refactors how the code parses `u32` values.  The cumulative impact on the runtime of the `decode/generated-png:noncompressed-8x8.png` benchmark looks as follows:

* [-7.1475% -6.9281% -6.6885%] (p = 0.00 < 0.05)
* [-7.7189% -7.5262% -7.3241%] (p = 0.00 < 0.05)
* [-7.4631% -7.1972% -6.9319%] (p = 0.00 < 0.05)

I expect that this PR will have minimal impact on bigger images, so I didn't even try running the other benchmarks (I assume that any changes observed in the other benchmarks would be just statistical noise).

I think that this PR turned out quite nicely from code simplification and readability perspective.  Even if there was no performance impact, I think it might be desirable to land it from this perspective.

Various notes:

* I wondered whether it might be desirable to add some additional test coverage for partially read `u32` values.
    - FWIW this path seems to be covered by `decoder::tests::no_data_dup_on_finish` - only one test... :-(
    - Once `BufRead` changes land, maybe we should tweak the `render_images` test to work with `BufReader::with_capacity(1, <original input>)` which should test dripping the input byte-by-byte.
    - I *did* try to run all the tests after temporarily tweaking `BufReader::with_capacity` calls in `src/decoder/mod.rs` so that a buffer of only 1 byte was used.
* I wondered if `U32ValueKind::Type` after an `IDAT` chunk really needs to go through 2 state transitions (with a detour via `Decoded::ImageDataFlushed` during which `self.state` is set to the old/unchanged value).  This is probably in the unnecessary-microoptimizations territory, but I was hoping to simplify the code a bit more.  Still... I don't see how to do this cleanly.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -30,6 +30,12 @@ const CHECKSUM_DISABLED: bool = cfg!(fuzzing);
 /// Kind of `u32` value that is being read via `State::U32`.
 #[derive(Debug)]
 enum U32ValueKind {
+    /// First 4 bytes of the PNG signature - see
+    /// http://www.libpng.org/pub/png/spec/1.2/PNG-Structure.html#PNG-file-signature
+    Signature1stU32,
+    /// Second 4 bytes of the PNG signature - see
+    /// http://www.libpng.org/pub/png/spec/1.2/PNG-Structure.html#PNG-file-signature
+    Signature2ndU32,
     /// Chunk length - see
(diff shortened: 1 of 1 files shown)


anforowicz:
@fintelia, can you PTAL?

fintelia:
:+1: This change looks good to me, but I haven't yet had a chance to do a detailed read through. Hopefully should get to that soon

PR #429, opened 2023-11-16
Cap buffer sizes via `ZlibStream::set_max_total_output`.

anforowicz:
PTAL?  This results in quite a significant performance improvement for the `noncompressed-8x8.png` benchmark (bigger improvement than https://github.com/image-rs/image-png/pull/427 and https://github.com/image-rs/image-png/pull/428 combined - I am surprised).

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -1313,6 +1313,16 @@ impl StreamingDecoder {
             }
         };
 
+        if let Some(mut raw_row_len) = color_type.checked_raw_row_length(bit_depth, width) {
+            if interlaced {
+                // This overshoots, but overestimating should be fine.
+                // TODO: Calculate **exact** IDAT size for interlaced images.
+                raw_row_len = raw_row_len.saturating_mul(2);
+            }
+            self.inflater
+                .set_max_total_output((height as usize).saturating_mul(raw_row_len));
+        }
+
         self.info = Some(Info {
             width,
             height,
@@ -1841,7 +1851,7 @@ mod tests {
         let png = {
             let mut png = Vec::new();
             write_fdat_prefix(&mut png, 2, 8);
-            let image_data = generate_rgba8_with_width(8);
+            let image_data = generate_rgba8_with_width_and_height(8, 8);
             write_fdat(&mut png, 2, &image_data[..30]);
             write_fdat(&mut png, 3, &image_data[30..]);
             write_iend(&mut png);
@@ -1891,5 +1901,34 @@ mod tests {
         }
 
         assert_eq!(first_frame, second_frame);
+    }
+
+    #[test]
+    fn test_idat_bigger_than_image_size_from_ihdr() {
+        let png = {
+            let mut png = Vec::new();
+            write_png_sig(&mut png);
+            write_rgba8_ihdr_with_width(&mut png, 8);
+
+            // Here we want to test an invalid image where the `IDAT` chunk contains more data
+            // (data for 8x256 image) than declared in the `IHDR` chunk (which only describes an
+            // 8x8 image).
+            write_chunk(
+                &mut png,
+                b"IDAT",
+                &generate_rgba8_with_width_and_height(8, 256),
+            );
+
+            write_iend(&mut png);
+            png
+        };
+        let decoder = Decoder::new(png.as_slice());
+        let mut reader = decoder.read_info().unwrap();
+        let mut buf = vec![0; reader.output_buffer_size()];
+
+        // TODO: Should this return an error instead?  For now let's just have test assertions for
+        // the current behavior.
+        reader.next_frame(&mut buf).unwrap();
+        assert_eq!(3093270825, crc32fast::hash(&buf));
     }
 }
--- a/src/decoder/zlib.rs
+++ b/src/decoder/zlib.rs
@@ -14,6 +14,10 @@ pub(super) struct ZlibStream {
     out_buffer: Vec<u8>,
     /// The cursor position in the output stream as a buffer index.
     out_pos: usize,
+    /// Limit on how many bytes can be decompressed in total.  This field is mostly used for
+    /// performance optimizations (e.g. to avoid allocating and zeroing out large buffers when only
+    /// a small image is being decoded).
+    max_total_output: usize,
     /// Ignore and do not calculate the Adler-32 checksum. Defaults to `true`.
     ///
     /// This flag overrides `TINFL_FLAG_COMPUTE_ADLER32`.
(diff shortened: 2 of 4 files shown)


src/decoder/stream.rs
```diff
@@ -1229,6 +1229,15 @@ impl StreamingDecoder {
             }
         };
 
+        if let Some(mut raw_row_len) = color_type.checked_raw_row_length(bit_depth, width) {
+            if interlaced {
+                // This overshoots, but overestimating should be fine.
+                raw_row_len = raw_row_len.saturating_mul(2);
```
anforowicz, line 1235: Does this look okay?

Reading https://en.wikipedia.org/wiki/Adam7_algorithm, I see that:

* Pass 1 fills 1/64 of pixels
* Pass 2 fills additional 1/64 of pixels (for a total of 1/32)
* Pass 3 fills additional 1/32 of pixels
* Pass 4 fills additional 1/16 of pixels
* Pass 5 fills additional 1/8 of pixels
* Pass 6 fills additional 1/4 of pixels
* Pass 7 fills additional 1/2 of pixels
 
And this unsurprisingly adds up to 100% of pixels.  But I am not sure if there may be a need for some extra data (more "filter" type bytes;  maybe pixel bits need to be rounded up to the nearest byte), so I thought that mutiplying by 2 would be safest.

anforowicz:
Hmmm... I guess I need to look into the fuzzing failure.  I suspect that IHDR can declare a smaller image size than the real image - I'll try to author a regression test that would do that.

anforowicz:
> Hmmm... I guess I need to look into the fuzzing failure. I suspect that IHDR can declare a smaller image size than the real image - I'll try to author a regression test that would do that.

This is done now.  I am not sure if I need to do something special to kick-off the CI tests again?

fintelia:
Haven't forgotten about this PR. But I've been wanting to finish https://github.com/image-rs/fdeflate/pull/14 first so that we don't have to reserve extra space just to make fdeflate happy

src/decoder/zlib.rs
```diff
@@ -37,9 +42,21 @@ impl ZlibStream {
...
         *self.state = Decompressor::new();
     }
 
+    pub(crate) fn set_max_total_output(&mut self, n: usize) {
+        // Add 2 and then round-up to the next multiple of 8.
+        //
+        // Increasing the original `n` helps to provide a safety margin near the end of output - in
+        // particular it helps to deal with `assert!(output.len() >= output_position + 2)` in
+        // `fdeflate::Decompressor::read`.
+        let n = n.saturating_add(2 + 7) & !7;
```
fintelia, line 55: That assert is gone in newer versions of `fdeflate` so there should be no need to round up anymore

src/decoder/stream.rs
fintelia: Would be nice to precisely figure out how many bytes we need, but I'm OK with leaving this as a TODO for now

anforowicz (commented):
 
Thanks for the `fdeflate` changes.  I've tweaked the PR-under-review to simplify things as allowed by the new `fdeflate` behavior.

FWIW, I've re-measured the impact on the non-compressed 8x8 image and still get 20+% improvement.

src/decoder/zlib.rs
```diff
@@ -101,10 +111,9 @@ impl ZlibStream {
             return Ok(());
         }
 
-        loop {
+        while !self.state.is_done() {
             self.prepare_vec_for_appending();
-
-            let (in_consumed, out_consumed) = self
```
anforowicz, line 107: `in_consumed` is always 0, since (after removal of `in_buffer`) we always pass an empty slice `&[]` to the `read` method on old line 109 /  new line 118.

src/decoder/zlib.rs
```diff
@@ -101,10 +111,9 @@ impl ZlibStream {
             return Ok(());
         }
 
-        loop {
+        while !self.state.is_done() {
```
anforowicz, line 114: I've tweaked the loop to get the `debug_assert!(!self.state.is_done()` inside `prepare_vec_for_appending`.

fintelia (approved):

fintelia:
Left a small comment, but otherwise looks good to me. As a followup I plan to experiment with changing `ZlibStream` to directly write into a caller provided buffer instead of using its own `out_buffer`.

src/decoder/stream.rs
```diff
@@ -1892,4 +1902,33 @@ mod tests {
...
+
+    #[test]
+    fn test_idat_bigger_than_image_size_from_ihdr() {
+        let png = {
+            let mut png = Vec::new();
+            write_png_sig(&mut png);
+            write_rgba8_ihdr_with_width(&mut png, 8);
+
+            // Here we want to test an invalid image where the `IDAT` chunk contains more data
+            // (data for 8 image) than declared in the `IHDR` chunk (which only describes an 8x8
```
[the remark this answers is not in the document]
anforowicz: Ooops.  Done.

anforowicz:
> Left a small comment, but otherwise looks good to me.

Ack.  Thanks for reviewing!

> As a followup I plan to experiment with changing `ZlibStream` to directly write into a caller provided buffer instead of using its own `out_buffer`.

Thanks for looking into this.  Let me try to share a few notes, hoping that it helps (i.e. I am mostly trying to share knowledge and past experience, and definitely *not* trying to influence the changes one way or another - I don't claim to know which way is "right").

First of all - I note that there are interesting interdependencies between the data:

* Final image doesn't have the "filter type" byte at the front of each row, but the raw decompressed data includes such byte.
* Decompressor needs to retain 32kB for looking back.

My early attempts at a similar change can be found in https://github.com/anforowicz/image-png/commit/6071020bf83857ec27ef355cc74d23bd38030c43.  (Although this is slightly different than what you are considering - here I am avoiding passing `&mut Vec<u8>` into `ZlibStream`'s `decompress` and `finish_compressed_chunks` methods, but I still keep `ZlibStream::out_buffer`.)  I've abandoned this attempt because:

* I was using this as a prerequisite for an avoiding extra copies into `prev` row, but this copying has been taken care of with a different, simpler approach - https://github.com/image-rs/image-png/pull/422 (thanks!)
* I was getting mixed performance results (probably because I approached this wrong - I was mixing and measuring multiple related/dependent changes, which made it difficult to understand which changes had a positive/negative impact).

At the same time, removing or replacing the `&mut Vec<u8>` parameter from `ZlibStream`'s methods can result in some code simplification (and therefore it may be desirable even if it is performance-neutral).  For example, there are some callers of `ZlibStream::decompress` which pass `&mut vec![]` or `Vec::new` (`read_header_info` and `finish_decoding` in `mod.rs`).

PR #430, opened 2023-11-18
Option to ignore ancillary chunks with invalid CRC

sophie-h:
Skipping non-critical chunks if the CRC check fails is the default it libpng. Therefore adding this
option and enabling it by default.

image-rs/image#2057

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -385,6 +385,7 @@ pub struct DecodeOptions {
     ignore_adler32: bool,
     ignore_crc: bool,
     ignore_text_chunk: bool,
+    skip_ancillary_crc_failures: bool,
 }
 
 impl Default for DecodeOptions {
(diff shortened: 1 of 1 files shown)


src/decoder/stream.rs
```diff
@@ -647,6 +664,12 @@ impl StreamingDecoder {
                             } else {
                                 Ok((1, Decoded::ChunkComplete(val, type_str)))
                             }
+                        } else if self.decode_options.skip_ancillary_crc_failures
+                            && !chunk::is_critical(type_str)
+                        {
+                            // Ignore ancillary chunk with invalid CRC
+                            self.state = Some(State::U32(U32ValueKind::Length));
+                            Ok((1, Decoded::Nothing))
```
sophie-h, line 672: O don't understand why it's a `1` here. Is that correct?

fintelia: The `1` indicates that a single additional byte of the input has been consumed, which is correct

fintelia (approved):

anforowicz:
@sophie-h - would you be able to add test coverage for the new option?  I am doing some refactoring in https://github.com/image-rs/image-png/pull/428 and the test coverage would be an extra reassurance that I've correctly resolved the merge conflicts (both of our PRs edited `stream.rs`).

PR #434, opened 2023-12-02
new_with_info construtor for encoder

DeniseBryson:
Hi,
I added a new constructor for the encoder. You can pass the info struct to create a new encoder. With that you can decode an image change something and encode it again more easily.
I also added an example to shows how to use the function, called change-png-info.rs .
Thanks to @hoijui for the help.

### DIFF
--- a/examples/change-png-info.rs
+++ b/examples/change-png-info.rs
@@ -1,0 +1,51 @@
+// For reading and opening files
+use std::fs::File;
+use std::io::BufWriter;
+use std::path::Path;
+pub type BoxResult<T> = Result<T, Box<dyn std::error::Error + Send + Sync>>;
+
+fn main() -> BoxResult<()> {
+    //Add image called test.png in the target folder of the library
+    let path = Path::new(r"./target/test.png");
+    //Decode
+    // The decoder is a build for reader and can be used to set various decoding options
+    // via `Transformations`. The default output transformation is `Transformations::IDENTITY`.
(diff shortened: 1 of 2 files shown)


examples/change-png-info.rs
```diff
@@ -0,0 +1,53 @@
+// SPDX-FileCopyrightText: 2023 Lukas Schattenhofer <lukas.schattenhofer@ose-germany.de>
+// SPDX-License-Identifier: MIT OR Apache-2.0
```
fintelia, line 2: Please remove. We track authorship via git history and all source code in the repository is MIT OR Apache-2.0 as per the LICENSE files in the root directory

src/encoder.rs
```diff
@@ -169,6 +169,14 @@ impl<'a, W: Write> Encoder<'a, W> {
         }
     }
 
+    pub fn new_with_info(w: W, info: Info<'a>) -> Encoder<'a, W> {
```
fintelia, line 172: Please name this `with_info` to be consistent with the "General constructors" entry in the [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case).

fintelia (commented):
Thanks for this PR! This will make it much easier to re-encode with the same metadata as an existing image

I see the CI is failing due to a formatting issue. You should be able to fix that by running `cargo fmt`

DeniseBryson:
Thanks for the comments, should be all fixed now.

PR #435, opened 2023-12-04
Regression test for invalid CRC ancillary chunks

sophie-h:
Issue #430

Not sure how to test for failure with checks enabled.

cc @anforowicz


fintelia (approved):

anforowicz:
Thanks - this is great!

> Not sure how to test for failure with checks enabled.

You could add a `#[test]` in `src/decoder/mod.rs` which `include_bytes!("../../tests/bugfixes/issue#430.png")` and then in the test configure the non-default options for the decoder + check that the expected result gets returned.  It seems a lower priority though, so it's probably fine to postpone it or do it in a separate PR.

fintelia:
FYI: [GitHub will automatically close an issue when a PR is merged](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) if the PR description says something like "Fixes #430"

PR #437, opened 2023-12-10
Update `unstable` nightly portable-simd imports

okaneco:
Import `SimdOrd`, `SimdPartialEq` from `std::simd::cmp`
Import `SimdInt`, `SimdUint` from `std::simd::num`

The most recent `portable-simd` subtree sync moved some traits out of the top level modules.


fintelia:
Thanks!

PR #442, opened 2023-12-17
Validate provided info in Encoder::with_info

fintelia:
This would normally be a breaking change, but the Encoder::with_info API hasn't yet been included in a release.

### DIFF
--- a/examples/change-png-info.rs
+++ b/examples/change-png-info.rs
@@ -29,7 +29,7 @@ fn main() -> BoxResult<()> {
 
     // Edit previous info
     info_out.interlaced = info_default.interlaced;
-    let mut encoder = png::Encoder::with_info(w, info_out);
+    let mut encoder = png::Encoder::with_info(w, info_out)?;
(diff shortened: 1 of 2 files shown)


hoijui:
Good call!
I now imagine something like `Info::validate(&self) -> Result<(), ...>`, which is called from withing `Encoder::with_info()`.
What I do not understand though, is how to prevent a lot of code duplication between this `validate` function
and all the `set_*` functions in `Encoder`, as the former would validate all at once, and the later validates parts separately.
Would we want to have separate `Info::validate_*` functions as well, and call the appropriate one from each `Encoder::set_*` method, and also call all of them in sequence from `Info::validate()`?

hoijui:
ouh you already did this in code.. sorry, did not see it is a PR *ups*.
... Is that already all the validation necessary?

fintelia:
This is all the validation we were already doing in other cases. Not sure if any further validation makes sense

PR #445, opened 2023-12-31
Reject zero sized images

### LINKED ISSUE #438: Crash when width=0
oss-fuzz detected a crash when given a PNG with a zero-sized width. We should reject any image that has a width or height of zero.

 Test case: 
clusterfuzz-testcase-minimized-decode-4787215086256128

backtrace:
```
#11 0x5a3b7cac04f0 in core::slice::_$LT$impl$u20$$u5b$T$u5d$$GT$::chunks_exact_mut::h0e0c1e7467f87015 /rustc/d12c6e947ceacf3b22c154caf9532b390d8dc88a/library/core/src/slice/mod.rs:1225:9
#12 0x5a3b7cac04f0 in png::decoder::Reader$LT$R$GT$::next_frame::h42665d048034458e [image-png/src/decoder/mod.rs:525](https://github.com/image-rs/image-png/blob/3b3f2e379040a029d14487ea2eaae1da4d477bd6/src/decoder/mod.rs#L525):24
#13 0x5a3b7cad8dbe in decode::png_decode::h8eccdf7a9c7ab6d8 [image-png/fuzz/fuzz_targets/decode.rs:18](https://github.com/image-rs/image-png/blob/3b3f2e379040a029d14487ea2eaae1da4d477bd6/fuzz/fuzz_targets/decode.rs#L18):26

### LINKED ISSUE #444: Panic on decoding a crafted image: `chunk size must be non-zero`
The following file found by the in-tree fuzzer results in a panic when decoded:

```
thread '<unnamed>' panicked at image-png/src/decoder/mod.rs:534:18:
chunk size must be non-zero
```

[crash-127a81c8af6d6817feb3c560e0d45c80047ac6d5.png.gz](https://github.com/image-rs/image-png/files/13799528/crash-127a81c8af6d6817feb3c560e0d45c80047ac6d5.png.gz)

(gzipped so that github would accept the upload)

fintelia:
The PNG spec states:

> Width and height give the image dimensions in pixels. They are PNG four-byte unsigned integers. Zero is an invalid value.

Fixes #444 
Fixes #438


CryZe:
Wouldn't it make sense to change the types to `NonZeroU32` then?

fintelia:
Switching to `NonZeroU32` would be an API breaking change. There's also a _lot_ of places throughout the image-rs collection of crates that work with image widths and heights represented as normal integers rather than non-zero integers, so that would have to be a broader discussion than just this one place

Shnatsel:
The image in the unit test declares the size of `0 x 3`, so the decoder is entirely correct in rejecting it.

I've changed the width from 0 to 1 using a hex editor and the file is still broken, just in a different way.

Looking at [the PR that introduced it](https://github.com/image-rs/image-png/pull/216/), it seems it was original file was generated by a fuzzer. I don't think it would be feasible to craft a file that triggers the same now-fixed issue, at least not without using a fuzzer and tying it to the implementation details of the library again.

I propose to simply remove the failing test, since there is no reasonable way for us to keep it _and_ fix the 0-size panic.

PR #450, opened 2024-01-09
Adds `Reader::finish` to read all the auxillary chunks that comes after the image

### LINKED ISSUE #343: Text chunks after image data
Hello, thank you for the excellent crate and related documentation. I am having trouble reading text chunks, specifically `tEXt` chunks, that appear after the image data chunks but before the image end chunk. According to the documentation for the [text_metadata](https://docs.rs/png/latest/png/text_metadata/index.html) module:

>They may appear either before or after the image data chunks.

Then, in the [Reading text chunks](https://docs.rs/png/latest/png/text_metadata/index.html#reading-text-chunks) section, there is the following example:

```rust
use std::fs::File;
use std::iter::FromIterator;
use std::path::PathBuf;

// Opening a png file that has a zTXt chunk
let decoder = png::Decoder::new(
    File::open(PathBuf::from_iter([
        "tests",
        "text_chunk_examples",
        "ztxt_example.png",
    ]))
    .unwrap(),
);
let mut reader = decoder.read_info().unwrap();
// If the text chunk is before the image data frames, `reader.info()` already contains the text.
for text_chunk in &reader.info().compressed_latin1_text {
    println!("{:?}", text_chunk.keyword); // Prints the keyword
    println!("{:#?}", text_chunk); // Prints out the text chunk.
    // To get the uncompressed text, use the `get_text` method.
    println!("{}", text_chunk.get_text().unwrap());
}
```

When I implement this example with the file name replaced with the two example images, the text metadata works for the image with the text chunks before the image data, but it does not work for the PNG with the text chunks after the image data, the text chunks STDOUT is empty.

The above example does have the following comment:

```rust
// If the text chunk is before the image data frames, `reader.info()` already contains the text.
```

and the [read_info](https://docs.rs/png/latest/png/struct.Decoder.html#method.read_info) documentation:

```
Reads all meta data until the first IDAT chunk
```

The behavior for the PNG with the image data after the IDAT chunk is expected since this chunk has not been read. When I run the `pngcheck` example, the text chunks are found for both PNGs. The `pngcheck` example uses a `StreamingDecoder`, but this appears more complicated and lower level.

Sorry for the long-winded question, but do I need to implement `StreamingDecoder` in order to ensure all of the text chunks have been read, regardless of location relative to the IDAT chunk? Alternatively, how do I continue the decoding/reading after all frames have been read but there are post IDAT chunks/pre IEND chunks without using the `StreamingDecoder`?

I am more than willing to implement `StreamingDecoder`, but I just wanted to check if there was a higher level interface that I could use and if I am missing something.

text_after_IDAT
text_before_IDAT

jupyterkat:
Closes #343

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -598,6 +598,30 @@ impl<R: Read> Reader<R> {
         }))
     }
 
+    /// Read the rest of the image and chunks and finish up, including text chunks or others, will discard the rest of the image data
+    pub fn finish(&mut self) -> Result<(), DecodingError> {
+        self.next_frame = SubframeIdx::End;
+        self.data_stream.clear();
+        self.current_start = 0;
+        self.prev_start = 0;
+        loop {
+            let mut buf = Vec::new();
+            let state = self.decoder.decode_next(&mut buf)?;
+
+            match state {
+                Some(Decoded::ImageEnd) => break,
+                None => {
+                    return Err(DecodingError::Format(
+                        FormatErrorInner::MissingImageData.into(),
+                    ))
+                }
+                _ => {}
+            }
+        }
+
+        Ok(())
+    }
+
     /// Fetch the next interlaced row and filter it according to our own transformations.
     fn next_interlaced_row_impl(
         &mut self,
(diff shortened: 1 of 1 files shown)


fintelia:
Thanks!

PR #447, opened 2024-01-05
Avoid 32kB decompression lag + compact less often.

anforowicz:
Avoiding 32kB decompression lag
===============================

Before this commit, decompressed data would be accumulated in `ZlibStream::out_buffer` and returned via `image_data` with 32kB lag corresponding to `CHUNCK_BUFFER_SIZE`:

    ```
    fn transfer_finished_data(&mut self, image_data: &mut Vec<u8>) -> usize {
        let safe = self.out_pos.saturating_sub(CHUNCK_BUFFER_SIZE);
        image_data.extend(self.out_buffer.drain(..safe));
        ...
    ```

32kB is a typical size of L1 cache, so the lag would mean that the data passed to `image_data.extend(...)` would be already cold and evicted from the L1 cache.

This commit avoids the lag by always returning into `image_data` all the data from `out_buffer` (i.e. data up to `out_pos`):

    ```
    fn transfer_finished_data(&mut self, image_data: &mut Vec<u8>) -> usize {
        let transferred = &self.out_buffer[self.read_pos..self.out_pos];
        image_data.extend_from_slice(transferred);
        self.read_pos = self.out_pos;
        ...
    ```

Compacting less often
=====================

The changes above mean that `Vec::drain` no longer compacts `out_buffer`. Therefore this commit also refactors how this compaction works.

Before this commit, not-yet-returned data would be shifted to the beginning of `out_buffer` every time `transfer_finished_data` is called. This could potentially mean that for 1 returned byte, N bytes have to be copied during compaction.

After this commit, compaction is only done when the compaction cost if offset by many read bytes - for 3 returned bytes 1 byte has to be copied during compaction.

Performance impact
==================

The commit has a positive impact on performance, except for:

* `decode/Transparency.png` - regression between 15% and 20% is reported in 3-out-of-3 measurements.
* `decode/kodim17.png` - a regression of 2.1% has been reported in 1-out-of-3 measurements (an improvement of 0.6% - 1.13% has been reported in the other 2-out-of-3 measurements).
* `generated-noncompressed-64k-idat/128x128.png` - a regression of 25% has been reported in 1-out-of-3 measurements (an improvement of 21% - 29% has been reported in the other 2-out-of-3 measurements).

The results below have been gathered by running the `decoder` benchmark. First a baseline was saved before this commit, and then a comparison was done after the commit.  This (the baseline + the comparison) was repeated a total of 3 times.  All results below are for the relative impact on the runtime.  All results are with p = 0.00 < 0.05.

[...]

Issue #286, opened 2021-02-19
Zlib decompression may exceed allocation limits


HeroicKatora:
What about doing the actual capping? It seems like finish_compressed_chunks() will happily keep allocating as much as the the zlib stream wants.

_Originally posted by @jrmuizel in https://github.com/image-rs/image-png/pull/285#discussion_r578756052_

That is, the caller can not control the final size which makes the limits set in `Decoder::limits` ineffective.

PR #452, opened 2024-01-10
Use a single Vec between Decoder and ZlibStream

fintelia:
This simplifies buffer management and speeds up decoding. Still some lingering bugs/TODOs. Also need to figure out API considerations as right now this is a breaking change

### DIFF
--- a/examples/pngcheck.rs
+++ b/examples/pngcheck.rs
@@ -174,7 +174,7 @@ fn check_image<P: AsRef<Path>>(c: Config, fname: P) -> io::Result<()> {
             }
             buf = &data[..n];
         }
-        match decoder.update(buf, &mut Vec::new()) {
+        match decoder.update(buf, &mut Vec::new(), &mut 0) {
             Ok((_, ImageEnd)) => {
                 if !have_idat {
                     // This isn't beautiful. But it works.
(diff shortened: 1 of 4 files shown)


Shnatsel:
This `Vec` can also exceed allocation limits: #286 I was about to work on adding support for limits, but I see there are in-flight PRs touching this code, so I'll wait until this lands.

I'm excited to see this refactoring! Let me know if you want me to do a before/after test run on a large image corpus, or if there's any other way I can help test this.

anforowicz:
Very interesting - thanks for looking into this!  AFAIU `fdeflate`'s requirement for retaining 32kB **unmodified** decompressed data is met by only calling `unfilter` on data that was decompressed **earlier**.  Did I get that right?  If so, this PR is to some extent reverting the changes from https://github.com/image-rs/image-png/pull/447 but in exchange this PS enables getting rid of an additional buffer / memcpy hop, right?

fintelia:
Yes, but this enables another trick: for PNGs with less than 256 KB of uncompressed image data, it does all the decompression in one go. Which means that for those images we don't have to save any space for lookback because we'll be completely done processing the deflate stream by the time we start modifying the data using `unfilter`.

Issue #393, opened 2023-04-01
Very slow decoding of paletted PNG images


Shnatsel:
[`zune-png` benchmarks](https://etemesi254.github.io/assets/criterion/png_%20PNG%20decoding%20palette%20image/report/index.html) show that the `png` crate is much slower than other decoders on indexed images - a whopping 3x slower than `zune-png`.

[CPU profile](https://share.firefox.dev/3MqTcun) shows that 71% of the time is spent in `png::utils::unpack_bits`, specifically [this hot loop](https://github.com/image-rs/image-png/blob/2f53fc40b91a5e1d0ad1801b746c04a7fe1d8603/src/utils.rs#L35-L37).

The code is full of indexing and doesn't look amenable to autovectorization. I think the entire function will have to be rewritten; it's probably a good idea to copy `zune-png` here.

Shnatsel:
Measured on `v0.17.8-rc` with this image: [speed_bench_palette](https://user-images.githubusercontent.com/291257/229297591-30e20b14-62f9-42b9-978a-563b66db5578.png)

Shnatsel:
`zune-png` uses vectorization-friendly implementation of expansion, which explains the difference in performance. It can be found here: https://github.com/etemesi254/zune-image/blob/60fbeef1e58a7dec08e27bf2ed190e061f27cb45/zune-png/src/utils.rs#L1

fintelia:
The version in this crate handles paletted images with 1/2/4-bit indices, whereas `zune-png` seems not to. However, probably the overwhelming majority of images use 8-bit indices... so it is probably worth special casing it.

Shnatsel:
Based on [this comment](https://github.com/etemesi254/zune-image/blob/60fbeef1e58a7dec08e27bf2ed190e061f27cb45/zune-png/src/utils.rs#L169-L171) I understand `zune-png` expands the bits to bytes first, and then resolves those to the correct palette values in another pass.

So it does handle lower bit depth indices correctly, but they're expanded to 8 bits before use.

fintelia:
This reminded me that I started working on more refactoring which included changing the transformations to not operate in place. Just created https://github.com/image-rs/image-png/pull/396 with the current state

anforowicz:
Looking at top-500 website, paletted images account for 35% of PNG images.  Source:
* https://github.com/image-rs/image-png/discussions/416#discussioncomment-7436871
[...]

PR #453, opened 2024-01-12
End-to-end decoding benchmarks of paletted PNG images.

anforowicz:


anforowicz:
@fintelia, can you PTAL? 

Special thanks to @Shnatsel for [pointing out](https://github.com/image-rs/image-png/issues/393#issuecomment-1888019052) the mistake I made in my earlier attempt to cover `expand_paletted` in end-to-end benchmarks.  Thanks!

fintelia:
Could we use a smaller test image? 8 MB is pretty large for a file checked into the git repo

Shnatsel:
FWIW the larger the decompressed image is, the more the palette expansion pass dominates the runtime. On 1024x1024 images it's about 20% of the time, on the map image it's about 40%, on the zune-png benchmarking image shared at the start of #393 it's about 60%.

I think it is possible to create an image e.g. with large areas of flat fill that would be small when compressed, decompress into an image with large dimensions, and have the palette pass account for a large portion of the image decoding time.

Or maybe the palette pass should just be benchmarked separately.

Shnatsel:
I suggest using this image instead: [Stadt_Onex_2021_posterized_med](https://github.com/image-rs/image-png/assets/291257/82d1d171-2184-47f3-b933-3cdc7bac78c8)

It still spends 40% of the time in palette expansion while only being 1.4MB in size (as opposed to the 8MB image proposed originally).

anforowicz:
FWIW, I feel that it is desirable to have a somewhat realistic paletted test image that also highlights the performance of `expand_paletted`.  FWIW using palette reduces the size of the image from 9,408,318 bytes to 8,396,692 (and so it seems reasonable that this optimization would be used in the "real world" + shows that the original "real world" image is even bigger).

OTOH, if we decide that 8MB is unreasonable, then we can try testing with existing benchmarks and a function-level benchmark that I am working on at https://github.com/anforowicz/image-png/tree/palette-benchmarks-func2.  We would still want to land the current PR to ensure that `expand_paletted` is actuall covered by the benchmarks. (I think that other set of changes is ready for review, but I'll hold off with opening a pull request until we've figure out what to do with the current one under review.  This is mostly because the performance measurements in the other one use the current PR as the baseline.)

anforowicz:
> I suggest using this image instead: [Stadt_Onex_2021_posterized_med](https://github.com/image-rs/image-png/assets/291257/82d1d171-2184-47f3-b933-3cdc7bac78c8)
> 
> It still spends 40% of the time in palette expansion while only being 1.4MB in size (as opposed to the 8MB image proposed originally).

Thanks - let me switch to this image.

anforowicz:
> > I suggest using this image instead: [Stadt_Onex_2021_posterized_med](https://github.com/image-rs/image-png/assets/291257/82d1d171-2184-47f3-b933-3cdc7bac78c8)
> > It still spends 40% of the time in palette expansion while only being 1.4MB in size (as opposed to the 8MB image proposed originally).
> 
> Thanks - let me switch to this image.

Actually, I wasn't able to [find](https://commons.wikimedia.org/w/index.php?search=Stadt_Onex_2021_posterized_med&title=Special:MediaSearch&go=Go&type=image) the image that you are referring to.  But maybe this is fine - I just grabbed https://commons.wikimedia.org/wiki/File:Stadt_Onex_2021.png at a slightly smaller resolution and converted to `png:IHDR.color_type: 3 (Indexed)` using ImageMagick.  The resulting file has 1,146,578 and `png::decoder::expand_paletted` takes 22.84% of time.

anforowicz:

PR #469, opened 2024-02-12
Update the `bitflags` dependency to `2.x.x`.

anforowicz:


anforowicz:
This PR will help to ensure that in my dependency graph there is only a single version of `bitflags`.  Having v1 and v2 is not the end of the world, but it seems slightly nicer to be able to just depend on v2 of `bitflags`.

Also - after sending this PR out, I've realized that removing `Ord` and `PartialOrd` may make some sense from purity perspective, but having these traits is useful (e.g. if a client of the `png` crate ever wanted to put `Transformations` in `BTreeSet`) and avoids having to (somewhat unnecessarily) bump up the semver version of the `png` crate.  I'll go ahead and try to fix this.

anforowicz:
> Also - after sending this PR out, I've realized that removing `Ord` and `PartialOrd` may make some sense from purity perspective, but having these traits is useful (e.g. if a client of the `png` crate ever wanted to put `Transformations` in `BTreeSet`) and avoids having to (somewhat unnecessarily) bump up the semver version of the `png` crate. I'll go ahead and try to fix this.

This is done now.

anforowicz:
@fintelia, can you PTAL?

fintelia:
This looks good to me, but has some overlap with #400.

Hoping to start working on a png `0.18` release in the next few weeks once image `0.25` is released.

PR #475, opened 2024-02-22
Ignore iCCP section if it is invalid

nathaniel-daniel:
This PR makes invalid iCCP sections get ignored instead of throwing an error. Let me know if I should add an option to toggle this behavior.

Ideally, I think the `icc_profile` field should be an `Option<Result<_, _>>`, but that would be breaking. To ensure duplicate iccp sections are handled correctly, I introduced a new `have_iccp` field.


fintelia:
This might change if we switch to lazy loading of metadata chunks, but looks good to me

PR #478, opened 2024-03-10
Parallel image encoding

fintelia:
Adds a feature flag (`rayon`) that enables parallel encoding of PNGs. This drastically reduces the time taken to encode a single image. With target-cpu=native, my 6 core/12 thread CPU is able to exceed **1 billion pixels/second** encoding speed on the QOI benchmark suite:

```
                                              Ratio             Encode
---------                                    -------     --------------------
Single-threaded                              27.861%     245 mps  0.829 GiB/s
Single-threaded (target-cpu=native)          27.861%     308 mps  1.042 GiB/s

Parallel                                     27.861%     917 mps  3.099 GiB/s
Parallel (target-cpu=native)                 27.861%    1084 mps  3.664 GiB/s
```

Depends on https://github.com/image-rs/fdeflate/pull/23.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ crc32fast = "1.2.0"
 fdeflate = { path = "../fdeflate" }
 flate2 = "1.0.11"
 miniz_oxide = { version = "0.7.1", features = ["simd"] }
-rayon = "1.9.0"
+rayon = { version = "1.9.0", optional = true }
 
 [dev-dependencies]
 byteorder = "1.5.0"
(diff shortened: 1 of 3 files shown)


Shnatsel:
FWIW https://crates.io/crates/mtpng exists and can be used as inspiration/reference

fintelia:
I should add that this code is fully functional. I abandoned the PR because I was burnt out dealing with complaints from entitled users, and it felt like a headache to design an API where users could get the benefits without having to specifically opt-in, but leaving a way to force single-threaded encoding for those who wanted it

Some1and2-XC:
@fintelia I feel as though this still would be a great feature to add. If you don't want to, I'm willing to try take a swing at getting this working. What kind of API would this need to not cause headaches for maintainers? Also, what kinds of issues were users having?

And truthfully I don't think its such a bad thing for something like this to be exclusively opt-in.

PR #484, opened 2024-06-29
Remove usages of `extern crate`

waywardmonkeys:


waywardmonkeys:
I'm not able to build the fuzz targets here at the moment (only Apple Silicon hardware), but I'm pretty sure I fixed them correctly (and marked them appropriately as being `edition = 2018`).

kornelski:
Thanks

PR #485, opened 2024-06-29
Fix typos.

### LINKED ISSUE #357: Typo in message for NoMoreImageData
There is typo in: https://github.com/image-rs/image-png/blob/master/src/decoder/stream.rs#L296
"IDAT or fDAT chunk is has not enough data for image."

waywardmonkeys:
Fixes #357.


kornelski:
It seems that `CHUNCK_BUFFER_SIZE` isn't publicly accessible, so it's not a silly semver break.

PR #488, opened 2024-07-11
Add compression level

marcosc90:
This PR adds `Compression::Level(u32)` to allow finer control over compression settings. It maps directly to [`flate2::Compression::new(level)`](https://docs.rs/flate2/1.0.30/src/flate2/lib.rs.html#195-197).

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -325,6 +325,8 @@ pub enum Compression {
         note = "use one of the other compression levels instead, such as 'fast'"
     )]
     Rle,
+    /// Integer representing the compression level, typically on a scale of 0-9
+    Level(u32),
 }
 
 impl Default for Compression {
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1702,6 +1702,7 @@ impl Compression {
             Compression::Default => flate2::Compression::default(),
             Compression::Fast => flate2::Compression::fast(),
             Compression::Best => flate2::Compression::best(),
+            Compression::Level(level) => flate2::Compression::new(level),
             #[allow(deprecated)]
             Compression::Huffman => flate2::Compression::none(),
             #[allow(deprecated)]


fintelia:
Could you say a bit about why you want this functionality? 

One weird aspect of this PR is that `CompressionLevel::Level(1)` is actually different from `CompressionLevel::Fast` due to this match statement:
https://github.com/image-rs/image-png/blob/2cfde02d3a50c531e3160023d1e4b0ae92869f4a/src/encoder.rs#L696-L697

Another complication is that `CompressionLevel` isn't marked an non_exhaustive so this needs to wait for the next breaking release.

marcosc90:
Apologies for the short PR description was a bit late for me when I opened it.

> Could you say a bit about why you want this functionality?

I'm building a Wasm package for compressing PNG, mostly canvas-generated images which lack compression. And for my specific case, QR Images, `Fast` increases the image size, while `Default` is a bit slow, using `Compression::Level(2-3)` gives me the best ratio between image size and speed.

Shnatsel:
I also need this for the drop-in replacement for `imagemagick` that I am building.

Shnatsel:
Could we instead add this as a new method on the encoder, like `set_deflate_compression()`, and avoid the API break that way?

0rvar:
I would also like this option, since the "Best" compression level in this crate does not actually use maximum compression, and there are no other variants that are higher either:
```rust
    /// Higher compression level
    ///
    /// Best in this context isn't actually the highest possible level
    /// the encoder can do, but is meant to emulate the `Best` setting in the `Flate2`
    /// library.
    Best,
```

HeroicKatora:
Given that `None` is the _same_ API breaking change, it seems fine to me to bump version here. If any action results from this, it should be considering whether we want the compression enum to be `non_exhaustive` to avoid such issues in the future including the ability to deprecate some levels if semantic problems such as `Best` ever come up. Of course, anything that can be observed will be relied on but at least it gives us slightly more leverage.

fintelia:
The documentation on `Best` is actually wrong. It corresponds to flate2 level 9, which is the highest that the encoder supports.

As far as the overall API design, I've held off responding because I didn't see any elegant way to design this API that wouldn't be confusing / risk making future changes harder:
* One option would be to name the variant `Flate2Level` and make it specific to the levels that the flate2 crate exposes. However, this would it really awkward to switch to a different compression backend (you can see the deprecated variants that are lingering from the last time we switched...) 
* We could make it a 0-10 range with 0=fdeflate-stored, 1=fdeflate-normal, and remap `png::level(N) -> flate2::level(N-1)`. This would be logical for now, but what would we do if we wanted to add more fdeflate levels? Would renumbering the levels be considered backwards compatible?
* We could use a "standard" 0-9 range with the same first few levels as above but only skip one of the flate2 levels. Would give us the precedent for inserting new levels later. Basically expands the current three `{Fast, Default, Best}` levels into a 10-level system

kornelski:
It's too detailed, and backend-specific. We could add another preset if you really need something in between the existing ones.

Shnatsel:
I'm still going to need detailed compression presets for `wondermagick`, simply because I need to conform to an existing API that exposes these knobs. Should I fork this crate? I don't think that's a good idea, and it's probably not going to fly with Linux distros for packaging reasons.

And I do believe optionally giving more control over the compression levels to the API user is ultimately a good thing, as long as it is an "advanced" API that users are not required to familiarize themselves with for basic usage.

fintelia:
The [compression settings in imagemagick](https://github.com/ImageMagick/ImageMagick/blame/de884e5cf43d3580f33a4a5214c983474ba4a25c/coders/png.c#L10279-L10309) are deeply tied to zlib / libpng's capabilities. Notably in addition to the 0-9 levels, they also include `Z_HUFFMAN` and `Z_RLE` which aren't exposed by the `flate2` crate. But when it comes to numeric levels, I'm not even sure what guarantees the API even has? Would it be valid to map 1-3 to `Fast`, 4-6 to `Default`, and 7-9 to `Best`?

I do think it is worth adding at least one compression level between `Fast` and `Default`.

fintelia:
> The only question is, what do we do about fdeflate's mode that doesn't use a fixed dictionary? I want to benchmark it against miniz_oxide and zlib-rs, but I don't think it's exposed in any way right now? Is there a way I can actually get to it through the png crate for benchmarking purposes?

It hasn't been written yet... so it's a bit early for that sort of benchmarking :smile:

PR #490, opened 2024-07-29
StoredOnlyCompressor feature is here, and we can now access it.

arichnad:
StoredOnlyCompressor feature is here, and we can now access it:

> Why

Creating a png with `Compression::None` makes file creation much faster than `Compression::Fast`.

For hard to compress images, file size between `Compression::None` and `Compression::Fast` is is similar.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -694,7 +694,7 @@ impl<W: Write> Writer<W> {
         let adaptive_method = self.options.adaptive_filter;
 
         let zlib_encoded = match self.info.compression {
-            Compression::None => self.stored_only_compressor(data, in_len)?,
+            Compression::None => Self::stored_only_compressor(data, in_len)?,
             Compression::Fast => {
                 let mut compressor = fdeflate::Compressor::new(std::io::Cursor::new(Vec::new()))?;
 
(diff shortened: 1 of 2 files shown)


HeroicKatora (approved):
Breaking extension to the enumeration but this is good. Just need to bump the version for the next release, I suppose. Let me check if a point-release should be done before merging.

src/encoder.rs
```diff
@@ -817,6 +812,15 @@ impl<W: Write> Writer<W> {
         Ok(())
     }
 
+    fn stored_only_compressor(&self, data: &[u8], in_len: usize) -> Result<Vec<u8>> {
```
fintelia, line 815: This method doesn't use `self`, so it shouldn't take it by reference

arichnad:
I've addressed your comments and rebased onto master.  Thanks @fintelia and @HeroicKatora !

fintelia:
Looks good. We'll just want to hold off merging until we've gotten out a minor release with all the non-breaking changes we want to lank

Shnatsel:
This has been implemented as `Compression::NoCompression` variant in #503, which is now merged, so I'm going to go ahead and close this PR.

Despite this exact PR not getting merged, it has informed #503 significantly and the `NoCompression` option would not have been included without this PR. So thank you!

arichnad:
Thank you, @Shnatsel !

PR #493, opened 2024-08-16
Add `Reader.read_row` for decoding into caller-provided buffer.

anforowicz:
Add `Reader.read_row` for decoding into caller-provided buffer.

`Reader.read_row` is an alternative to `Reader.next_row`.

* `Reader.next_row` is convenient when the caller doesn't want to manage
  their own buffer, and doesn't mind having extended borrows on `Reader`
  to accommodate `Row` data.
* OTOH the new `Reader.read_row` avoids an extra copy in some scanarios
  which may lead to better runtime performance - see the benchmark
  results below.

The PR adds a `row-by-row/128x128-4k-idat` benchmark (initially based on
`Reader.next_row` which requires an extra copy).  Using the new         
`Reader.read_row` results in the following performance gains in the
benchmark:

time:   [-16.414% -16.196% -15.969%] (p = 0.00 < 0.05)
time:   [-15.570% -15.218% -14.856%] (p = 0.00 < 0.05)
time:   [-16.101% -15.864% -15.629%] (p = 0.00 < 0.05)

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -483,14 +483,45 @@ impl<R: Read> Reader<R> {
         Ok(())
     }
 
-    /// Returns the next processed row of the image
+    /// Returns the next processed row of the image (discarding `InterlaceInfo`).
+    ///
+    /// See also [`Reader.read_row`], which reads into a caller-provided buffer.
     pub fn next_row(&mut self) -> Result<Option<Row>, DecodingError> {
         self.next_interlaced_row()
             .map(|v| v.map(|v| Row { data: v.data }))
     }
 
-    /// Returns the next processed row of the image
+    /// Returns the next processed row of the image.
+    ///
+    /// See also [`Reader.read_row`], which reads into a caller-provided buffer.
     pub fn next_interlaced_row(&mut self) -> Result<Option<InterlacedRow>, DecodingError> {
+        let mut output_buffer = mem::take(&mut self.scratch_buffer);
+        output_buffer.resize(self.output_line_size(self.info().width), 0u8);
+        let result = self.read_row(&mut output_buffer);
+        self.scratch_buffer = output_buffer;
+        result.map(move |option| {
+            option.map(move |interlace| {
+                let output_line_size = self.output_line_size_for_interlace_info(&interlace);
+                InterlacedRow {
+                    data: &self.scratch_buffer[..output_line_size],
+                    interlace,
+                }
+            })
+        })
+    }
+
+    /// Reads the next row of the image into the provided `output_buffer`.
+    /// `Ok(None)` will be returned if the current image frame has no more rows.
+    ///
+    /// `output_buffer` needs to be long enough to accommodate [`Reader.output_line_size`] for
+    /// [`Info.width`] (initial interlaced rows may need less than that).
+    ///
+    /// See also [`Reader.next_row`] and [`Reader.next_interlaced_row`], which read into a
+    /// `Reader`-owned buffer.
+    pub fn read_row(
+        &mut self,
+        output_buffer: &mut [u8],
+    ) -> Result<Option<InterlaceInfo>, DecodingError> {
         let interlace = match self.subframe.current_interlace_info.as_ref() {
             None => {
                 self.finish_decoding()?;
@@ -507,24 +538,21 @@ impl<R: Read> Reader<R> {
                 self.info().raw_row_length_from_width(width)
             }
         };
+
+        let output_line_size = self.output_line_size_for_interlace_info(&interlace);
+        let output_buffer = &mut output_buffer[..output_line_size];
+
+        self.next_interlaced_row_impl(rowlen, output_buffer)?;
+
+        Ok(Some(interlace))
+    }
+
+    fn output_line_size_for_interlace_info(&self, interlace: &InterlaceInfo) -> usize {
         let width = match interlace {
-            InterlaceInfo::Adam7(Adam7Info { width, .. }) => width,
+            InterlaceInfo::Adam7(Adam7Info { width, .. }) => *width,
             InterlaceInfo::Null(_) => self.subframe.width,
         };
-        let output_line_size = self.output_line_size(width);
-
-        // TODO: change the interface of `next_interlaced_row` to take an output buffer instead of
-        // making us return a reference to a buffer that we own.
-        let mut output_buffer = mem::take(&mut self.scratch_buffer);
-        output_buffer.resize(output_line_size, 0u8);
-        let ret = self.next_interlaced_row_impl(rowlen, &mut output_buffer);
-        self.scratch_buffer = output_buffer;
-        ret?;
-
-        Ok(Some(InterlacedRow {
-            data: &self.scratch_buffer[..output_line_size],
-            interlace,
-        }))
+        self.output_line_size(width)
     }
 
     /// Read the rest of the image and chunks and finish up, including text chunks or others
--- a/benches/decoder.rs
+++ b/benches/decoder.rs
@@ -42,11 +42,11 @@ fn load_all(c: &mut Criterion) {
     bench_noncompressed_png(&mut g, 12288, 0x7fffffff); // 576 MB
     g.finish();
 
-    // Incremental decoding via `next_row`
+    // Incremental decoding via `read_row`
     let mut g = c.benchmark_group("row-by-row");
     let mut data = Vec::new();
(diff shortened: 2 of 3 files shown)


anforowicz:
This PR replaces https://github.com/image-rs/image-png/pull/421.  (This PR can be seen as a non-breaking-change alternative of the other, earlier PR.)

kornelski (approved):

src/decoder/mod.rs
```diff
@@ -536,20 +536,16 @@ impl<R: Read> Reader<R> {
...
-        let width = self.info().width;
         if self.info().interlaced {
-            while let Some(InterlacedRow {
-                data: row,
-                interlace,
-                ..
-            }) = self.next_interlaced_row()?
-            {
+            let width = self.info().width;
+            let mut row = vec![0; self.output_line_size(width)];
```
HeroicKatora, line 541: Is this additional allocation benchmarked anywhere? And from the looks below, maybe it could remain unchanged with the code in `next_interlaced_row` being preserved?

HeroicKatora (commented):
I like this approach much better as well.

src/decoder/mod.rs
anforowicz: Before this commit the following allocations happen when handling interlaced images:

1. `self.scratch_buffer` is taken and [resized](https://github.com/image-rs/image-png/blob/4d5a4b1f2667e69a4f0c2db843098599d3e88e6d/src/decoder/mod.rs#L612) via [`output_buffer.resize(output_line_size, ...)`] once per call to `next_row` / `next_interlaced_row`.  For interlaced images `output_line_size` can grow with each interlacing pass and so potentially there are 7 (re)allocations per image frame.
2. There is no `row` vector allocated inside `fn next_frame`

After this commit:
1. `self.scratch_buffer` is never allocated when going via `next_frame` (`Vec::new()` [here](https://github.com/image-rs/image-png/blob/4d5a4b1f2667e69a4f0c2db843098599d3e88e6d/src/decoder/mod.rs#L233) doesn't allocate)
2. `next_frame` will allocate the `row` vector, as you've highlighted above.

So, when decoding interlaced images:
- Before this commit: 1 call to `next_frame` can result in 1-7 allocations of row storage
- After this commit: 1 call to `next_frame` always results in 1 allocation of row storage

-------------------------

I note that `next_frame` only goes through `read_row` (or `next_interlaced_row` before this commit) for _interlaced_ images.  Nevertheless `next_row` / `read_row` are useful not just for interlaced images, but for any user/caller who wants 1) incremental decoding (e.g. wants decoded rows as soon as they are available in the input) and/or 2) wants to post-process rows (e.g. RGB => RGBA, or maybe gamma correction) while the row is still hot in L1 cache.  So the performance of `read_row` matters not only for interlaced images, but for all images.  (Incremental decoding is tricky and may need additional PRs to expose the currently-internal-only     `UnexpectedEof` and `UnexpectedEndOfChunk` errors.)

-------------------------

> Is this additional allocation benchmarked anywhere?

No, I haven't specifically benchmarked the impact of this allocation (other than executing end-to-end decoding benchmarks).

> And from the looks below, maybe it could remain unchanged with the code in next_interlaced_row being preserved?

I am not sure if I understand your proposal.  We can't decode the interlaced row into the caller/user-provided `buf`, because the decoded interlaced row may be "squished" (in the first 6 out of 7 adam7 passes) and needs to be "splat" into the final image `buf` via `adam7::expand_pass`.  In theory during pass 7 we could indeed decode directly into the user-provided buffer, but this seems complicated (and interlaced images are not that common anyway - only around 1% of images on top-500 websites - see my year-old corpus experiments [here](https://docs.google.com/spreadsheets/d/1CEgr4ttd2agOZi-MFOjqJGqfDVFUKDp-zKbrF-8Xwyk/edit?gid=1170535878#gid=1170535878)).

anforowicz: After thinking a little bit more about this feedback I now see that the changes in `fn next_frame` are somewhat separate and independent from 1) introducing `read_row` and 2) using `read_row` 2a) from `next_interlaced_row` and 2b) from the new benchmark.  And I expect this part of the changes to be mostly performance-neutral.  So let me revert this part of the changes.

Reverting this part of the changes will also let me avoid merge conflicts as I try to experiment with figuring out the best shape of a public API for exposing `adam7::expand_pass`.  (I've initially planned to implement `Reader.expand_interlaced_row(dst, row, interlace_info, option_to_override_bytes_per_pixel)` but realized that hard-coding using image width won't work if some APNG frames use smaller width.  Now I lean toward exposing `InterlaceInfo.dest_row_number() -> u32-or-usize` and `InterlaceInfo.expand(expanded_row, interlaced_row, bytes_per_pixel, maybe_no_need_to_pass_width_herre)`.  I think that `dest_row_number` will be helpful for finding the target row bytes regardless of whether the client works with image-sized buffers or frame-sized buffers.)

anforowicz: >  [...] let me revert this part of the changes.

Done.

HeroicKatora: Thanks for experimenting with different combinations, feel free to choose the best method based on those numbers. I don't feel very strongly about the interlaced case. One allocation per image isn't so much of a problem (if the numbers don't show it). Just one correct: `resize` does _at most_ one allocation, not exactly one. This should be notable for APNG where later frames can re-use the same buffer for in subsequence `fdAT` chunks and re-use the same interlacing.

Shnatsel:
There doesn't seem to be any reason not to merge this. @anforowicz could you resolve the conflicts? I'll push the merge button as soon as that's done.

anforowicz:
> @anforowicz could you resolve the conflicts?

Ack.  I'll do that later today.  My apologies for missing this message earlier. 

> There doesn't seem to be any reason not to merge this. 

Let me share some of the lingering doubts I have regarding this PR:

* Cons of this PR are small, but non-zero:
    - New public API (and public APIs are "forever";  or at least changing the APIs requires the friction of a new major version release and/or imposing some pain on crate users)
    - A bit of an unclear story wrt keeping 1) keeping [`next_row`](https://docs.rs/png/latest/png/struct.Reader.html#method.next_row) / [`next_interlaced_row`](https://docs.rs/png/latest/png/struct.Reader.html#method.next_interlaced_row) *and also* `read_row` from this PR, vs 2) deprecating `next_row` / `next_interlaced_row` and only exposing `read_row` in some future major version release
* Pros of this PR are unclear:
    - Helps to avoid an extra copy to widen the scanline in [`SkPngRustCodec::expandDecodedInterlacedRow`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/experimental/rust_png/impl/SkPngRustCodec.cpp;l=491-501;drc=d8ad496cdb21f54da1a354e2ecd59bc5911a2607), but interlaced images are relatively rare + `SkPngRustCodec` has other inefficiencies around processing them.
    - This PR definitely helps if the output of `png` doesn't need any further transformations before being usable by the crate client.  But... I now think that maybe this is **not** the case for Skia/Chromium, which requires alpha-premultiplied output.  Depending on `png` output:
        - `Indexed`: can't save the extra copy / palette expansion hop (this hop is done at the Skia/Chromium layer - see https://crbug.com/356882657)
        - `Rgb`, `Gray` or `GrayAlpha`: can't save the extra copy (need to transform into RGBA)
        - `Rgba` (~54% of images from top 500 websites): 
            - Even with this PR there would still be a need for an extra transformation loop for alpha-premul (and sometimes for rgba=>bgra, or for gamma/iccp).  (I have my brief notes about post-decoding transforms in [a doc here](https://docs.google.com/document/d/1nvp8uvHA99vHv2CNiFSpZOJxML15PYSpooG7hbKugg4/edit?usp=sharing).)
            - Maybe this PR can save a `memcpy` and then apply alpha-premul and/or rgba=>bgra transformation in-place.  But this is tricky, because [`SkSwizzler`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/codec/SkSwizzler.h;l=72;drc=534633fb4bd959681be73c45b0c7a0e56ff73c96) doesn't support in-place transformations, so we'd have to switch to [`skcms_Transform`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/modules/skcms/src/skcms_public.h;l=375-385;drc=36fe81ef6a5e07d43ce02dbffa3835ed44d4f21b) in some cases.

kornelski:
I think it's okay to merge this. Having `read_row` doesn't hurt, because even if we wanted to remove it, we can always just make a copy into the buffer.

PR #494, opened 2024-08-24
Update miniz_oxide to 0.8.0.

khuey:
Self-explanatory.


torokati44:
I'd appreciate a release with this change, as it would deduplicate `miniz_oxide` in our builds:
https://github.com/ruffle-rs/ruffle/actions/runs/11012252965/job/30578066740?pr=18061#step:5:620
Thanks in advance! ^^

PR #496, opened 2024-09-04
Support for resuming decoding after `UnexpectedEof`.

anforowicz:
This commit supports resuming decoding after `UnexpectedEof` in two ways:

1. Support for detecting an unexpected EOF using the public API of `DecodingError`.  Before this commit `UnexpectedEof`, `UnexpectedEndOfChunk`, and `NoMoreImageData` errors were represented as a crate-internal `FormatErrorInner` type.  After this commit, these errors have a representation that can be detected using the public API: `DecodingError::IoError(std::io::ErrorKind::UnexpectedEof.into())`.

2. Support for resuming decoding after an unexpected EOF.  Before this commit `fn next_interlaced_row` would unconditionally call `next_pass` - advancing to the next row.  After this commit this will only happen if the previous row has been successfully decoded.

### DIFF
--- a/fuzz/fuzz_targets/buf_independent.rs
+++ b/fuzz/fuzz_targets/buf_independent.rs
@@ -3,37 +3,24 @@
 use libfuzzer_sys::fuzz_target;
 
 use std::mem::discriminant;
-use std::io::{BufRead, Read, Result};
+use std::io::{Read, Result};
 
 /// A reader that reads at most `n` bytes.
-struct SmalBuf<R: BufRead> {
+struct SmalBuf<R: Read> {
     inner: R,
     cap: usize,
 }
 
-impl<R: BufRead> SmalBuf<R> {
+impl<R: Read> SmalBuf<R> {
     fn new(inner: R, cap: usize) -> Self {
         SmalBuf { inner, cap }
     }
 }
 
-impl<R: BufRead> Read for SmalBuf<R> {
+impl<R: Read> Read for SmalBuf<R> {
     fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
         let len = buf.len().min(self.cap);
         self.inner.read(&mut buf[..len])
-    }
-}
-
-impl<R: BufRead> BufRead for SmalBuf<R> {
-    fn fill_buf(&mut self) -> Result<&[u8]> {
-        let buf = self.inner.fill_buf()?;
-        let len = buf.len().min(self.cap);
-        Ok(&buf[..len])
-    }
-
-    fn consume(&mut self, amt: usize) {
-        assert!(amt <= self.cap);
-        self.inner.consume(amt)
     }
 }
 
@@ -48,7 +35,7 @@ fuzz_target!(|data: &[u8]| {
 });
 
 #[inline(always)]
-fn png_compare<R: BufRead, S: BufRead>(reference: png::Decoder<R>, smal: png::Decoder<S>)
+fn png_compare<R: Read, S: Read>(reference: png::Decoder<R>, smal: png::Decoder<S>)
     -> std::result::Result<(), ()>
 {
     let mut smal = Some(smal);
--- a/fuzz/fuzz_targets/buf_independent.rs
+++ b/fuzz/fuzz_targets/buf_independent.rs
@@ -2,7 +2,6 @@
 
 use libfuzzer_sys::fuzz_target;
 
-use std::mem::discriminant;
 use std::io::{Read, Result};
 
 /// A reader that reads at most `n` bytes.
@@ -59,8 +58,13 @@ fn png_compare<R: Read, S: Read>(reference: png::Decoder<R>, smal: png::Decoder<
         match (rref, rsmal) {
             (Ok(info), Ok(sinfo)) if ref_data == smal_data => assert_eq!(info, sinfo),
             (Ok(_), Ok(_)) => panic!("Deviating data decoded"),
-            (Err(er), Err(es)) if discriminant(&er) == discriminant(&es) => break Ok(()),
-            (Err(ferr), Err(serr)) => panic!("Deviating errors {:?} vs {:?}", ferr, serr),
+            // All errors are treated as equivalent - it is okay to get different errors depending
+            // on how the input is chunked.  One example is deflating [88, 9, 99, 30, 5, 99] in one
+            // chunk will fill up `ZlibStream.out_buffer` and hold onto the final byte in
+            // `fdeflate`'s state.  But when deflating byte-by-byte, `out_buffer` will be resized
+            // before handling the final byte, leading to `InvalidLiteralLengthCode` being returned
+            // by `fdeflate`.
+            (Err(_), Err(_)) => break Ok(()),
             (Ok(_), Err(err)) => panic!("Small buffer failed {:?}", err),
             (Err(err), Ok(_)) => panic!("Unexpected success: {:?}", err),
         }
--- a/fuzz/fuzz_targets/buf_independent.rs
+++ b/fuzz/fuzz_targets/buf_independent.rs
@@ -1,8 +1,32 @@
+//! This fuzzer tests that decoding results are the same regardless of the
+//! details of how the `Read` trait exposes the underlying input via
+//! the `fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize>`
+//! method:
+//!
+//! * Whole slice - `impl Read for &[u8]`:
+//!     - The whole slice is available for reading (if it fits into `buf`)
+//!     - No IO errors are expected
+//!     - Motivation: This is the baseline
+//! * Byte-by-byte - `SmalBuf<R>`:
+//!     - At most 1 byte can be read in a single call to `read`
+//!     - Motivation: Testing that decoding works regardless of how the input is split
+//!       into multiple `read` calls.  (The test checks every possible `read` boundary in the input
+//!       buffer, even though in practice file or network buffers would split the input into only a
+//!       handful of chunks.)
+//! * TODO: Intermittent EOFs:
+//!     - Intermittently `read` report 0 available bytes.
+//!     - Still no IO errors at the `Read` trait level
+//!     - Motivation: Testing support for decoding a streaming or partial input
+//!       (i.e. scenarios where initially only the first few interlaced passes
+//!       can be decoded, and where decoding is resumed after getting more complete
+//!       input).
+
 #![no_main]
 
 use libfuzzer_sys::fuzz_target;
 
-use std::io::{Read, Result};
+use std::fmt::Debug;
+use std::io::Read;
 
 /// A reader that reads at most `n` bytes.
 struct SmalBuf<R: Read> {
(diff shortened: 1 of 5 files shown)


anforowicz:
This PR is still work-in-progress:
* I need to investigate and address CI failures (error mismatch when fuzzing + cargo crate resolution failure in Rust 1.57.0)
* This hasn't yet been fully polished / self-reviewed

Let me nevertheless share some early notes and thoughts...

(Please shout if you'd prefer me to resolve issues in such an "uncooked" PRs internally, without necessarily bothering you.  It's a bit of a balancing act to be transparent, but also respecting and avoiding wasting your time and energy.)

---------------------

To represent an unexpected-eof condition I decided to use `DecodingError::IoError(std::io::ErrorKind::UnexpectedEof.into()`.  I've considered the following alternatives:

* Adding `DecodingError::UnexpectedEof` variant.  This seems undesirable because AFAIU this would be [considered](https://rust-lang.github.io/rfcs/1105-api-evolution.html#major-change-adding-new-variants) a breaking change.
* Somehow exposing `FormatErrorInner::UnexpectedEof` (and `FormatErrorInner::UnexpectedEndOfChunk` and `FormatErrorInner::MissingImageData`) through the public API.  One idea would be a new `FormatError::is_unexpected_eof(&self) -> bool` method.  This API seems less elegant that reusing `IoError` variant.

FWIW it seems desirable to me that `DecodingError::IoError(std::io::ErrorKind::UnexpectedEof.into()` is used for unexpected EOFs that originate from either 1) `png` crate decoder recognizing a truncated input and/or 2) the input stream (e.g. `std::io`) recognizing an end-of-file.  (I am not 100% sure about the latter - [`Read::read`](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) documents that it returns `Ok(0)` to indicate EOF, but I imagine that an `io::Error` may nevertheless be used in some scenarios.)

---------------------

I am still trying to figure out if I should add a doc comment with an example somewhere.  This PR is not introducing a _new_ API, so there isn't an obvious place for this.  And it seems hard to construct a _succint_ example.

--------------------

This PR focuses on resuming after `Reader`'s methods report an unexpected EOF.

After this PR there is still no way to resume after [`Decoder::read_info`](https://docs.rs/png/latest/png/struct.Decoder.html#method.read_info), because this method _consumes_ `self` (so the caller cannot retry calling `read_info`).  This seems ok:

* The PR still adds useful functionality (resuming in some scenarios)
* In my particular use case, it's okay to fail when the input doesn't yet contain metadata (e.g. if [`SkPngRustDecoder::Decode`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/experimental/rust_png/SkPngRustDecoder.h;l=19;drc=cbd5898acbd8b51f11c344b5278a435e3e182918) fails with [`kIncompleteInput`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/include/codec/SkCodec.h;l=81-84;drc=ade36690942cc1126513eb4ee58b98f4352a4b55), then the rest of Skia expects the caller to [rewind the stream](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/include/core/SkStream.h;l=96-100;drc=3844fc48802d20886e628b1d3111f7a31d3fba72) and retry from scratch).

PS. Now that I think about it, I probably should add test coverage for resuming after unexpected error in [`Decoder::reader_header_info`](https://docs.rs/png/latest/png/struct.Decoder.html#method.read_header_info).

----------------------------

I am not sure how to address the crate resolution failure that the CI found in Rust 1.57.0.  I guess one way would be to avoid adding a dependency on the `tempfile` crate and use a different streaming/partial input mechanism, but I am not sure what mechanism I could/should use.

I guess I could have a `impl Read` for a custom, test-only struct that wraps a `Vec<u8>` + `available_len: usize` and uses `RefCell` to share this between a `Decoder` and a test (the latter to control when `available_len` grows).  This seems a little baroque and complex when temp files work...

anforowicz:
Let me record some notes about the fuzzing failure - decoding 6 bytes of zlib stream in one chunk may produce a different result than decoding byte-by-byte (and this is okay and therefore the fuzzer shouldn't insist on having the same error in both cases).  The difference happens, because `fdeflate` is given a bigger `out_buffer` in the byte-by-byte case and this lets `fdeflate` proceed further and report the `InvalidLiteralLengthCode` error.  In the single-chunk case, `fdeflate` exits early because the `out_buffer` is already full - this means that at the `png` layer we will realize that we have an `UnexpectedEof` scenario **before** giving `fdeflate` a chance to realize that there is a problem with the 6th byte of the zlib input:

```
$ cargo fuzz run --dev buf_independent fuzz/artifacts/buf_independent/crash-adb8400b4ae976c554178b4848c504dacbf49255
...
Running: fuzz/artifacts/buf_independent/crash-adb8400b4ae976c554178b4848c504dacbf49255
reference.next_frame...
decompress: data.len=6, data = [88, 9, 99, 30, 5, 99, ],  out_buf.len=116, out_pos=0
decompress: in_consumed=6, out_consumed=116
UnexpectedEof: finish_decoding
reference.next_frame... done.
smal.next_frame...
decompress: data.len=1, data = [88, ],  out_buf.len=116, out_pos=0
decompress: in_consumed=1, out_consumed=0
decompress: data.len=1, data = [9, ],  out_buf.len=116, out_pos=0
decompress: in_consumed=1, out_consumed=0
decompress: data.len=1, data = [99, ],  out_buf.len=116, out_pos=0
decompress: in_consumed=1, out_consumed=0
decompress: data.len=1, data = [30, ],  out_buf.len=116, out_pos=0
decompress: in_consumed=1, out_consumed=1
decompress: data.len=1, data = [5, ],  out_buf.len=116, out_pos=1
decompress: in_consumed=1, out_consumed=115
decompress: data.len=1, data = [99, ],  out_buf.len=32884, out_pos=116
decompress: CorruptFlateStream
smal.next_frame... done.
thread '<unnamed>' panicked at fuzz_targets/buf_independent.rs:67:39:
Deviating errors IoError(Kind(UnexpectedEof)) vs Format(FormatError { inner: CorruptFlateStream { err: InvalidLiteralLengthCode } })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

anforowicz:
I think this is now ready to be reviewed - can you PTAL?

Motivation for this PR is to support using the `png` crate from `SkCodec` API in Skia/Chromium.  In particular, this PR supports [a Skia CL here](http://review.skia.org/896876) which adds support for:
* Returning `SkCodec::kIncompleteInput` from `SkCodec::incrementalDecode` when the `png` crate encounters an unexpected EOF
* Passing Blink tests like [`AnimatedPNGTests.ByteByByteSizeAvailable`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/png/png_image_decoder_test.cc;l=340-349;drc=bde3d0d4b13aee737848caba9b62c37adbbe5806)

kornelski:
How this is used in products using Skia or Blink?

Are they using it as a way to support async I/O? Are they trying to fix broken files?

anforowicz:
> How this is used in products using Skia or Blink?
> 
> Are they using it as a way to support async I/O? Are they trying to fix broken files?

Before answering the question, I want to highlight that the proposed changes are independent from Skia and Blink - Rust standard library itself allows resuming after `Read::read` (see [the docs here](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#tymethod.read:~:text=for%20File%2C%20it%20is%20possible%20to%20reach%20the%20end%20of%20file%20and%20get%20zero%20as%20result%2C%20but%20if%20more%20data%20is%20appended%20to%20the%20file%2C%20future%20calls%20to%20read%20will%20return%20more%20data)):

> [...] for `File`, it is possible to reach the end of file and get zero as result, but if more data is appended to the file, future calls to read will return more data

In fact in an earlier version of this PR, I've used `std::io::File` in the `test_streaming_input` test - see [here](https://github.com/image-rs/image-png/commit/adc6dc71b9aa5147110fb4e3050d558bdbd378ba#diff-ef73ce77825d3cecdd3c7c8f44f94c4be566bd325c5d0294603a2d65cd211f9fR1992-R2016).  (I backed out of using `std::io::File` to avoid taking a dependency on the `tempfile` crate which led to some CI failures with crate resolution in Rust 1.57.0.)

-------------------------

In case of Blink the answer looks more or less like this:

* The input buffer is represented as a ref-counted [`SegmentReader`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/segment_reader.h?q=blink::SegmentReader&ss=chromium).
* The `SegmentReader` stores bytes of a HTTP response in memory - you can think of it as a read-only wrapper around `std::vector<uint8_t>`, although the real impl may "chunk" the buffer into smaller segments (e.g. see [`SegmentedBuffer::Segment`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/wtf/shared_buffer.h;l=56-72;drc=8e948282d37c0e119e3102236878d6f4d5052c16)).
* As the Network Service process receives data over HTTP, it will stream the data into the Renderer process, where the data will be eventually appended to the buffer underneath the `SegmentReader`
* A Blink image decoder gets the `SegmentReader` in [`ImageDecoder::OnSetData`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/image_decoder.h;l=283;drc=39736f289b63165a3427e7b09c83c65e4ed7f323).
* Blink/Skia integration code understands that `kIncompleteInput` doesn't indicate a failure unless the HTTP network stream has actually been closed/drained - e.g. see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/skia/skia_image_decoder_base.cc;l=81-85;drc=b74f39b352a0fa4e36354239c22653b063356218).

Some more details that are probably not that important:

* Quick clarification about `ThreadSafeRefCounted<SegmentReader>`: ref-counting is thread-safe (so the pointer may be safely passed across threads), but the object itself is not (in practice all accesses are done from a single thread and this property guarantees thread safety).
* Skia/Blink integration code wraps the `SegmentReader` in [`SegmentStream`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/segment_stream.h) which exposes Skia's [`SkStream`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/include/core/SkStream.h?q=SkStream&ss=chromium) API - see [this snippet in `SkiaImageDecoderBase::OnSetData`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/skia/skia_image_decoder_base.cc;l=62-67;drc=b74f39b352a0fa4e36354239c22653b063356218).
* `SkiaImageDecoderBase` retains a raw pointer to the `SegmentStream` (this is lifetime-safe, because `SkiaImageDecoderBase` owns `codec_` which owns the `SegmentStream`).
* When `SkiaImageDecoderBase::OnSetData` is called again, then it [updates](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/skia/skia_image_decoder_base.cc;l=58;drc=b74f39b352a0fa4e36354239c22653b063356218) what input data is available to the codec

anforowicz:
> Are they using it as a way to support async I/O? Are they trying to fix broken files?

Ooops... I noticed that I didn't directly answer some aspects of the question...

> Are they trying to fix broken files?

No.

> Are they using it as a way to support async I/O? 

Kind of:

* Blink is effectively single-threaded (per HTML frame at least).  (Which I guess would argue for answering "no, there is no async".)
* But in practice 1) receiving a HTTP response, and 2) rendering (and decoding images) happens 3) concurrently:
    * The Network Service process handles HTTP responses and uses IPC to send bytes to a Renderer Process
    * The Renderer Process runs a task/message handling loop in its `main`.  Some tasks say "render the web page" and some tasks say "handle these new HTTP response bytes".

anforowicz:
I also note that PNG is inherently designed to support partially-received / streaming inputs - AFAIU this is the main reason why interlacing exists.

fintelia:
To understand, after this PR decoding can resume following an `UnexpectedEof` error _if_ all the chunks before the first IDAT have already been read _and_ you only read rows one at a time via `next_interlaced_row` / `next_row`?

I think it is worthwhile to guarantee decoding can resume after EOFs, given the demonstrated interest in the feature. Though it is worth noting that it'll likely add some maintenance complexity going forward. As you note, the `Read::read` method does have well defined behavior following an I/O error. However, the `Read::read_exact` method doesn't:
> If this function returns an error, it is unspecified how many bytes it has read, but it will never read more than would be necessary to completely fill the buffer.

This library doesn't current read from the input via `read_exact` (it directly implements a state-machine to read bytes one-by-one) but I've found that it can make for simpler decoders in some of the other image-rs codecs.

anforowicz:
> To understand, after this PR decoding can resume following an `UnexpectedEof` error _if_ all the chunks before the first IDAT have already been read

Yes.

>  _and_ you only read rows one at a time via `next_interlaced_row` / `next_row`?

I plan to use `next_interlaced_row` / `next_row` and this is what is tested in the new `test_stream_input_and_decode_single_next_frame` test added in this PR.

But, resuming should also work (I think) after `next_frame`.  I probably should add a test to cover this.

> I think it is worthwhile to guarantee decoding can resume after EOFs, given the demonstrated interest in the feature. Though it is worth noting that it'll likely add some maintenance complexity going forward. As you note, the `Read::read` method does have well defined behavior following an I/O error. However, the `Read::read_exact` method doesn't:

Good catch - I didn't think about this aspect.

I note that `read_exact` would _almost_ work - I _think_ it would: 1) return `std::io::ErrorKind::UnexpectedEof` but 2) the partially read bytes would be discarded and lost forever.  And usage of `Read::read` also has the risk of loosing the partially-read data - it works in the `png` crate because this crate always saves the full state after each `read` (mostly by the virtue of going through `BufRead::fill_buf`).  In other words - `reader.read(&mut local_buffer_that_wont_be_preserved_in_long_lived_state` is also risky. 

I hope that we can nevertheless proceed with the current shape of the PR - being able to resume decoding is an important requirement for my usage.  In theory I could preserve compatibility with the old Chromium/Skia implementations if we just take half of this PR (the public-detectability-via-`std::io::ErrorKind::UnexpectedEof` part) if after EOF I would always resume from scratch.  But the current implementation doesn't resume from scratch, so I worry that doing this would be seen as a regression ("just" a performance regression though + we don't know how often this happens).

Do you think proceeding with the current shape of the PR is ok?  I am not sure if we have a good alternative, but let me try to share some (brinastorming-quality) ideas:

* Maybe we can wrap the client/user-provided `Read` in a `NoReadExactReader` wrapper that is `#[repr(transparent)]` and forwards all the method calls to the wrapper `read` implementation, except for ones that may be incompatible with resuming (e.g. panics if `read_exact` is ever called).  This may help a bit, but as I point out, `read` into local/short-lived/non-preserved state would still be a risk.
* Maybe we can add more test coverage - beyond the handful of unit tests.  We could either extend the fuzzing (today it compares with byte-by-byte reader, but we could add a byte-by-byte-and-temporary-eof-after-each-byte mode).  Or we could extend functional-comparison tests.

> > If this function returns an error, it is unspecified how many bytes it has read, but it will never read more than would be necessary to completely fill the buffer.
> 
> This library doesn't current read from the input via `read_exact` (it directly implements a state-machine to read bytes one-by-one) but I've found that it can make for simpler decoders in some of the other image-rs codecs.

Ack.

anforowicz:
I think that for now this PR is blocked until I take care of:

* [ ] Adding a unit test for detecting-temporary-eof-and-resuming-via-`next_frame`
* [ ] Adding temporary-eof coverage to fuzzing tests

kornelski:
Maybe keep the `BufRead` interface, and implement as much as possible on slices? 

`BufRead` reduces copying when reading from a `Vec`. Slices have much cheaper error/eof handling, and won't be exposed to `read_exact` gotcha.

fintelia:
@anforowicz I'm fine moving forward with this approach. Adapting `next_frame` might be worth doing in a separate PR since right now the logic doesn't record how many rows have already been processed.

@kornelski The `BufRead` changes in this PR seem to just be test/fuzzing related.

More broadly, trying to just use `BufRead` actually does run into a variation of the same issue: The `fill_buf` method might return only one byte and you have to consume it before you can get anymore bytes. And the refill operation after consuming that single byte can fail... 

I do think this is solvable. The byte-at-a-time state machine used now does work. Another idea would be to combine `read_exact` and `seek`. The reader would keep track of the current position before reading each chunk, and trying to read again after an EOF would start by seeking back to the start of the chunk. 

(The [`PngDecoder`](https://docs.rs/image/latest/image/codecs/png/struct.PngDecoder.html) struct in the main _image_ crate already has both `BufRead` and `Seek` bounds to enable possible changes here.)

kornelski:
Yes, even with BufRead you need to handle 1 byte at a time. It's just cheaper to get 1 byte from a slice than io::Read. When you issue non-exact read you may get 1 byte too.

anforowicz:
Just a quick status update.  TL; DR: I still need to work on adding fuzzing coverage, but I've updated this PR to cover `UnexpectedEof`-resumability via `next_frame`.

-----------------

> I'm fine moving forward with this approach. Adapting `next_frame` might be worth doing in a separate PR since right now the logic doesn't record how many rows have already been processed.

Ack.  It's fine either way (and I can split/trim this PR if this is what you'd like).  FWIW I am leaning toward adding `next_frame` support to this PR:

* Con: the PR grows bigger
* Pro: I feel that the PR is more robust this way
    * API-wise: `png` users probably would expect resumability across all public APIs of `Reader` and `Decoder` (where they retain `self` - so including `next_frame` but excluding `Decoder.read_info`)
    * Content-wise: e.g. in the way how it tweaks `mod.rs` to avoid prematurely advancing state + to support resuming at the same row/offset
* Pro: fuzzing coverage through `next_frame` covers most of the call paths (including `next_interlaced_row_impl` _and_ APNG) 

-----------------

I think that fuzzing coverage should be added before accepting and merging this PR.  I am still fairly confident that using `fill_buf` makes `png` resilient against `UnexpectedEof`, but I am starting to appreciate that there indeed are extra requirements for maintaining `png`-internal state in a way that is kind of "idempotent" and doesn't prematurely advance state.  I think that fuzzing coverage should give us the confidence we need.

This PR grew a little bit, but I hope that the changes are still readable - I tried to keep the commits fairly small and independent.  Some notable changes so far:

* I am changing the `InterlaceInfo::Null` variant to add some payload.  In theory this is a breaking change, but `InterlaceInfo` itself wasn't part of the public API until https://github.com/image-rs/image-png/pull/495 (which hasn't been released yet in a `png` version on crates.io), so I think we are still ok.  I think both of the variants of `InterlaceInfo` should be fairly future-proof - both are public structs, but they are opaque (no public fields, and for now no accessor functions/methods).
* Extracting a separate `interlace_info.rs` isn't _required_ for this PR, but I think this is desirable on its own + it helps with making the changes needed in the "Support for resuming decoding after UnexpectedEof" commit.  (Otherwise controlling when to advance the state would be more inter-twined/coupled across `Reader` and `InterlaceInfoIter`.)

-----------------

FWIW I plan to do the fuzzing changes in 2-3 commits (either tomorrow/Tuesday, or otherwise it will have to wait until ~Thursday):

* First commit: support comparison across a _vector_ of readers (not just across a pair of regular+small reader).
* Second commit: add a reader that every other `read` returns 0 bytes (and wraps a small, one-byte-at-a-time reader when used in practice)
* Third commit: tweak fuzzing so that it tries to decode additional frames to cover APNG-related transitions

anforowicz:
Status update - I uploaded the first draft of commits that add intermittent-EOFs coverage to the fuzzer.  OTOH, this is blocked by pre-existing fuzzing issues that I've reported in https://github.com/image-rs/image-png/issues/497.

anforowicz:
@fintelia and/or @kornelski - can you please consider merging the PR in the current shape?

I've previously [said](https://github.com/image-rs/image-png/pull/496#issuecomment-2338937503:~:text=I%20think%20that%20fuzzing%20coverage%20should%20be%20added%20before%20accepting%20and%20merging%20this%20PR.) that we should add fuzzing coverage before merging this PR.  I still think that ideally we should have passing fuzzing coverage, but merging as-is has some benefits:

Arguments for merging without cleaning up the fuzzing-based tests:

* The fuzzing issues existed before this PR - this PR doesn't make things worse from that perspective.  And the timeline for fixing these issues is a bit uncertain at this point (it's possible that the initial test case from https://github.com/image-rs/image-png/issues/497 is just the tip of the iceberg).
* Merging will remove the breaking-API-change hazard that I have mentioned in https://github.com/image-rs/image-png/pull/495#issuecomment-2338872525
* Merging will help to "hold the line" for the new requirements.  (i.e. keeping the new tests passing / preventing PRs that would have resulted in non-resumable behavior)
* Merging now will prevent potential future merge conflicts. (This PR has grown quite big).
* Merging will help me land Chromium-side and Skia-side CLs for my project.  (Well, this only benefits me.  This argument has no impact on maintenance of the `png` nor `fdeflate` crates .)

Arguments against merging without cleaning up the fuzzing-based tests:

* This possibly digs a deeper hole for the fuzzing problem (since if we merge then we not only are on the hook for fixing byte-by-byte-VS-whole differences but also intermittent-EOF differences).
* I can commit to focusing on fixing the fuzzing problem for at least 50% of my available work time for the next month or so.  I think it should be enough to clean up the problem (even though I may need to take some time off in the next week or so).  But... if I get hit by the proverbial bus, then this problem will fall on your/maintainer's shoulders... :-/
* This commits to the new requirements (i.e. to being able to resume/retry `next_frame` etc. after `UnexpectedEof`).  This may be a bit premature - we don't know for 100% that this new requirement is met today.

WDYT?

kornelski (approved):

PR #498, opened 2024-09-19
Fix resuming `next_frame` after `UnexpectedEof` (APNG trouble).

anforowicz:
Fuzzing found an example of an animated PNG that would trigger an assert in `fn read_until_image_data`, indicating that some `image_data` is unexpectedly appended into a throw-away `buf`.  This example is being added to `fuzz/corpus/buf_independent/regressions/...`.  This example was originally reported at
https://oss-fuzz.com/testcase-detail/6611224761008128

In the example, a sequence of `IDAT` chunks was interleaved with some other (ignored, unrecognized) chunks.  During normal, non-interrupted decoding `fn next_frame` would advance to the invalid next sequence of consecutive chunks and decode them.  OTOH when resuming decoding after `UnexpectedEof` `fn next_frame` would see that it is still not at the expected frame, and `fn read_until_image_data` would end up decoding some image data and failing `assert!(buf.is_empty())`.

The problem described above is fixed by adding checks that an `IDAT` sequence may only start once per PNG file + that an `fdAT` sequence may only start if we've seen an `fcTL` chunk after the previous `fdAT` sequence has terminated.


src/decoder/stream.rs
```diff
@@ -770,22 +793,25 @@ impl StreamingDecoder {
                     self.current_chunk.type_ = type_str;
                     self.inflater.finish_compressed_chunks(image_data)?;
                     self.inflater.reset();
+                    self.ready_for_idat_chunks = false;
```
kornelski, line 796: Why is it set to `false` after `IDAT` too? There could be multiple consecutive `IDAT`s.

anforowicz: Here we do not have **consecutive** `IDAT`s:

* At a high-level this piece of code handles transitioning from 1-or-more-IDAT-chunks to some-other-chunk.  For example - when that happens we flush the remaining decompressed data out of `self.inflater`).
* On old lines 767-768 (new lines 790-791) above the new `type_str` is not an `IDAT` anymore, because it is different from the previous chunk's `self.current_chunk.type_`.

PR #499, opened 2024-09-19
Avoid infinite loops in `buf_independent` fuzzer after real EOF.

anforowicz:
A truncated PNG file can legitimately result in an `UnexpectedEof` error being reported.  Before this commit, `fn retry_after_eofs` in the `buf_independent.rs` fuzzer would repeatedly try to resume such unresumable input.

Fuzzing found an example that illustrates the problem described above. This example is being added to
fuzz/corpus/buf_independent/regressions/.... This example was originally reported at https://oss-fuzz.com/testcase-detail/5790077345660928


fuzz/fuzz_targets/buf_independent.rs
```diff
@@ -75,49 +75,65 @@ mod intermittent_eofs {
...
-            if self.enabler.is_enabled() && self.eof_soon {
+            if self.controller.are_intermittent_eofs_enabled() && self.eof_soon {
                 self.eof_soon = false;
                 return Ok(0);
             }
 
             self.eof_soon = true;
-            self.inner.read(buf)
+            let inner_result = self.inner.read(buf);
+            match inner_result.as_ref() {
```
kornelski, line 97: Just FYI this can be codegolfed:

```rust
    if res.as_ref().is_ok_and(|&b| b == 0) {
    }

    if res.as_ref().ok() == Some(&0) {
    }

    if let Ok(0) = &res {
    }
```

fuzz/fuzz_targets/buf_independent.rs
```diff
@@ -75,49 +75,65 @@ mod intermittent_eofs {
...
+            }
+
+            inner_result
         }
     }
 
-    pub struct Enabler {
-        is_enabled: RefCell<bool>,
+    pub struct EofController {
+        are_intermittent_eofs_enabled: RefCell<bool>,
```
kornelski, line 107: For copyable types `Cell<bool>` works as well, and has zero overhead

fuzz/fuzz_targets/buf_independent.rs
anforowicz: Nice!  Thanks for the suggestion.  Done.

fuzz/fuzz_targets/buf_independent.rs
anforowicz: Thanks for the feedback - I didn't realize this was an option.  I like the resulting simplification.  Done.

HeroicKatora (approved):

PR #500, opened 2024-09-20
Restore using `NoMoreImageData` errors.

anforowicz:
https://github.com/image-rs/image-png/pull/496 has incorrectly started to return `UnexpectedEof` in a situation when when there is no `Read`-level EOF.  In particular, this may happen when an `IDAT` of `fdAT` chunk has been fully decoded, decompressed, and flushed, but still doesn't contain the expected number of image pixels.  In such a situation the old `NoMoreImageData` error was appropriate.  In a sense, this commit is a partial revert and/or refactoring of the earlier commit https://github.com/image-rs/image-png/pull/496/commits/b5b0674a0cd50a1f3c1e09f699c04b168648a237

This commit fixes a timeout found locally by the `buf_independent` fuzzer.  The repro case has been saved under
`fuzz/corpus/buf_independent/regressions`.


src/decoder/mod.rs
```diff
@@ -702,7 +702,9 @@ impl<R: Read> Reader<R> {
         // Read image data until we have at least one full row (but possibly more than one).
         while self.data_stream.len() - self.current_start < rowlen {
             if self.subframe.consumed_and_flushed {
-                return Err(DecodingError::IoError(ErrorKind::UnexpectedEof.into()));
+                return Err(DecodingError::Format(
+                    FormatErrorInner::NoMoreImageData.into(),
+                ));
```
anforowicz, line 707: The change above is needed to avoid a timeout when `cargo fuzz run buf_independent fuzz/corpus/buf_independent/regressions/minimized-from-110a360793767757e44c9e4b95c495666e39e672` is run.

src/decoder/mod.rs
```diff
@@ -716,12 +718,9 @@ impl<R: Read> Reader<R> {
 
             match self.decoder.decode_next(&mut self.data_stream)? {
                 Some(Decoded::ImageData) => {}
-                Some(Decoded::ImageDataFlushed) => {
+                Some(Decoded::ImageDataFlushed) | None /* after IEND chunk */ => {
                     self.subframe.consumed_and_flushed = true;
                 }
-                None => {
-                    return Err(DecodingError::IoError(ErrorKind::UnexpectedEof.into()));
-                }
```
anforowicz, line 724: The change above is not associated with a regression test:

* I haven't been able to hit this branch when fuzzing
* I think that `ImageDataFlushed` will always come first - this makes me question if this branch can be hit in practice.
* Maybe we could hit this branch when we get into this loop _outside_ of `IDAT` processing.  This seems tricky, given that we search for `IDAT` when creating the `Reader` and when calling `next_frame` (and `next_row` doesn't progress beyond the current frame).

Nevertheless, I think this change is still desirable:

* This change means that this commit restores **all** `NoMoreImageData` cases touched by https://github.com/image-rs/image-png/commit/b5b0674a0cd50a1f3c1e09f699c04b168648a237
* I think the simplification is desirable and okay:
    - I think that treating this as `consumed_and_flushed` makes sense (it seems desirable to check the loop condition one last time in case there is some additional `image_data` + it seems okay to rely on the other check above to return `NoMoreImageData` error)
    - I think that it is okay to not restore distinguishing between `NoMoreImageData` and `UnexpectedEndOfChunk` (the difference was not exposed through the public API [well, except through `Debug` implementation];  and in both cases / at a high-level we want more rows/pixels but can't get them - this seems to be covered well by `NoMoreImageData`)

fintelia: In general, if a specific case is impossible we should `panic!` rather than returning an error or silently continuing. That way fuzzing can detect if they actually can be hit, which usually signals a bug elsewhere in the code.

This crate used to have tons of error handling code for cases that weren't even reachable in the first place, which I've tried to clean up over time.

anforowicz: Ack.  You're right that my attempt to avoid risk here leads to suboptimal code.  Let me try to think through the options available.

* Option #1: `None => unreachable!()`.  Special-casing `None` seems wrong, because `None` is not the only unreachable option here - e.g. `Some(Decoded::Header)` is another unreachable option.
* Option #2: Removing `| None`.  This simplifies the code.  And fuzzing indeed increases our confidence that things still work fine without this branch (and without leading to a timeout / without making this an infinite loop).
* Option #3: Exhaustively list all the possible options and panic for ones we think are unreachable.  I don't like this option because:
    - This adds noise to the code, making it harder to read and understand this loop
    - This also seems a bit risky - fuzzing doesn't cover all valid ways of using the public API (it only covers `next_frame` today and doesn't cover `next_row`;  and even if we added `next_row` coverage, then there are various permutations of public API calls that we may need to fuzz).

I am leaning toward option #2.  For now, let me try option #2 locally and double-check that fuzzing doesn't complain.

PR #502, opened 2024-09-24
Advanced compression controls

### LINKED ISSUE #349: [Feature Request] Enabling uncompressed PNG support
Currently, `image` provides support for many PNG compression levels, all the way from Fast, Best, Huffman, and RLE.

However, according to the PNG specification, it is possible to store only uncompressed deflate blocks.

While this would be a niche feature addition, it would be very useful for scenarios where the highest encoding speed is required, as the encoding speed of uncompressed PNGs is obviously the highest among all other settings, and unlike PPM, we get greater compatibility with general software with it.

It shouldn't be too hard to add it to `image`'s PNG implementation, so hopefully it won't take too much time to do so.

Summary of feature request: Enabling/Implementing uncompressed PNG support for maximum encoding/decoding speed.

Example program of something that would heavily benefit from this:
https://github.com/shssoichiro/butter-video

Shnatsel:
Exposes advanced DEFLATE compression options. Supersedes #488, fixes #349.

Also adds an option for store-only compression, like #490, but without a semver break.

Expands roundtrip tests to cover more compression modes.

### Motivation

I am going to need this in [`wondermagick`](https://github.com/Shnatsel/wondermagick). I'll have to roughly match `imagemagick`'s own compression levels using controls over both filters and DEFLATE compression level.

### Why this API?

The API we have right now is a mix of high-level and low-level concepts. We only have a limited selection of DEFLATE compression levels, but also expose the full complexity of filter selection to the user without any heuristics and requiring the user deal with it.

Ideally we'd expose a high-level API with a few simple presets that take care of **both** DEFLATE and filtering, plus a separate low-level API that exposes the full complexity of DEFLATE and filtering options for users who do need it, such as `wondermagick`.

Evolving `Compression` to set the filter as well is an API break, but we can add the low-level API in a semver-compatible way, which is what this PR does.

### Semver and longevity

This API it can be easily evolved later: new modes can be easily added since the enum is marked `#[non_exhaustive]`, and removed modes can be marked `#[deprecated]` and mapped to the closest thing we do have.

The only observable change in behavior is converting from `Info` to `PartialInfo` and back will reset the `compression` field since it gets lowered to the advanced representation now, and that field is public. If that's an issue, I can make `PartialInfo` preserve the original compression setting.


Shnatsel:
PR for a nicer high-level API I mentioned in the description is now up: #503

Shnatsel:
We're doing semver-breaking changes now anyway, so I'm going to close this in favor of #503 which includes all of this and more.

PR #503, opened 2024-09-25
High-level compression options API

### LINKED ISSUE #349: [Feature Request] Enabling uncompressed PNG support

Shnatsel:
Cleans up the high-level compression ratio API and makes it easier to use.

 - Makes setting `Compression` also automatically select the filtering mode for you
 - Removes obsolete `Huffman` and `Rle` compression options
 - Makes the `Compression` enum non-exhaustive so we could add e.g. `Moderate` for [fdeflate's new mode](https://github.com/image-rs/fdeflate/pull/29) and `Extreme` for Zopfli in the future
 - Renames `Best` to `High` since it isn't actually best, as its documentation already states. Especially if we add Zopfli later.
 - Renames `Default` to `Balanced` and makes it the default

Builds upon #502, so the diff includes changes from #502 when compared with main. Diff without the changes from #502 can be found here: https://github.com/Shnatsel/image-png/compare/advanced-compression...Shnatsel:image-png:high-level-compression-control?expand=1

Supersedes #490, fixes #349 and #505

Unlike #502, this is a semver-breaking change.

The exact choice of filters for each mode is debatable, but I think this is a reasonable default. I'm happy to change the mapping from the high-level `Compression` modes based on your input - I'm sure you have more experience with these than I do.

**Update:** also implements #505 in the same PR, as requested


Shnatsel:
Looks like the expanded roundtrip fuzzing that now also covers adaptive filtering has found a bug, so that's cool

Shnatsel:
I cannot reproduce the crash locally just by fuzzing, and I don't have the permissions to access the testcase that fuzzing has found. @HeroicKatora do you have access? If so, could you post the file on the issue tracker so I could investigate what went wrong? I'm pretty sure this is a bug that also applies to the current version, since I didn't change much about the algorithms, so it could be a significant real-world issue in existing versions.

Shnatsel:
I left the fuzzer running locally overnight, but it failed to rediscover the issue after 1 billion executions.

src/common.rs
```diff
@@ -333,6 +333,67 @@ impl Default for Compression {
...
+    /// much faster than what is typical for PNG images.
+    /// Other decoders (e.g. libpng) do not get a decoding speed boost from this mode.
+    FdeflateUltraFast,
+
+    /// Uses [flate2](https://crates.io/crates/flate2) crate with the specified [compression level](flate2::Compression::new).
+    ///
+    /// Flate2 has several backends that make different trade-offs.
+    /// See the flate2 documentation for the available backends for more information.
+    Flate2(u32),
+    // TODO: Zopfli?
```
kornelski, line 369: Should this be dependent on feature flags?

If they exist unconditionally, then they can either fail at run time if a dependency is not added, or we'll have to have all the deps included all the time.

Shnatsel: All the modes listed right now are unconditional dependencies.

I'm really not a fan of things failing at runtime due to improper build time configuration, so for Zopfli I was going to gate that enum variant behind a feature. That way we could let the API user decide whether want it and enable the feature, decide they don't want it and disable the feature, or pass on the optionality of this by exposing their own feature, which would affect this enum.

I'm happy to hear your thoughts on how an optional Zopfli dependency could be added in the future, and how can we adjust this API to make that easier.

Shnatsel: There are even nice rustdoc annotations to make it clear that something is only available when a certain feature is enabled.

Shnatsel:
TODO: finish making adaptive filtering the default and document it as such. Needs #506 to be merged first because I don't want to create editing conflicts with myself.

src/common.rs
```diff
@@ -303,28 +305,26 @@ impl AnimationControl {
...
-    #[deprecated(
-        since = "0.17.6",
-        note = "use one of the other compression levels instead, such as 'fast'"
-    )]
-    Rle,
+    /// Significantly outperforms libpng and other popular encoders
+    /// by using a [specialized DEFLATE implementation tuned for PNG](https://crates.io/crates/fdeflate).
+    Fast,
+    /// Balances encoding speed and compression ratio
+    Default,
```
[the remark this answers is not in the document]
Shnatsel: Renamed to Balanced

src/encoder.rs
```diff
@@ -295,11 +296,28 @@ impl<'a, W: Write> Encoder<'a, W> {
...
+        self.info.compression_deflate = DeflateCompression::from_simple(compression);
+        // choose the filter based on the requested compression
+        match compression {
+            Compression::None => {
+                self.set_filter(FilterType::NoFilter); // with no DEFLATE filtering would only waste time
+                self.set_adaptive_filter(AdaptiveFilterType::NonAdaptive);
+            }
+            Compression::Fast => {
+                self.set_filter(FilterType::Up); // fast and avoids long backreferences in DEFLATE stream
+                self.set_adaptive_filter(AdaptiveFilterType::NonAdaptive);
```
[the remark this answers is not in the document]
Shnatsel: I've seen adaptive filtering degrade encoding performance by 2x on this file: 
https://commons.wikimedia.org/wiki/File:Exoplanet_Phase_Curve_(Diagram)_(01HK57P2YHV18MMV0RG5N7HY70).png

I suspect the non-vectorized Paeth filter (without nightly) is to blame.

Do you think it makes sense to have `Fast` and `Fastest` presets that use `Adaptive` and `Up` filters respectively?

src/common.rs
```diff
@@ -333,6 +333,67 @@ impl Default for Compression {
     }
 }
 
+/// Advanced compression settings with more customization options than [Compression].
```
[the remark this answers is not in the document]
Shnatsel: I'm not sure what you mean by that. If we change the mapping, we can just deprecate the original option like we did with `Huffman` and `Rle`, then start mapping it to something else. The non-deprecated ones work as advertised, so no need to add disclaimers, I think?

src/encoder.rs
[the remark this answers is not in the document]
Shnatsel: I've switched `Fast` to fdeflate+adaptive and made a new `Fastest` preset with fdeflate+up

src/common.rs
```diff
@@ -333,6 +333,67 @@ impl Default for Compression {
...
+///
+/// Note that this setting only affects DEFLATE compression.
+/// Another setting that influences the compression ratio and lets you choose
+/// between encoding speed and compression ratio is the *filter*,
+/// See [FilterType] and [AdaptiveFilterType].
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy)]
+pub enum DeflateCompression {
```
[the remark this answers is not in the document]
Shnatsel: I don't know if we can make any promises about decoding speed without targeting a specific version of a specific decoder. Fdeflate, miniz_oxide, zlib, zlib-ng, Go and WUFFS might very well have entirely different happy paths.

Shnatsel:
I think I've addressed all the feedback so far (good calls btw!) and also added the implementation of #505 to this PR, as you requested.

src/common.rs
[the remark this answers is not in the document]
fintelia: Yeah, would take more investigation. Not something we need to decide on now.

src/common.rs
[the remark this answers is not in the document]
fintelia: Arguably, changing the documented behavior was a breaking change even though we deprecated at the same time. I'd rather just be explicit that we don't consider changing the underlying implementation of any of these to be a breaking change.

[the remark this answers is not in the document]
Shnatsel: I've documented it explicitly, PTAL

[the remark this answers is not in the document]
fintelia: Looks good to me, thanks!

src/common.rs
```diff
@@ -303,33 +305,103 @@ impl AnimationControl {
...
-        note = "use one of the other compression levels instead, such as 'fast'"
-    )]
-    Huffman,
-    #[deprecated(
-        since = "0.17.6",
-        note = "use one of the other compression levels instead, such as 'fast'"
-    )]
-    Rle,
+    /// Significantly outperforms libpng and other popular encoders
+    /// by using a [specialized DEFLATE implementation tuned for PNG](https://crates.io/crates/fdeflate),
```
okaneco, line 324: A link backing up these claims would be nice here and in the new updated readme.

src/common.rs
kornelski: I'd prefer to ship this with a zopfli Cargo feature, even if it's default on, even if it doesn't actually disable Zopfli yet, only the config option. 

This is because addition of a feature flag would be a breaking change, but after that the change whether the flag actually removes zopfli dependency isn't breaking.

Shnatsel: The enum is marked `#[non_exhaustive]` so the addition of new variants is not a breaking change.

src/common.rs
```diff
@@ -303,33 +305,103 @@ impl AnimationControl {
...
+    /// while still providing better compression ratio than the fastest modes of other encoders.
+    Fast,
+    /// Balances encoding speed and compression ratio
+    Balanced,
```
kornelski, line 328: Does this setting interact with Zopfli? Zopfli is so slow, that Zopfli's "Fast" is other compressors "High".
Should the speed be about the same across all back-ends, or should the speed be relative to backend's speed?

src/filter.rs
```diff
@@ -269,54 +269,87 @@ mod simd {
...
+            RowFilter::Paeth => Filter::Paeth,
+        }
+    }
+}
+
+impl Filter {
+    pub(crate) fn from_simple(compression: Compression) -> Self {
+        match compression {
+            Compression::NoCompression => Filter::NoFilter, // with no DEFLATE filtering would only waste time
+            Compression::Fastest => Filter::Up, // fast and avoids long backreferences in DEFLATE stream
```
kornelski, line 307: Is this always `Up`? No filter may compress better, so this could be adaptive that selects between None and Up.

Which makes me think that `Filter::Adaptive` is not sufficiently configurable. Filtering can have many strategies and try different sets of filters.

src/common.rs
Shnatsel: We'd just add another tier for zopfli, "ultra slow" or something.

src/filter.rs
Shnatsel: Yeah, `Up` is probably not a good choice for `Fastest` mode. It's not actually the fastest. I'll switch it over to `None`, good call.

src/filter.rs
```diff
@@ -269,54 +269,87 @@ mod simd {
...
+            RowFilter::Paeth => Filter::Paeth,
+        }
+    }
+}
+
+impl Filter {
+    pub(crate) fn from_simple(compression: Compression) -> Self {
+        match compression {
+            Compression::NoCompression => Filter::NoFilter, // with no DEFLATE filtering would only waste time
+            Compression::Fastest => Filter::None,
```
fintelia, line 307: `Filter::None` combines very poorly with `FdeflateUtraFast`, since fdeflate assumes a specific hard-coded symbol distribution. I'd expect a substantially better compression ratio with `Filter::Up` with minimal performance difference

src/filter.rs
Shnatsel: https://github.com/image-rs/image-png/pull/503#discussion_r1872203847

src/common.rs
kornelski: But a conditional removal of `Zopfli` variant via Cargo feature is a breaking change.

It would have to be completely removed before release in order to make it backwards-compatible to add it behind a flag.

Shnatsel: Just so we're on the same page: I have no plans to add a Zopfli variant in this PR. If/when a Zopfli variant arrives, it will be gated on a Cargo feature from the start.

Shnatsel: I've removed the mentions of Zopfli from this PR entirely.

src/common.rs
```diff
@@ -303,33 +305,103 @@ impl AnimationControl {
...
+    ///
+    /// Images encoded in this mode can also be decoded by the `png` crate slightly faster than usual.
+    /// Other decoders (e.g. libpng) do not get a decoding speed boost from this mode.
+    FdeflateUltraFast,
+
+    /// Uses [flate2](https://crates.io/crates/flate2) crate with the specified [compression level](flate2::Compression::new).
+    ///
+    /// Flate2 has several backends that make different trade-offs.
+    /// See the flate2 documentation for the available backends for more information.
+    Flate2(u32),
```
fintelia, line 378: After more experimentation with adding higher compression levels to `fdeflate`, I'm pretty confident that we'd be able to expose "standard" compression levels ranging from 1 to 9 for any alternative we switch to.

One aspect is that the various different flate2 backends actually have a somewhat significant about of variability about what each compression level means, so we wouldn't have to worry too much about exactly calibrating to match existing behavior. But additionally, these algorithms have like 5 to 10 different knobs to tune. So the challenge is really condensing them into a single number (which you want to do regardless, because users won't want to deal with setting all the parameters!) 

Long story short, I think we should drop the `Flate2` naming and just call this `Level` or something to that effect. And perhaps switch to a `u8` given that values above 9 have no impact.

Shnatsel: **TL;DR:** since this is the *advanced* API, not the high-level one, I'd prefer to keep it `Flate2` and also keep the documentation about the way it's implemented. We would add a deprecation notice to it if we ever start emulating it like we did with `Huffman` and `Rle`.

I am going to calibrate `wondermagick` this way against a known reference with a specific flate2 backend, zlib-rs. If/when the underlying implementation changes away from flate2, I would much rather get a warning from the compiler and a deprecation message saying that it's now emulated so that I could re-evaluate the tagets for the new backend and adjust the mapping on my end, as opposed to the targets silently changing. 

Knowing that `flate2` is no longer used would also let me drop the direct `flate2` dependency from my crate that I added to select the backend. Although perhaps `png` should just expose a `zlib-rs` feature and not have users add a direct `flate2` dependency at all? And we would just make it a no-op feature if we ever move away from it.

Shnatsel: On `u32` vs `u8`: originally I chose to stick to the `flate2` API which used `u32`, but you're right - it makes no sense and just confuses people. I've switched it over to `u8`.

fintelia:
Would this API be compatible with adding logic to automatically detect whether an image contains a small number of colors and automatically perform a palette transform for it?

That's a trick that libwebp uses to improve compression by a bit. Before encoding, they count whether the image contains 257 or more colors. If not, the consider encoding using a palette transform. According to their code comments, it is nearly always a benefit if the palette indexes fit in 4-bits (so they do it unconditionally) and is often beneficial for 8-bit palettes (so they compare the compression ratio both with and without).

Shnatsel:
As far as the configuration goes (ignoring streaming and such not touched by this PR) - yes, absolutely. 

The high-level compression options can transparently configure whether we're trying to palettize the image depending on the compression level, just like they already select the filtering mode. We can also expose an explicit control for it, separate from the advanced compression enum.

I think that would fit into this configuration API structure very well.

Shnatsel:
I resolved the merge conflicts that arose from the API-breaking changes in master.

Now that we're doing breaking changes anyway, I'd appreciate getting this merged before the branches diverge once again.

anforowicz:
Thanks @Shnatsel, @kornelski, @okaneco, and @fintelia for the PR, reviews, and for merging!

Any chance that you could please push a new release/version to https://crates.io/crates/png/versions?

A new official version would help me absorb some latest functional fixes (like ExiF support) into Chromium, but I would also like to experiment a little bit with the additional compression levels - there is a weird bug (see https://crbug.com/396176046#comment11 and https://crbug.com/396176046#comment15) where Rust PNG is having trouble sticking in Chromium (a prerequisite for Dev channel field trials) because it slows down tests that run debug binaries (yes, debug!  not release!).

Shnatsel:
@anforowicz You can adjust your build system to enable optimizations for PNG crate even in debug mode. In Cargo you would add this to the `Cargo.toml`:

```toml
[profile.dev]
opt-level = 3
```
(opt levels 1 and 2 may also be sufficient, check if the trade-off is worth it)

Or if you have a bigger Cargo project that depends on `png` and you want to enable optimizations only for the PNG crate, that would be 

```toml
[profile.dev.package.png]
opt-level = 3
```

If you're using bazel/blaze then there are probably ways to pass `-C opt-level=3` to the compiler even in debug mode, but I'm not familiar with those build systems so I cannot provide the exact recipe.

PR #506, opened 2024-09-26
Improve documentation on filters

Shnatsel:


src/filter.rs
```diff
@@ -297,12 +299,14 @@ impl FilterType {
     }
 }
 
-/// The filtering method for preprocessing scanline data before compression.
+/// Adaptive filtering tries every possible filter for each row and keeps the best one.
+/// This improves compression ratio, but makes encoding slower.
```
fintelia, line 303:
```suggestion
/// Adaptive filtering tries every possible filter for each row and uses a heuristic to select the best one.
/// This improves compression ratio, but makes encoding slightly slower.
```

src/encoder.rs
```diff
@@ -1346,8 +1346,11 @@ impl<'a, W: Write> StreamWriter<'a, W> {
     /// Set the used filter type for the next frame.
     ///
     /// The default filter is [`FilterType::Sub`] which provides a basic prediction algorithm for
-    /// sample values based on the previous. For a potentially better compression ratio, at the
-    /// cost of more complex processing, try out [`FilterType::Paeth`].
+    /// sample values based on the previous.
+    ///
+    /// For optimal compression ratio you should enable adaptive filtering
```
fintelia, line 1351: The current adaptive filter approach is usually better than any specific hardcoded option, but it isn't optimal.

The heuristic works by summing up the absolute values of the bytes (each treated as a signed integer) and picking whichever filter gets the lowest value. That works surprisingly well, but there are likely a couple ways we could do better

Shnatsel: Well, best compression ratio this library is capable of, anyway. I'm struggling to communicate it better than this without making the doc comment long and unwieldy.

Shnatsel:
Thanks for the review and corrections!

I'm going to go ahead and merge this to avoid editing conflicts with myself, but please feel free to tweak the wording further, since you clearly know more than I do.

Issue #511, opened 2024-09-28
Vectorize Paeth filtering on stable


Shnatsel:
Paeth filtering was remarkably resistant to autovectorization, and it is the only instance where we had to resort to the nightly-only portable SIMD API.

Now that we're looking to make adaptive filtering the default, Paeth filter performance is going to become more important.

I've taken a stab at getting it to autovectorize on latest stable, and the results are promising!

Portable SIMD version (what we're looking to match): https://godbolt.org/z/Pdhx4Kdd8

What I've got so far: https://godbolt.org/z/63av9hbbY

You can see that it vectorized everything except the final loop that selects values; that just got unrolled into lots and lots of conditional moves, so at least it's unrolled and branchless and might benefit from ILP.

Both versions also benefit from `-C target-cpu=x86-64-v3` over the default, but the difference isn't dramatic.

Shnatsel:
I also tried a 1:1 port of the SIMD filtering algorithm, but this is as far as I got before things started falling apart:
https://godbolt.org/z/W49qeGv3o
I got all the SIMD operations polyfilled, but if I go and actually swap them out more than I already did, the vectorization seems to fall apart completely.

Shnatsel:
I got it to vectorize the comparisons too, so only the final value selection is still scalar: https://godbolt.org/z/TPdoWPPMd

Shnatsel:
But the direct translation of Portable SIMD code is still pretty gnarly and doesn't look any faster: https://godbolt.org/z/b7G3xnsj8

Shnatsel:
I've wired up my most successful attempt, let's see if it beats the autovectorization we already have in place:
https://github.com/Shnatsel/image-png/tree/simder-paeth

Not sure how to benchmark it though, the filters are already cheap compared to the rest of encoding

Shnatsel:
Well, turns out the solution was right in front of me all along. The filtering code currently in use already vectorizes perfectly, resulting in code identical to the explicit SIMD version: https://godbolt.org/z/P8WWsTs6Y

Let me double-check if explicit SIMD still gets us any gains in unfiltering. If so, then we can adapt this trivial approach to get the same benefits on stable.

Shnatsel:
Okay, I have a [branch](https://github.com/Shnatsel/image-png/tree/autovec-paeth) where I've simply replaced the handwritten SIMD implementations with autovectorized ones, and the results are pretty surprising.

[...]

PR #512, opened 2024-09-28
Replace handwritten SIMD implementation with autovectorization for +10% perf

Shnatsel:
This gives a +10% performance boost on 3bbp and 6bbp unfiltering benchmarks. 

cc @anforowicz who wrote the explicit SIMD code

The downside of autovectorization is that it can be finicky and is not at all guarateed. So this may potentially be more fragile than explicit SIMD.

On the flip side, we've been relying on autovectorization for performance for ages and it has always worked once we got it working once. I've checked [this sample](https://godbolt.org/z/8jTEr1Kbv) on Godbolt and it didn't vectorize on Rust v1.60, but every single version starting from 1.61 does vectorize it.

ARM seems to benefit as well, resulting in vectorized and shorter code on godbolt. I don't have an aarch64 machine to measure it though.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -13,6 +13,32 @@ mod simd {
     use std::simd::num::{SimdInt, SimdUint};
     use std::simd::{u8x4, u8x8, LaneCount, Simd, SimdElement, SupportedLaneCount};
 
+    /// Scalar Paeth function wrapped in SIMD scaffolding.
+    ///
+    /// This is needed because simply running the function on the inputs
+    /// makes the compiler think our inputs are too short
+    /// to benefit from vectorization.
+    /// Putting it in SIMD scaffolding fixes that.
+    /// https://github.com/image-rs/image-png/issues/511
+    ///
+    /// Funnily, the autovectorizer does a better job here
+    /// than a handwritten algorithm using std::simd!
+    /// We used to have a handwritten one but this is just faster.
+    fn paeth_predictor<const N: usize>(
+        a: Simd<i16, N>,
+        b: Simd<i16, N>,
+        c: Simd<i16, N>,
+    ) -> Simd<i16, N>
+    where
+        LaneCount<N>: SupportedLaneCount,
+    {
+        let mut out = [0; N];
+        for i in 0..N {
+            out[i] = super::filter_paeth_decode_i16(a[i].into(), b[i].into(), c[i].into());
+        }
+        out.into()
+    }
+
     /// This is an equivalent of the `PaethPredictor` function from
     /// [the spec](http://www.libpng.org/pub/png/spec/1.2/PNG-Filters.html#Filter-type-4-Paeth)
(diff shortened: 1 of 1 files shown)


Shnatsel:
There's some background on this in #511

I'm still hoping to coax autovectorization into working without any `std::simd` usage at all, but I haven't managed it so far. If I succeed, that will probably supersede this PR - but I figured it's a good idea to start the discussion about replacing `std::simd` with autovectorization early.

fintelia:
> figured it's a good idea to start the discussion about replacing std::simd with autovectorization early, in case anyone is opposed to this approach.

I support the switch. Using `std::simd` was only begrudgingly accepted when we couldn't figure out a way to get the relevant code to autovectorize.

Shnatsel:
To clarify, this PR still requires `std::simd` types in function signatures, and thus still requires nightly compiler.

I'm still trying (and failing) to get it to work on stable. For the purposes of this review let's assume that nightly is still going to be required, even for the autovec implementation.

Shnatsel:
Okay I've finally figured out the 3 and 6 bpp case, PR up: #513

Shnatsel:
This performs better in benchmarks, and produces shorter but fully vectorized assembly on both x86_64 and Aarch64. 

Since this seems to be an unambiguous improvement, I'm going to go ahead and merge this so that the PR doesn't stall.

src/filter.rs
```diff
@@ -32,28 +32,26 @@ mod simd {
...
+    /// [the spec](http://www.libpng.org/pub/png/spec/1.2/PNG-Filters.html#Filter-type-4-Paeth)
+    /// except that it simultaneously calculates the predictor for all SIMD lanes.
+    /// Mapping between parameter names and pixel positions can be found in
+    /// [a diagram here](https://www.w3.org/TR/png/#filter-byte-positions).
+    ///
+    /// Examples of how different pixel types may be represented as multiple SIMD lanes:
+    /// - RGBA => 4 lanes of `i16x4` contain R, G, B, A
+    /// - RGB  => 4 lanes of `i16x4` contain R, G, B, and a ignored 4th value
+    ///
+    /// The SIMD algorithm below is based on [`libpng`](https://github.com/glennrp/libpng/blob/f8e5fa92b0e37ab597616f554bee254157998227/intel/filter_sse2_intrinsics.c#L261-L280).
```
okaneco, line 52: This comment isn't correct.

As it states on L72, the logic is a translation of how pa, pb, and pc are calculated in `filter::filter_paeth` https://github.com/image-rs/image-png/blob/3fbbbb188e967547563496282591c52fc5892746/src/filter.rs#L362-L401

Shnatsel: Thanks for pointing this out!

I didn't actually add this comment, it's an artifact of how Github calculated the diff. You can see the same lines removed further up.

I'd be happy to merge a PR correcting it, though!

PR #513, opened 2024-09-29
Autovectorize 3 and 6 bpp Paeth unfiltering on stable

Shnatsel:
Make 3 and 6 bpp codepaths operate on i16 vectors of widths 4 and 8 respectively. Gains +5% for 3bpp and +10% for 6bbp case. This works on stable and matches the current `std::simd` implementation in master in terms of performance.

Part of #511

This does not obsolete #512 which pushes SIMD performance even further.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -340,6 +340,26 @@ fn filter_paeth_decode(a: u8, b: u8, c: u8) -> u8 {
     out
 }
 
+fn filter_paeth_decode_i16(a: i16, b: i16, c: i16) -> i16 {
+    // Decoding seems to optimize better with this algorithm
+    let pa = (b - c).abs();
+    let pb = (a - c).abs();
+    let pc = ((a - c) + (b - c)).abs();
+
+    let mut out = a;
+    let mut min = pa;
+
+    if pb < min {
+        min = pb;
+        out = b;
+    }
+    if pc < min {
+        out = c;
+    }
+
+    out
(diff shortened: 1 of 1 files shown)


okaneco:
This scalarizes the code for the bpp=3 case
https://godbolt.org/z/qKsYas8TP
and seems to scalarize part of the bpp=6 case
https://godbolt.org/z/c4oeEWsoh

The current version was massaged through several iterations to eventually get auto-vectorization. I'm entirely open to the idea that scalarized code can be faster than auto-vectorized SIMD but I'm not sure that's the case here.

We have to resort to SIMD types to force vectorization in these non-power-of-two cases. The compiler sees that you never write to part of the array and discards any chunk handling you try to force even though you're using 4 and 8 element arrays.

Shnatsel:
Thanks for the info! I didn't stare at the assembly for this PR after doing a great deal of it in other ones, I just measured it and posted the version that improved the benchmarks. It is possible that this is only faster on the latest CPUs with very wide ILP.

Could you run `cargo bench --bench=unfilter --features=benchmarks Paeth` on your machine with and without this PR? It would be a valuable data point, so that I'm not optimizing for a single specific CPU.

Shnatsel:
Since this scalarizes code and that probably only benefits the latest CPUs with very high ILP potential, and the gains aren't significant enough to switch over, I'm going to go ahead and close this.

PR #516, opened 2024-10-01
Upgrade bitflags to 2.0.

khuey:


HeroicKatora (commented):
The new version of `bitflags` no longer generates [`from_bits_unchecked`](https://docs.rs/png/latest/png/struct.Transformations.html#method.from_bits_unchecked). I think the next release is planned to be a breaking one anyways, just qualifying that there technically is an interface difference.

khuey:
Mmm, yes, that's a good point.

HeroicKatora (approved):
I've never seen a need for the removed function though. So this should go in the next major version.

atouchet:
Note that there is a similar existing PR in #400.

hydra:
any chance we can get this or #400 merged?  I agree with this comment, https://github.com/image-rs/image-png/pull/400#issuecomment-1913325342

I personally find it easier to deal with a few releases each with small breaking changes that one release which has lots of breaking changes as you can fix them one at a time instead of all in one go.  This makes resulting downstream changes/upgrades easier to review for people/teams using this crate.

fintelia:
Closing in favor of #400

PR #517, opened 2024-10-03
Next frame tracking tests and refactoring

anforowicz:
PTAL?  Please see the commit descriptions for more details.

Issue #510, opened 2024-09-27
Need public API to seek to another APNG frame


anforowicz:
# Disclaimer

This issue is based on my current, evolving understanding of the requirements that Chromium and Skia architectures impose upon an underlying codec implementation.  I am opening this issue early to facilitate a transparent discussion, but before accepting the new APIs we should definitely require that I first finish prototyping the APNG-focused part of the Chromium/Blink/Skia integration (as a proof-of-concept that the new APIs actually work and solve my use case).

# Problem 1a: No way to get to next image data

When [`Reader.next_interlaced_row`](https://docs.rs/png/latest/png/struct.Reader.html#method.next_interlaced_row) reaches the end of a frame, then it gets "stuck" - subsequent calls will continue [returning `None`](https://github.com/image-rs/image-png/blob/29a8342e647a66d8ae44e43713826a9fdb5f81c0/src/decoder/mod.rs#L570) (which correctly indicates that no more rows in the current frame) and there is no way to proceed to the next frame.

# Problem 1b: No way to read the next `fcTL` chunk

To successfully decode a frame, one needs to have its `FrameControl` data, but `Info.frame_control` won't be updated until the `fcTL` chunk is encountered and parsed.  There is no way to ask `Reader` to get the next `FrameControl` metadata.

# Problem 2: No way to go back and decode an earlier frame

Skia and Blink abstractions allow asking a codec to (again) decode an earlier frame:

* Skia allows that via [`Options.fFrameIndex`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/include/codec/SkCodec.h;l=349-354;drc=f522344e45882da4c7f7cb1b3a0a7bd747d654bb) parameter of `SkCodec::onStartIncrementalDecode` (and `onGetPixels`) which can specify an earlier frame.
    - Example implementation can be found in [the `SkWuffsCodec` here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/codec/SkWuffsCodec.cpp;l=463;drc=24b401ddedc61ee5e5f6dfe6d2a0b4a853c8eace)
[...]

PR #518, opened 2024-10-03
New public api to advance to the next frame

anforowicz:
PTAL?

Note that this PR is a continuation of (i.e. is based on) https://github.com/image-rs/image-png/pull/517.  This means that you should only look at the last commit really.  And this is why I am only submitting this PR as a draft for now.  I am splitting things into 2 PRs, because I think the other PR is relatively non-controversial, but we may want to give ourselves a little bit of time to review this PR.

This PR _partially_ address the issues raised in https://github.com/image-rs/image-png/issues/510

This PR introduces a new public API that can be used to 1) read an `fcTL` chunk **before** starting to decode a frame via [`next_frame`](https://docs.rs/png/latest/png/struct.Reader.html#method.next_frame) and 2) go to the next frame after finishing all rows via [`next_row`](https://docs.rs/png/latest/png/struct.Reader.html#method.next_row) / [`next_interlaced_row`](https://docs.rs/png/latest/png/struct.Reader.html#method.next_interlaced_row).  I see the following potential open questions:

* **Return type**:
    - I currently propose to return `Result<&FrameControl, …>`.  I think that this helps to reinforce that the API moves to the next `FrameControl` / `fcTL` chunk.
    - But in theory the new API could just return `Result<(), DecodingError>`
* **Naming**:
    - `next_frame_info` as currently proposed?
    - Maybe `next_frame_control` (since the struct is called `FrameControl` rather than `FrameInfo`)?
    - Alternatives that reflect that the private API wrapped here is called `read_until_image_data`: `read_until_next_image_data`?  `read_until_next_frame_data`?
* **Behavior**
    - The internal `read_until_image_data` panics if called when the previous image data hasn’t been fully processed (i.e. when `subframe.consumed_and_flushed` is still `false`).  The panic comes from `assert!(buf.is_empty())` in `fn read_until_image_data`.
    - I propose that the public API checks `subframe.consumed_and_flushed` and avoids the panic by discarding the previous image data if necessary (by calling `finish_decoding`).
    - Alternatively we could forbid calling the new public API when `subframe.consumed_and_flushed` is `false` (either panicking, or returning a `DecodingError::Parameter`).
        - One undesirable result of this alternative affects a scenario where the client wants to skip the IDAT frame (e.g. because it doesn’t have an `fcTL` chunk and is not part of the animation).  To handle this scenario the client would have to remember a piece of state that says what to do when recovering from `UnexpectedEof`: whether to 1) continue discarding the image data vs 2) advance to the new image data by calling the new `next_frame_info` API
* **Docs**
    - Anything else to mention (or change) in the docs?
    - I wondered whether an example in the doc comment would be useful, but I couldn’t come up with something illustrative, self-contained, and succinct.

WDYT?


src/decoder/stream.rs
```diff
@@ -2282,6 +2282,15 @@ mod tests {
...
 
+    fn get_fctl_sequence_number(reader: &Reader<impl Read>) -> u32 {
+        reader
+            .info()
+            .frame_control
+            .as_ref()
+            .unwrap()
+            .sequence_number
+    }
+
```
anforowicz, line 2293: The new helper function helps to make the tests more compact...

src/decoder/stream.rs
```diff
@@ -2295,7 +2304,27 @@ mod tests {
...
+        let mut reader = create_reader_of_ihdr_idat();
+
+        let err = reader
+            .next_frame_info()
+            .expect_err("Main test - expecting error");
+        assert!(
+            matches!(&err, DecodingError::Parameter(_)),
+            "Unexpected kind of error: {:?}",
+            &err,
+        );
```
anforowicz, line 2327: This test would fail (returning `FormatError` / `MissingImageData` error instead) without taking into account the "One remaining frame will be consumed by the `finish_decoding` call below" piece at the top of the new API.

anforowicz:
Oh, I forgot to mention that I've prototyped using this API in WIP http://review.skia.org/c/skia/+/902941 and WIP https://crrev.com/c/5786777 where the new API helps to pass additional tests (`AnimatedPNGTests.ParseAndDecodeByteByByte` in Chromium's `blink_platform_unittests`).  So, I think I can say that I've validated that the new API is useful and helpful in at least some contexts :-).  FWIW I note that the prototype ignores/discards the return/ok value from the new API.

kornelski (approved):

PR #523, opened 2024-10-09
Simplify and encapsulate `ReadDecoder` implementation

anforowicz:
Hello!

This PR isn't strictly necessary for any major feature requests or bug fixes, but I think that this PR is still desirable - I think the resulting encapsulation helps with readability of the code.

I hope that splitting this PR into smaller, granular commits helps to follow along and review the changes.

PTAL?

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -213,6 +213,7 @@ impl<R: Read> Decoder<R> {
             transform: self.transform,
             transform_fn: None,
             scratch_buffer: Vec::new(),
+            finished: false,
         };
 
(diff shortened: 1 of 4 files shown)


anforowicz:
My appologies for the force push above.  I've realized that in the last commit (when moving some of the code into the new file) I didn't preserve the comment tweaks that I've made in the earlier commits after a self-review...

fintelia (commented):
I think this is a positive change

kornelski:
BTW, I've redesigned the same bits in the `gif` crate.

PR #527, opened 2024-10-30
Extract a separate `unfiltered_rows_buffer` module.

anforowicz:
Just a minor refactoring to simplify `mod.rs` and push some of the complexity into a separate, new `unfiltered_rows_buffer` module.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -2,11 +2,13 @@ mod interlace_info;
 mod read_decoder;
 pub(crate) mod stream;
 pub(crate) mod transform;
+mod unfiltering_buffer;
 mod zlib;
 
 use self::read_decoder::{ImageDataCompletionStatus, ReadDecoder};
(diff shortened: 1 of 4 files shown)


fintelia:
The code itself looks good to me, but I'm not thrilled with the naming. The buffer generally holds both filtered and unfiltered data at the same time and while it is true that the current and previous rows are present, there also might be tens/hundreds of kilobytes of data for rows beyond that. I'm not quite sure what a better name might be

anforowicz:
> The code itself looks good to me, but I'm not thrilled with the naming. The buffer generally holds both filtered and unfiltered data at the same time and while it is true that the current and previous rows are present, there also might be tens/hundreds of kilobytes of data beyond that which doesn't correspond to a specific number of rows. I'm not quite sure what a better name might be

Fair feedback :-).  `RawImageData`?  `RawScanlines`?  `UnfilteringBuffer`?

HeroicKatora (approved):

HeroicKatora:
Thank you, both code and naming fit well in the existing code flow :+1:

PR #528, opened 2024-10-31
Add support for parsing `mDCv` and `cLLi` chunks.

anforowicz:
PTAL?

This PR is needed for parity with the existing PNG decoder in Blink / Chromium - see https://crbug.com/376550658.

IIUC Chromium and Blink only require an ability to decode these chunks - there is no need for encoding support (e.g. [the search query here](https://source.chromium.org/search?q=cLLi%20f:png&sq=&ss=chromium) only finds decoder bits).


anforowicz:
The new test is based on [the `PNGTests.HDRMetadata` test](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/png/png_image_decoder_test.cc;l=1536-1565;drc=bde3d0d4b13aee737848caba9b62c37adbbe5806) from Chromium.

src/decoder/stream.rs
```diff
@@ -1271,6 +1274,83 @@ impl StreamingDecoder {
...
+                max_frame_average_light_level,
+            })
+        }
+
+        // Ignoring subsequent, duplicated chunks (if any).
+        if info.content_light_level.is_some() {
+            return Decoded::Nothing;
+        }
+
+        info.content_light_level = parse(&self.current_chunk.raw_bytes[..]).ok();
```
fintelia, line 1350: Might be cleaner to invert the condition and do the parsing if statement:
```suggestion
        if info.content_light_level.is_none() {
            info.content_light_level = parse(&self.current_chunk.raw_bytes[..]).ok();
        }
```

src/decoder/stream.rs
```diff
@@ -1271,6 +1274,83 @@ impl StreamingDecoder {
...
+
+        info.mastering_display_color_volume = parse(&self.current_chunk.raw_bytes[..]).ok();
+        Decoded::Nothing
+    }
+
+    // NOTE: This function cannot return `DecodingError` and handles parsing
+    // errors or spec violations as-if the chunk was missing.  See
+    // https://github.com/image-rs/image-png/issues/525 for more discussion.
+    fn parse_clli(&mut self) -> Decoded {
+        let info = self.info.as_mut().unwrap();
```
fintelia, line 1332: nit: Could you move this line after the nested function?

fintelia (approved):
This looks good to me. Left a few minor style comments

src/decoder/stream.rs
anforowicz: Done

src/decoder/stream.rs
anforowicz: Done

PR #529, opened 2024-10-31
Add support for parsing `cICP` chunks.

anforowicz:
This commit is needed for parity with the existing PNG decoder in Blink / Chromium - see https://crbug.com/376758571.

I think that we should opaquely return `u8` for `color_primaries` and `transfer_function`.  Being opaque will help with extensibility (in case ITU-T-H.273 gets augmented in the future with other table entries / configurations).

I am less sure about the other 2 fields:

* I think that we should also represent `matrix_coefficients` as `u8` and return it even though the parser rejects anything other than 0 (same as `blink::PNGImageDecoder` in Chromium - see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/png/png_image_decoder.cc;l=210-214;drc=39736f289b63165a3427e7b09c83c65e4ed7f323)).
* I think it is okay to represent `video_full_range_flag` as an `enum`.
    - AFAICT only 2 values are supported (see [the spec snippet here](https://www.w3.org/TR/png-3/#cICP-chunk:~:text=If%20Video%20Full%20Range%20Flag,is%20a%20narrow%2Drange%20image)).  And in that case having named values is nice.
    - The spec says that "each of the fields of the cICP chunk corresponds to the parameter of the same name in ITU-T-H.273", but I wasn't able to find "Video Full Range Flag" in there (or in an earlier revision from 2017) when searching for various keywords, including "range" and "narrow".


fintelia:
> I think that we should also represent matrix_coefficients as u8 and return it even though the parser rejects anything other than 0 (same as blink::PNGImageDecoder in Chromium - see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/image-decoders/png/png_image_decoder.cc;l=210-214;drc=39736f289b63165a3427e7b09c83c65e4ed7f323)).

Seems reasonable.

> The spec says that "each of the fields of the cICP chunk corresponds to the parameter of the same name in ITU-T-H.273", but I wasn't able to find "Video Full Range Flag" in there (or in an earlier revision from 2017) when searching for various keywords, including "range" and "narrow".

It is described in section 8.3 of the [7/24 version of ITU-T-H.273](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-H.273-202407-I!!PDF-E&type=items) under the name "VideoFullRangeFlag". The values don't seem to have names and are just listed as 0 and 1.

I think it might be worth encoding it just as a `bool` rather than an enum. Honestly, given that narrow range images are [extremely rare](https://github.com/w3c/png/issues/312#issuecomment-2327349614) we might even want to add a note that users probably shouldn't bother handling them.

fintelia (approved):

anforowicz:
> > The spec says that "each of the fields of the cICP chunk corresponds to the parameter of the same name in ITU-T-H.273", but I wasn't able to find "Video Full Range Flag" in there (or in an earlier revision from 2017) when searching for various keywords, including "range" and "narrow".
> 
> It is described in section 8.3 of the [7/24 version of ITU-T-H.273](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-H.273-202407-I!!PDF-E&type=items) under the name "VideoFullRangeFlag". The values don't seem to have names and are just listed as 0 and 1.

Ack.

> I think it might be worth encoding it just as a `bool` rather than an enum. Honestly, given that narrow range images are [extremely rare](https://github.com/w3c/png/issues/312#issuecomment-2327349614) we might even want to add a note that users probably shouldn't bother handling them.

Thanks for the feedback.  Done (changed to `bool` + tweaked the name of the last field;  and also tweaked the doc comments of the last 2 fields).

Issue #525, opened 2024-10-19
Corrupt auxillary chunks should be ignored, not treated as fatal errors


fintelia:
The PNG spec says:

> Unexpected values in fields of known chunks (for example, an unexpected compression method in the IHDR chunk) shall be checked for and treated as errors. However, it is recommended that unexpected field values be treated as fatal errors only in critical chunks. An unexpected value in an ancillary chunk can be handled by ignoring the whole chunk as though it were an unknown chunk type.

Right now, [`parse_chunk`](https://github.com/image-rs/image-png/blob/ffed4ded8e8a7e41d4dabcf6201218dc51cd4821/src/decoder/stream.rs#L951) instead treats any error in any known chunk as fatal and sets the decoding state to `None`.

anforowicz:
Good catch - thanks for filing the bug.

A few thoughts that I hope are helpful:

* I wonder if [the PNG spec's notion of "auxiliary"](http://www.libpng.org/pub/png/spec/1.2/PNG-Structure.html#Chunk-naming-conventions) can be directly applied, because `acTL` and `fcTL` chunks are kind of critical to APNG-aware decoders.  OTOH, maybe an error parsing these chunks can be indeed treated as-if the chunk was missing.
* One way to fix this would be to tweak `fn parse_chunk` so that it turns `Err(_)` into `Ok(Decoded::Nothing)` for auxiliary chunks.  The main risk is whether a `fn parse_somechunk` leaves the decoded in a bad state when returning an error (i.e. we'd have to "be careful" going forward that `fn parse_fctl` doesn't do that - we'd have to move `self.inflated.reset()` and `self.ready_for_fdat_chunks = true` toward the end of the function - next to `self.info.as_mut().unwrap().frame_control = Some(fc)`).
* I've been recently reading https://joshlf.com/files/talks/Safety%20in%20an%20Unsafe%20World.pdf and in theory we could make reporting an error for an auxiliary chunk a compilation error.  One idea would be to make each chunk a separate type/struct and only allow constructing errors `where TChunkType: CriticalChunkTrait` or something like that.  This seems like quite a big and complicated refactoring - I am not sure if this is worth it...

kornelski:
APNG chunks started out as a non-standard hack, and it was long too late to fix their names when they got into the spec.

fintelia:
For reference, this is how the spec recommends handling APNG errors: 

[...]

PR #535, opened 2024-11-16
Nicer ignore of parse errors in sBIT

kornelski:


fintelia:
It would be nice to align the style with `parse_cicp`, `parse_mdcv`, and `parse_clli`. And ideally the rest of the optional metadata chunks as well.

kornelski:
Oh, that would be even bigger churn. I'd rather not.

If I were going to touch them, I'd change them all to the style of `sBIT`, because having parsing code that can't use `Result` is too cumbersome, and having extra inner helper function is unnecessary if the failure can be ignored in `parse_chunk`.

kornelski:
I've considered changing error handling for all ancillary chunks, but if we haven't had reports of unnecessary failures caused by them, then maybe they're not too broken in the wild, and it's okay to keep stricter checks.

fintelia:
Switching the other ancillary chunks is tracked in #525.

We don't have to switch everything in one PR. But I would like to decide on _one_ approach to ignore errors from chunks and then make sure all changes move us towards that

Issue #536, opened 2024-11-23
Feature Request: bKGD info


ocpkn:
Currently, there is no straightforward way to access an image's background color from the `bKGD` chunk.

`png::stream::StreamingDecoder` lacks a method to parse these chunks and `png::common::Info` lacks a field for this data. This chunk type is in common use and it should be accessible to the client.

The benefits of this would include displaying images with the intended background color as well as combining an image's alpha channel with the background color to produce an opaque image.

kornelski:
Is kBGD actually implemented by anyone? I think IE6 was the last application that couldn't handle alpha and had a reason to use it. I haven't seen any other application that used it.

ocpkn:
Some programs will erase color data from transparent pixels when exporting an image to PNG and/or save the background color in a bKGD chunk. The current version of GIMP does both by default.

fintelia:
Seems reasonable to add bKGD support. Feel free to make a PR if you'd like, there's been a few recent PRs adding other chunks that you can use for reference

PR #538, opened 2024-11-28
Add bKGD chunk parsing

ocpkn:
Implements changes discussed in #536


fintelia:
Please look at how https://github.com/image-rs/image-png/pull/529 handles errors. We don't want an optional chunk to completely prevent decoding the image (some other chunks do currently have that issue, we've got an open issue to fix it...)

kornelski:
Thank you

PR #539, opened 2024-12-01
Port of stb_image optimized paeth unfiltering

Shnatsel:
10% end-to-end performance gain on a Paeth-filtered image I concocted with `convert -quality 49 in.png out.png`, which is comparatively big and Paeth images are rather common.

Godbolt shows much shorter assembly that gets autovectorized even in the bpp=3 case: https://godbolt.org/z/fq3EjvT4b

TODO:

 - [x] Have someone else confirm this performance boost exists and isn't isolated to my machine
 - [x] Exhaustively test this implementation against the reference
 - [x] Do something about the nightly variant of Paeth unfiltering

Also done:

- [x] Fix `unstable` feature dramatically regressing performance on ARM
- [x] Add CI job to make sure `unstable` feature compiles
- [x] Add CI job to run tests on ARM

PR #540, opened 2024-12-01
Use stb's paeth predictor

kornelski:
The stb's implementation doesn't seem to be always faster.

For me on Apple M3 it's faster for 1, 6, and 8-byte pixels. About the same for 4. *Half* the speed for 2 and 3. I'm curious if that's universal across platforms. On a Ryzen most are faster, except 6 Bpp.

| Bpp | 1 | 2 | 3 | 4 | 6 | 8 |
| - | - | - | - | - | - | - |
| Apple M3 | y | N | N | = | y | y |
| Ryzen 5 5600X | y | y | y | y | N | y |

I've used `cargo +nightly bench -Fbenchmarks Paeth --bench unfilter -- --save-baseline=…`

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -318,24 +318,17 @@ impl Default for AdaptiveFilterType {
     }
 }
 
+/// Adapted from stb_image.h 2.30
 fn filter_paeth_decode(a: u8, b: u8, c: u8) -> u8 {
-    // Decoding seems to optimize better with this algorithm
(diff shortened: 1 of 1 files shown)


fintelia:
Thanks for the measurements! I've incidentally also got a 5600X, so I can't offer any further data points

HeroicKatora:
Thanks for the comparative measurements, can be closed with #539 as implemented and constrained to arch x86.

PR #547, opened 2024-12-08
Make gAMA and cHRM fallback optional for sRGB

kornelski:
The `sRGB` chunk [is sufficient on its own](https://www.w3.org/TR/png-3/#color-chunk-precendence), so it'd be nice to be able to skip writing `gAMA`/`cHRM` fallback. I don't think there's value in adding the fallback by default. I assume that decoders too old to understand `sRGB` chunk are also too old to have good color profile support, so passing through unlabelled sRGB pixels may be safer for them.

The fallback is still possible if someone sets the fallback values explicitly, but I haven't bothered to increase API surface for that.

`set_srgb` was inconsistent with `set_source_gamma/chromaticities`, so this change is backwards compatible.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -725,6 +725,17 @@ impl Info<'_> {
             .raw_row_length_from_width(self.bit_depth, width)
     }
 
+    /// Mark the image data as conforming to the SRGB color space with the specified rendering intent.
+    ///
+    /// Any ICC profiles will be ignored.
+    ///
+    /// Source gamma and chromaticities will be written only if they're set to fallback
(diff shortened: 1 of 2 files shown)


fintelia:
The spec does say:
> It is recommended that a PNG encoder that writes the sRGB chunk also write a [gAMA](https://www.w3.org/TR/png-3/#11gAMA) chunk (and optionally a [cHRM](https://www.w3.org/TR/png-3/#11cHRM) chunk) for compatibility with decoders that do not use the sRGB chunk.

That said, I'd be open to merging this since I agree with you that there probably isn't much software out there that gAMA/cHRM but no sRGB

PR #551, opened 2024-12-19
Decode-only Info (part 1/x)

kornelski:
My plan is to make `Info` purely into information about a decoded image, without living in `Encoder` as a *third* type storing encoding options. Here's a small step with a couple of fields.


HeroicKatora (approved):
Sure, I like that plan.

PR #554, opened 2024-12-22
Info refactorings

kornelski:
* There's no need to have `ParseChunkData` return to reading after reallocating. It wants the whole chunk in memory, and chunk sizes are known ahead of time, so the vec can be allocated without resizing it.
* Since the chunk vec can have the exact size needed now, `Info` can `mem::take` it.
* Chunks can be parsed lazily, using `sBIT` as an example.
* I've added `warnings` to the `StreamingDecoder` to collect errors from optional chunks. Unfortunately it can't easily be in `Info`, because `Info` supports clone, and `DecodingError` does not (`io::Error` can't).
* Bumped MSRV to 1.65 ([99.7% support](https://lib.rs/stats#rustc-usage)) to have `let else`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ["1.57.0", stable, beta, nightly]
+        rust: ["1.65.0", stable, beta, nightly]
         os: [ubuntu-latest, windows-latest, macos-latest]
         features: [""]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


fintelia:
This PR is doing a bunch of different things, and I think would be easier to review as multiple separate PRs.

A few specific points:
* I'm not totally sold on the warnings concept. Is this something that would be actually used, or just added maintenance burden on our end?
* I want to think through the lazy parsing a bit more. For something like iCCP, it seems appealing to avoid decompressing the chunk unless requested. If we add a `Seek` bound, it may may sense to defer even loading the chunk contents unless requested. But sBIT is at most 4 bytes. We already get a bit carried away allocating a whole `Vec` (with 24 bytes of overhead!) to store it, I'm not sure what we gain by delaying the validity checks

PR #555, opened 2024-12-22
Avoid returning to ReadChunkData state

kornelski:
The chunk size is known ahead of time, and it's always fully read, so there's no need to have a separate state for growing it incrementally.


fintelia:
This might be better to defer until after we've gotten 0.18.0 out.

It has a few consequences that are a bit hard to think through: it allocates space for entire chunks ahead of time rather than waiting to see whether we hit an EOF first and it `mem::take`'s vectors in certain places without a `shrink_to_fit` first, so their capacity could be far larger than their current length.

PR #561, opened 2024-12-30
Actually run tests on a big endian architecture on CI

Shnatsel:
instead of just building on PowerPC and calling it a day

workflow definition is cribbed from image-webp


fintelia:
I'd also be fine just dropping the show example. The majority of it looks like it is about how to use `glium` to display images rather than anything to do with PNG decoding.

Shnatsel:
I deleted the glium example and that got powerpc CI working. Still passes, too!

PR #563, opened 2024-12-30
Make flate2 a feature flag

fintelia:
Right now this crate ships with both `fdeflate` and `flate2` as dependencies. There's a lot of duplicated code between them including two full decompressors and a chunk of duplicated logic on the encode path. If `fdeflate` gains a general purpose compressor, it would be nice for us to remove the duplication by default.

This PR moves the user-facing guarantee that `flate2` is used as the compression backend behind a non-default feature flag. Internally, `flate2` is still used for all compression levels other than `NoCompression` and `FdeflateUltraFast`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,12 @@ rand = "0.8.4"
 term = "1.0.1"
 
 [features]
-unstable = ["crc32fast/nightly"]
+flate2 = [] # Expose flate2 specific functionality
+unstable = ["crc32fast/nightly"] # Enable nightly-only features
 benchmarks = []
+
+[package.metadata.docs.rs]
+features = ["flate2"]
 
 [lints.rust]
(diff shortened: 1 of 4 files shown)


Shnatsel:
I agree with the motivation. However, I find this particular implementation rather confusing from the API user perspective. For the most common use case of opting in to zlib-rs for faster compression than miniz_oxide, the steps to enable that really aren't clear.

In the spirit of making easy things easy and hard things possible, I think we should just outright rename `Flate2` variant to `Level` and add `zlib-rs` toplevel feature that enables zlib-rs for encoding. All the other uses for messing with flate2 are sufficiently niche that they're not worth complicating the API surface for everyone.

Shnatsel:
I'll make a PR for the option I suggested and we can see if that's any better

kornelski (approved):

fintelia:
Closing in favor of #564

PR #565, opened 2024-12-31
Reworded docs for encoder::Writer::write_chunk()

Some1and2-XC:
Reworded a section from the `write_chunk()` method.

I felt as though the previous doc comment's wording was somewhat awkward. I added the size limit as well based on both the code from the function as well as from the [png spec](https://www.w3.org/TR/png/#table51).


kornelski:
I don't see where it panics. It reports Err if the length can't be encoded.

Some1and2-XC:
> I don't see where it panics. It reports Err if the length can't be encoded.

Oh I missed that, I'll edit it.

Issue #567, opened 2025-01-15
Unfiltering is slower on Windows than on Linux (especially with `std::simd`)


anforowicz:
# Problem description

TL;DR:
* `rustc` auto-vectorization produces different binary code on Windows than on Linux
* Turning off `unstable` feature of the `png` crate (i.e. avoiding `std::simd`) helps

I have been running Chromium-based PNG benchmarks (see https://crrev.com/c/4860210/25).  In the past I have been using my 2023 corpus of 1655 PNG images from top 500 websites (see [here](https://github.com/image-rs/image-png/discussions/416#discussioncomment-7436871) for more details).  This week I have tried running these microbenchmarks against 100 biggest PNG images from this corpus on Windows and on Linux and discovered that the ratio of total-decoding-runtime-when-using-Rust / total-decoding-runtime-when-using-C++ differs considerably across these OS platforms: 79% for Linux (i.e. Rust is ~20% faster) vs 101% for Windows (i.e. Rust is comparable).

I have investigated further by looking at 1 image with the biggest ratio (color type = 6 / RGBA;  bits = 8;  width x height = 740 x 3640;  all 740 rows use Paeth filter).  For this image, it turned out that unfiltering takes the most time and so I experimented with turning on and off the `unstable` / `std::simd` feature of the `png` crate:

|                                 | Windows     | Linux     |
|---------------------------------|-------------|-----------|
| `png` features = `['unstable']` | 147% - 151% | 93% - 94% |
| `png` features = `[]`           | 114% - 118% | 94%       |

Profiling notes below show how often `png::decoder::unfiltering_buffer::UnfilteringBuffer::unfilter_curr_row` is present in ETW/profiling samples on Windows or `perf record -e cpu-clock` on Linux (decoding the single file).  I am also including `fdeflate::decompress:Decompressor::read` for comparison.  The percentages are "inclusive" although this doesn't really matter for `unfilter_curr_row` which IIUC doesn't call any other functions):

|                         | unfilter_curr_row | Decompressor::read
|-------------------------|-------------------|--------------------
| Windows with `unstable` | 56.3%             | 5.7%
| Windows w/o `unstable`  | 43.2%             | 7.5%
| Linux w/o `unstable`    | 54.4%             | 11.0%
| Linux with `unstable`   | 54.4%             | 10.9%

[...]

PR #573, opened 2025-02-16
Remove RGBA codepaths using std::simd types since they regress perfor…

Shnatsel:
…mance on Windows, see #567

Keeps the actually different and interesting RGB (3 and 6 bpp) codepaths for now.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -43,23 +43,6 @@ mod simd {
         out.into()
     }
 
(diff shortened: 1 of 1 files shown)


Shnatsel:
This is probably too sweeping - this is a significant regression for the bpp=8 case on Zen 4. Dropping the bpp=4 case only should eliminate the Windows regression without tanking performance for bpp=8.

Shnatsel:
I proposed a more conservative version of this in https://github.com/image-rs/image-png/pull/574

PR #574, opened 2025-02-16
Drop the bpp=4 std::simd codepath

Shnatsel:
It is not beneficial on Linux and may regress performance on Windows

See https://github.com/image-rs/image-png/issues/567

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -804,11 +804,9 @@ pub(crate) fn unfilter(
                     }
                 }
                 BytesPerPixel::Four => {
-                    #[cfg(all(feature = "unstable", target_arch = "x86_64"))]
-                    {
-                        simd::unfilter_paeth_u8::<4>(previous, current);
-                        return;
-                    }
+                    // Using the `simd` module here has no effect on Linux
+                    // and appears to regress performance on Windows, so we don't use it here.
+                    // See https://github.com/image-rs/image-png/issues/567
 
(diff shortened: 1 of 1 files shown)


kornelski (approved):

anforowicz:
I've run some additional tests on my Windows machine, using Chromium build config ("total" = total Rust / C++ runtime ratio;  "geom" = geometric mean of 100 individual Rust / C++ runtime ratios;  for both measurements lower is better):

* Baseline (ToT (249a6e02c00cdcefc0d722ff495b036c70fdc1da / r1421435) + enabling `unstable` feature of the `png` crate):
    - total=0.98313; geom=0.93416
* Experiment1: Applying the changes from https://github.com/image-rs/image-png/pull/574 (I.e. not using `std::simd` for Paeth unfiltering for bpp=4): 
    - total=0.93095;  geom=0.89143
* Experiment2: Same as baseline, but disabling the `unstable` feature of the `png` crate:
    - total=0.86126;  geom=0.84954

So, I think the results above support merging https://github.com/image-rs/image-png/pull/574.

And (at least "on my machine" / "in my benchmark") the results from experiment2 shows additional improvements when not using `std::simd` in some additional cases.  I am still working on starting the Dev-channel field trials of Rust PNG in Chromium, but once we get the numbers from the field trials we should have more confidence on what to do with `std::simd` / `unstable` going forward.

PR #576, opened 2025-02-26
Splat adam7 interlacing

thirumurugan-git:
As a follow-up to [this issue](https://issues.chromium.org/issues/379305021) in Chromium, I created a corresponding PR in `image-png`.

CC: @anforowicz

### DIFF
--- a/src/adam7.rs
+++ b/src/adam7.rs
@@ -226,10 +226,38 @@ pub fn expand_pass(
         let bytes_pp = bits_pp / 8;
 
         for (bitpos, px) in bit_indices.zip(interlaced_row.chunks(bytes_pp)) {
-            for (offset, val) in px.iter().enumerate() {
-                img[bitpos / 8 + offset] = *val;
-            }
+            let start_byte = bitpos / 8;
+            splat_expand(img, interlace_info, start_byte, px, img_row_stride, bytes_pp);
         }
+    }
+}
+
+fn copy_n_bytes_m_times(dst: &mut [u8], src: &[u8], n: usize, m: usize) {
+    assert!(n * m <= dst.len(), "Destination buffer is too small!");
+
+    for i in 0..m {
+        let start = i * n;
+        let end = start + n;
+        dst[start..end].copy_from_slice(src);
+    }
+}
+
+fn splat_expand(img: &mut [u8], interlace_info: &Adam7Info, start_byte: usize, px: &[u8], img_row_stride: usize, bytes_pp: usize) {
+    // Pairs are (x_wide_copy, y_wide_copy)
+    static COPY: &[(usize, usize)] = &[
+        (8, 8),
+        (4, 8),
+        (4, 4),
+        (2, 4),
+        (2, 2),
+        (1, 2),
+        (1, 1)
+    ];
+
+    let (x_repeat, y_repeat) = COPY[interlace_info.pass as usize - 1];
+    for i in 0..y_repeat {
+        let offset = start_byte + i * img_row_stride;
+        copy_n_bytes_m_times(&mut img[offset..], px, bytes_pp, x_repeat);
     }
 }
 
--- a/src/adam7.rs
+++ b/src/adam7.rs
@@ -9,14 +9,14 @@
 use std::cmp::min;
 
 struct PassConstants {
-    line_mul: u8,
-    line_off: u8,
-    samp_mul: u8,
-    samp_off: u8,
+    line_mul: usize,
+    line_off: usize,
+    samp_mul: usize,
+    samp_off: usize,
 }
 
 impl PassConstants {
-    const fn new(line_mul: u8, line_off: u8, samp_mul: u8, samp_off: u8) -> Self {
+    const fn new(line_mul: usize, line_off: usize, samp_mul: usize, samp_off: usize) -> Self {
         Self {
             line_mul,
             line_off,
@@ -24,10 +24,10 @@ impl PassConstants {
             samp_off,
         }
     }
-    const fn x_repeat(&self) -> u8 {
+    const fn x_repeat(&self) -> usize {
         self.samp_mul - self.samp_off
     }
-    const fn y_repeat(&self) -> u8 {
+    const fn y_repeat(&self) -> usize {
         self.line_mul - self.line_off
     }
 }
@@ -188,22 +188,12 @@ fn expand_adam7_bits(
     let pass = info.pass;
     let interlaced_width = info.width as usize;
 
-    let (line_mul, line_off, samp_mul, samp_off) = {
-        // The pass values range from 1 to 7, need to adjust them to start
-        // from 0 to 6.
-        let PassConstants {
-            line_mul,
-            line_off,
-            samp_mul,
-            samp_off,
-        } = PASS_CONSTANTS[pass as usize - 1];
-        (
-            line_mul as usize,
-            line_off as usize,
-            samp_mul as usize,
-            samp_off as usize,
-        )
-    };
+    let PassConstants {
+        line_mul,
+        line_off,
+        samp_mul,
+        samp_off,
+    } = PASS_CONSTANTS[pass as usize - 1];
 
     // the equivalent line number in progressive scan
     let prog_line = line_mul * line_no + line_off;
@@ -280,7 +270,7 @@ pub fn expand_pass(
 // Should we have a separate function, or can we use an enum to determine whether
 // to use default or splat?
 //
-// |width_in_pixels| is only needed for bit splat expand, but currently it's
+// |frame_width_in_pixels| is only needed for bit splat expand, but currently it's
 // required as a parameter for both splat expand types. Would it be possible to
 // pass the width only when it's a bit splat expand? Or is there an alternative
 // way to retrieve the width without passing it explicitly?
@@ -315,7 +305,7 @@ pub fn splat_expand_pass(
     interlaced_row: &[u8],
     interlace_info: &Adam7Info,
     bits_per_pixel: u8,
-    width_in_pixels: usize,
+    frame_width_in_pixels: usize,
 ) {
     let bits_pp = bits_per_pixel as usize;
 
@@ -330,7 +320,7 @@ pub fn splat_expand_pass(
                 interlace_info,
                 bits_pp,
                 pos,
-                width_in_pixels * bits_pp,
+                frame_width_in_pixels * bits_pp,
             );
         }
     } else {
@@ -359,22 +349,19 @@ fn bit_splat_expand(
     info: &Adam7Info,
     bits_pp: usize,
     bit_pos: usize,
-    img_width_in_bits: usize,
+    frame_width_in_bits: usize,
 ) {
     assert!(
-        img_width_in_bits <= stride_in_bits,
-        "The image width should be less than or equal to the row length."
+        frame_width_in_bits <= stride_in_bits,
+        "The frame width should be less than or equal to the row length."
     );
     let height = (img.len() * 8 - bit_pos + stride_in_bits - 1) / stride_in_bits;
     let pass_const = &PASS_CONSTANTS[info.pass as usize - 1];
-    let (x_repeat, y_repeat) = (
-        pass_const.x_repeat() as usize,
-        pass_const.y_repeat() as usize,
-    );
+    let (x_repeat, y_repeat) = (pass_const.x_repeat(), pass_const.y_repeat());
     for i in 0..min(y_repeat, height) {
         let offset = bit_pos + i * stride_in_bits;
         let max_fill = min(
-            (img_width_in_bits - bit_pos % stride_in_bits) / bits_pp,
+            (frame_width_in_bits - bit_pos % stride_in_bits) / bits_pp,
             x_repeat,
         );
         copy_nbits_mtimes(img, offset, bits_pp, px, max_fill);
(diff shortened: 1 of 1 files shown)


thirumurugan-git:
@anforowicz  Please review this PR and share your suggestions on the TODOs.

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
...
+    }
+}
+
+fn copy_nbytes_mtimes(dst: &mut [u8], src: &[u8], n: usize, m: usize) {
+    assert!(n * m <= dst.len(), "Destination buffer is too small!");
+
+    for i in 0..m {
+        let start = i * n;
+        let end = start + n;
+        dst[start..end].copy_from_slice(src);
```
kornelski, line 329: `for dst in dst.chunks_exact_mut(n)` will reduce bounds checks.

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
...
+    let height = ((img.len() - start_byte) as f32 / stride as f32).ceil() as usize;
+    let (x_repeat, y_repeat) = SPLAT_EXPAND[info.pass as usize - 1];
+    for i in 0..min(y_repeat, height) {
+        let offset = start_byte + i * stride;
+        let max_fill = min((stride - start_byte % stride) / bytes_pp, x_repeat);
+        copy_nbytes_mtimes(&mut img[offset..], px, bytes_pp, max_fill);
+    }
+}
+
+fn copy_nbits_mtimes(img: &mut [u8], bit_pos: usize, bits_pp: usize, px: u8, m: usize) {
```
anforowicz, line 312: IIUC the implementation assumes that we never need to copy across **two** bytes.  So maybe the asserts below could be used to document this:

```
debug_assert_eq!(bit_pos % bits_pp, 0);
debug_assert!(bits_pp == 1 || bits_pp == 2 || bits_pp == 4);
```

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
...
+
+fn byte_splat_expand(
+    img: &mut [u8],
+    stride: usize,
+    px: &[u8],
+    info: &Adam7Info,
+    bytes_pp: usize,
+    start_byte: usize,
+) {
+    let height = ((img.len() - start_byte) as f32 / stride as f32).ceil() as usize;
```
anforowicz, line 303: Same question here: can we avoid floating-point math here?

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
...
+                interlace_info,
+                bytes_pp,
+                start_byte,
+            );
+        }
+    }
+}
+
+// TODO: Can we eliminate the duplicate code in |bit_splat_expand| and |byte_splat_expand|?
+// |stride| is bit-count here, unlike in |byte_splat_expand|, where it is byte-count.
```
anforowicz, line 277: (no action needed) Maybe this can be tried as a follow-up PR?  I am not sure if the deduplication would be better or not, but hopefully looking at a specific/actual change will help us decide.

src/adam7.rs
```diff
@@ -285,6 +382,434 @@ fn test_adam7() {
...
+    splat_expand_pass(
+        actual_img,
+        img_row_stride,
+        &[11],
+        &it.next().unwrap(),
+        bp_pixel,
+    );
+    assert_eq!(actual_img, expected_img);
+
+    // Third pass
```
anforowicz, line 525: This confused me for a moment.  I wonder if you could please expand the comment to something like:

```
/// Third pass.  (Not "second pass" because for a 4-pixel-wide image `Adam7Iterator` will skip the no-op 2nd pass.)
```

Did I get that right in the proposed comment?

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
     }
 }
 
+// TODO: Export this function in lib.rs.
+// Should we have a separate function, or can we use an enum to determine whether
+// to use default or splat?
```
anforowicz, line 243: (no action needed) FWIW I vote for a separate function - ending up with `png::expand_interlaced_row` and `png::splat_interlaced_row`, but I am curious what others think.  At any rate, we can probably discuss later, in a PR that adds a public API for this.

anforowicz (commented):
Thanks for the PR!

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
...
+// |stride| is bit-count here, unlike in |byte_splat_expand|, where it is byte-count.
+fn bit_splat_expand(
+    img: &mut [u8],
+    stride: usize,
+    px: u8,
+    info: &Adam7Info,
+    bits_pp: usize,
+    bit_pos: usize,
+) {
+    let height = ((img.len() * 8 - bit_pos) as f32 / stride as f32).ceil() as usize;
```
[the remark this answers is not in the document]
fintelia: Also, this math should likely be using `u64` rather than `usize`. On 32-bit platforms it only takes 256 MB worth of bits to overflow a `usize`.

Another thing you might want to look at are the provided methods on the builtin integer types. In particular, [`div_ceil`](https://doc.rust-lang.org/std/primitive.u64.html#method.div_ceil) might be helpful here.

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
     }
 }
 
+// TODO: Export this function in lib.rs.
+// Should we have a separate function, or can we use an enum to determine whether
+// to use default or splat?
+pub fn splat_expand_pass(
```
fintelia, line 244: It would be good to add a doc comment here explaining what "splat expand" means. I understand from the linked chromium issue, but having it right in the code will help for future reference.

kornelski:
I think filling the image pixels completely is the expected behavior, so it would be nice to make this easy/default, but I'm not sure how such thing could be integrated with the library API.

Doing this in `expand_interlaced_row` would needlessly overdraw pixels when decoding multiple passes at once. So even though standalone function doesn't feel like the most elegant solution to me, I don't have a suggestion for anything better.

src/adam7.rs
thirumurugan-git: sure

src/adam7.rs
```diff
@@ -233,6 +238,98 @@ pub fn expand_pass(
...
+                img,
+                img_row_stride,
+                px,
+                interlace_info,
+                bytes_pp,
+                start_byte,
+            );
+        }
+    }
+}
```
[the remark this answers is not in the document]
thirumurugan-git: Keeping it as is for now.

src/adam7.rs
thirumurugan-git: Thank you for the update. I've added the doc comment.

src/adam7.rs
thirumurugan-git: Considering @kornelski [comment](https://github.com/image-rs/image-png/pull/576#issuecomment-2686303182), should we proceed with separate APIs and have Chromium use this new one? Once we verify that the new API works without functional issues, can we then enable splat filling as the default in the future?

src/adam7.rs
```diff
@@ -6,6 +6,42 @@
 /// applies to a decoded row.
 ///
 /// See also [Reader.next_interlaced_row](crate::decoder::Reader::next_interlaced_row).
+use std::cmp::min;
+
+struct PassConstants {
+    line_mul: u8,
+    line_off: u8,
+    samp_mul: u8,
+    samp_off: u8,
```
anforowicz, line 15: nit: Could you consider changing `u8` to `usize`?  This would help a tiny bit in another place below - there would be no need for `as usize` in `fn expand_adam7_bits`.

src/adam7.rs
```diff
@@ -6,6 +6,42 @@
...
+impl PassConstants {
+    const fn new(line_mul: u8, line_off: u8, samp_mul: u8, samp_off: u8) -> Self {
+        Self {
+            line_mul,
+            line_off,
+            samp_mul,
+            samp_off,
+        }
+    }
+    const fn x_repeat(&self) -> u8 {
```
anforowicz, line 27: nit: Could you please change the return value to `usize`?  (same for `fn y_repeat`)

* After changing the type of fields it shouldn't require changing the body of `fn x_repeat`
* It would avoid having to cast `as usize` in the only caller of `x_repeat()`.

src/adam7.rs
```diff
@@ -233,6 +276,151 @@ pub fn expand_pass(
...
+//   3 3 3 3 3 3 3 3
+//   3 3 3 3 3 3 3 3
+//   3 3 3 3 3 3 3 3
+pub fn splat_expand_pass(
+    img: &mut [u8],
+    img_row_stride: usize,
+    interlaced_row: &[u8],
+    interlace_info: &Adam7Info,
+    bits_per_pixel: u8,
+    width_in_pixels: usize,
```
anforowicz, line 318: nit: Could you please change `width_in_pixels` to `frame_width_in_pixels`?

src/adam7.rs
```diff
@@ -233,6 +276,151 @@ pub fn expand_pass(
...
+// TODO: Eliminate the duplicate code in |bit_splat_expand| and
+// |byte_splat_expand|, if possible.
+fn bit_splat_expand(
+    img: &mut [u8],
+    stride_in_bits: usize,
+    px: u8,
+    info: &Adam7Info,
+    bits_pp: usize,
+    bit_pos: usize,
+    img_width_in_bits: usize,
```
anforowicz, line 362: Can you please change `img_width_in_bits` to `frame_width_in_bits`?

anforowicz (approved):

anforowicz:
@kornelski and/or @fintelia - can you take another look please?

kornelski (approved):

src/adam7.rs
```diff
@@ -233,6 +266,145 @@ pub fn expand_pass(
...
+                frame_width_in_pixels * bits_pp,
+            );
+        }
+    } else {
+        let bytes_pp = bits_pp / 8;
+
+        for (bitpos, px) in bit_indices.zip(interlaced_row.chunks(bytes_pp)) {
+            let start_byte = bitpos / 8;
+            byte_splat_expand(
+                img,
```
fintelia, line 332: Would it make sense to call this with `&mut img[start_byte...]` and then not pass `start_byte` as a separate argument?

src/adam7.rs
```diff
@@ -233,6 +266,145 @@ pub fn expand_pass(
...
+
+fn byte_splat_expand(
+    img: &mut [u8],
+    stride: usize,
+    px: &[u8],
+    info: &Adam7Info,
+    bytes_pp: usize,
+    start_byte: usize,
+) {
+    let height = (img.len() - start_byte + stride - 1) / stride;
```
fintelia, line 379: I think this can use the [`div_ceil`](https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil) method

src/adam7.rs
```diff
@@ -6,6 +6,42 @@
 /// applies to a decoded row.
 ///
 /// See also [Reader.next_interlaced_row](crate::decoder::Reader::next_interlaced_row).
+use std::cmp::min;
+
+struct PassConstants {
+    line_mul: usize,
+    line_off: usize,
+    samp_mul: usize,
+    samp_off: usize,
```
fintelia, line 15: Rust convention is to mostly avoid abbreviations in variable names. So for instance, `sample_offset` rather than `samp_off`

src/adam7.rs
```diff
@@ -78,16 +114,21 @@ impl Adam7Iterator {
...
-        let (line_width, lines) = match self.current_pass {
-            1 => (w / 8.0, h / 8.0),
-            2 => ((w - 4.0) / 8.0, h / 8.0),
-            3 => (w / 4.0, (h - 4.0) / 8.0),
-            4 => ((w - 2.0) / 4.0, h / 4.0),
-            5 => (w / 2.0, (h - 2.0) / 4.0),
-            6 => ((w - 1.0) / 2.0, h / 2.0),
-            7 => (w, (h - 1.0) / 2.0),
-            _ => unreachable!(),
+        let (line_mul, line_off, samp_mul, samp_off) = {
```
fintelia, line 117: I think this whole function can be simplified a bunch. If I'm not missing something, should be a matter of...

```rust
let pass = PASS_CONSTANTS[self.current_pass as usize - 1];

self.line_width = (self.width - pass.samp_off).div_ceil(pass.samp_mul);
self.lines = (self.width - pass.line_off).div_ceil(pass.line_mul);
self.line = 0;
```

(And making the constants `u32` rather than `usize`)

thirumurugan-git:
@fintelia I've made the necessary changes for the splat expand pass. However, updating the parameter types as discussed, like changing `u32` for image width/height and using `u64` for byte handling, would require refactoring the existing codebase, including tests. Since that would introduce significant changes beyond the scope of this feature, I’d prefer not to include that refactoring here. 

Would it make sense to create a separate issue to track the type changes and have further discussions around using `u32::div_ceil` or `u64::div_ceil` there? Let me know your thoughts.

fintelia:
I'm OK with deferring changes to other parts of the codebase, but I do want to make sure that this PR doesn't introduce _new_ integer overflow bugs. At a minimum, we absolutely cannot do `img.len() * 8` which will overflow even on an RGB8 16K x 16K image (on 32-bit platforms).

thirumurugan-git:
@fintelia tested this(`multibyte_expand_pass_test_helper(1, 256000000, 32)`) code and confirmed that a buffer overflow does occur. However, the overflow originates in the existing code(`let line_start = prog_line * row_stride_in_bytes * 8;` method: `expand_adam7_bits`) before it reaches my changes. Would it be better to address the existing overflow in a separate PR and then proceed with fixing and merging this PR? Or would it be preferable to fix both overflows in the current PR and merge it together? What are your thoughts?

fintelia:
Small PRs are much preferred!

thirumurugan-git:
@197g Thank you for taking up my PR. I’ve been caught up with my own schedule, so I haven’t had the chance to review the changes yet. I’ll go through them as soon as I get some time. Also, is there any Discord or Telegram community for `image-rs` that I can join?

197g:
@thirumurugan-git There's a matrix space: [#image-rs_image:gitter.im](https://matrix.to/#/#image-rs_image:gitter.im)

PR #577, opened 2025-02-28
Release 0.18.0-rc

fintelia:
One open question is whether we should take this chance to remove `Decoder::{next_row, next_interlaced_row}` in favor of `Decoder::read_row`  (added in #493)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,18 @@
+## 0.18.0
+
+### API Breaking Changes
+
+* Removed deprecated `Info::encode` method.
+* Improved the compression settings API.
+* `Decoder` now requires a reader that implements `Seek` and `BufRead` traits.
+* Bump bitflags dependency to 2.0.
+
+### Other additions
+
+* Added `Reader::read_row` method.
+* Add support for parsing eXIf chunk.
+* Treat most auxiliary chunk errors as benign.
+
 ## 0.17.16
 
 * Make gAMA and cHRM fallback optional for sRGB ([#547])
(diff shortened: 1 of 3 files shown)


anforowicz:
My thoughts on removing `Decoder::{next_row, next_interlaced_row}`:

* I think this shouldn't block the 0.18 release
* If we want to remove those APIs, then I think the first step would be to mark them as [`#[deprecated]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute).  I think it doesn't matter that much if we do it in 0.18.0 vs 0.18.1 or later.  And then we can remove them in 0.19.
* I don't have a strong opinion about desirability of removing those APIs:
    - The old APIs are sometimes convenient (I guess mostly because buffer management on Rust side is easier than buffer management on C++ side).  Current Chromium implementation uses those old APIs (see [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/experimental/rust_png/ffi/FFI.rs;l=600;drc=4ff2e3a8fe7e08f1e8baf3f56544bf2445d9f5c1);  I now realize that the TODO/bug link should be updated), although using the new API would force an improvement for some scenarios (avoiding an extra memcpy [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/experimental/rust_png/decoder/impl/SkPngRustCodec.cpp;l=562-572;drc=58fb75d86a0ad2642beec2d6c16b1e6c008e33cd))
    - On the other hand, the old APIs are kind of redundant (i.e. removing them will result in smaller, simpler-to-understand API)

fintelia:
Published! If all looks good, hoping to do a full 0.18 release in 1-2 weeks

PR #585, opened 2025-05-28
Stop depending on `std::simd` / using `#[feature(portable_simd)]`.

anforowicz:
PTAL?


197g (commented):
Is okay with me, as previously discussed and now measured, with the nit resolved. It may be an interesting data point for the `std::simd` experiment upstream.

src/filter.rs
```diff
@@ -2,223 +2,6 @@ use core::convert::TryInto;
...
 
-/// SIMD helpers for `fn unfilter`
-///
-/// TODO(https://github.com/rust-lang/rust/issues/86656): Stop gating this module behind the
-/// "unstable" feature of the `png` crate.  This should be possible once the "portable_simd"
-/// feature of Rust gets stabilized.
-///
-/// This is only known to help on x86, with no change measured on most benchmarks on ARM,
-/// and even severely regressing some of them.
-/// So despite the code being portable, we only enable this for x86.
```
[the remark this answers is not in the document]
anforowicz: Good catch.  Done.  (I've rewritten+expanded these comments a little bit.  I wasn't sure where to put them, so I just appended them to the big comment bloc inside `fn unfilter` - not perfect, but probably good enough.)

Cargo.toml
```diff
@@ -38,8 +38,6 @@ rand = "0.8.4"
 term = "1.0.1"
 
 [features]
-# Use nightly-only features for a minor performance boost in PNG decoding
-unstable = ["crc32fast/nightly"]
```
fintelia, line 42: I think it might be worth keeping this feature around to enable the nightly feature on crc32fast (which enables an optimized implementation for aarch64)

anforowicz: Good idea.  Done.

(I also retained this crate feature in `README.md` and `.github/workflows/rust.yml`;  OTOH, I felt that it's still okay to stop referring to this feature in `benches/unfilter.rs`.)

PR #589, opened 2025-06-02
Shared stream

fintelia:
This PR builds on #588 but goes much farther, and results in even better performance. 

Right now, decoding involves copying data between two "streams". The `ZlibStream` object contains an internal buffer of uncompressed data that it uses for zlib decoding lookback (because `fdeflate` requires that the 32 KB of previously uncompressed data is provided for subsequent decompress calls). And then there's a second stream of data managed by the `UnfilteringBuffer` object.

Rather than copying between these two objects, it is better to directly uncompress data into the `UnfilteringBuffer` and make sure that the trailing 32 KB don't get unfiltered. (TODO: clearer enforcement of this invariant)

To be able to uncompress directly in the unfiltering buffer, this PR changes the `StreamingDecoder` to no longer handle decompression. It now just signals which bytes of the input correspond to IDAT / fDAT contents, and leaves it to the caller to handle the decompression portion. This does represent a breaking API change but overall I think I like the interface better this way.


src/decoder/unfiltering_buffer.rs
```diff
@@ -54,7 +57,7 @@ impl UnfilteringBuffer {
 
     /// Returns how many bytes of the current row are present in the buffer.
     pub fn curr_row_len(&self) -> usize {
-        self.data_stream.len() - self.current_start
+        self.unfilled_start - self.current_start
     }
 
     /// Returns a `&mut Vec<u8>` suitable for passing to
```
fintelia, line 63: TODO: Update this doc comment

src/decoder/unfiltering_buffer.rs
```diff
@@ -65,19 +68,30 @@ impl UnfilteringBuffer {
...
-        &mut self.data_stream
+        if self.unfilled_start + 8 * 1024 > self.data_stream.len() {
+            self.data_stream.resize(self.unfilled_start + 8 * 1024, 0);
+        }
+
+        let end = self.data_stream.len().min(self.unfilled_start + 64 * 1024);
+        (&mut self.data_stream[..end], self.unfilled_start)
+    }
+
+    pub fn marked_valid(&mut self, bytes: usize) {
```
fintelia, line 92: Better function name?

src/decoder/stream.rs
```diff
@@ -1668,8 +1636,8 @@ impl StreamingDecoder {
                 // TODO: Calculate **exact** IDAT size for interlaced images.
                 raw_row_len = raw_row_len.saturating_mul(2);
             }
-            self.inflater
-                .set_max_total_output((height as usize).saturating_mul(raw_row_len));
+            // self.inflater
+            //     .set_max_total_output((height as usize).saturating_mul(raw_row_len));
```
fintelia, line 1640: Should add back max total output calculation and use it to avoid overallocating for unfiltering buffer.

src/decoder/stream.rs
```diff
@@ -777,16 +749,16 @@ impl StreamingDecoder {
             ImageData(type_str) => {
                 debug_assert!(type_str == IDAT || type_str == chunk::fdAT);
                 let len = std::cmp::min(buf.len(), self.current_chunk.remaining as usize);
-                let buf = &buf[..len];
-                let consumed = self.inflater.decompress(buf, image_data)?;
-                self.current_chunk.crc.update(&buf[..consumed]);
-                self.current_chunk.remaining -= consumed as u32;
+                if !self.decode_options.ignore_crc {
```
fintelia, line 752: This check should be added regardless of this PR. We waste time computing image data CRCs even if CRC checks are disabled...

src/decoder/stream.rs
```diff
@@ -449,7 +447,7 @@ impl From<TextDecodingError> for DecodingError {
 /// Decoder configuration options
 #[derive(Clone)]
 pub struct DecodeOptions {
-    ignore_adler32: bool,
+    pub(crate) ignore_adler32: bool,
```
fintelia, line 450: This is a bit ugly, might want to think about a better way to handle disable adler32 checksums now.

197g:
I'm not sure if I agree with moving the `image_data` argument as a cleaner API. Changing the argument to be an `UnfilterBuffer` instead of a `Vec<u8>` is surely the right move but why is it necessary to move the inflater in that case? When passing our own type we might as well introduce the right invariants to communicate and preserve the buffer state. In the largest benchmark we've been shifting 600MB of data back in `copy_within` calls!

The dance around `decompress.pending_idat_byte` is really strange in my opinion.

Another thing I've been exploring yesterday is that the performance trace shows a surprising amount of time being spent in `update/next_step` which does not include unfiltering. The state machine implementation is the culprit. We're stutter-stepping the `ReadDecoder` state-by-state and chunk-by-chunk boundary (with lots of branches and unnecessary intermediate writes to `state`) even if the layer passed a very large buffer to decode. The way to fix this would be to step over multiple states at once but becomes more complicated if we're handling all the inflate calls above where we _must_ now break on each `decompress.pending_idat_bytes == 0` and then go the call stack all the way down again.

fintelia:
I initially started with trying to pass `UnfilteringBuffer`  into `StreamingDecoder::update`. But there were a few things that led me to thinking that this version is a cleaner design:
* We currently have a bunch of places that pass a throwaway `Vec` into `update` because they have no way to signal that they don't expect image data. Changing those to throwaway `UnfilteringBuffer`s seemed ugly.
* There is currently no way to indicate that you want to discard image data rather than uncompress it [as `ReadDecoder` sometimes wants to do](https://github.com/image-rs/image-png/blob/6ab0a44fcb09eff6b7c2a210d800f0220122c4e7/src/decoder/read_decoder.rs#L83). This design gives that capability basically for free.
* `StreamingDecoder` is a public API so we'd have to make`UnfilteringBuffer` public as well. But the details of what API to expose weren't obvious to me. And getting that API wrong would make future optimizations more difficult.
* As I thought about it more, having a public API that decompresses the pixel data but doesn't unfilter it seemed a little weird. 

If we accept that `StreamingDecoder` should return compressed data rather than uncompressing it, then the current design follows naturally: a single call to `update` can now return many MB or even several GB of compressed data, so having the caller process it a bit at a time and track the remaining amount in `pending_idat_bytes` makes sense.

I do want to take some measurements with [`innumerable`](https://docs.rs/innumerable) to get a sense of how this impacts the number of `update` calls and how much stutter-stepping is happening in the streaming decoder.

fintelia:
Closing in favor of #590

Issue #599, opened 2025-06-11
Fix Integer Overflow Issue in Adam7 Interlacing


thirumurugan-git:
As discussed in PR #576, the Adam7 interlacing implementation contains integer overflow bugs that need to be fixed.

PR #600, opened 2025-06-11
Fix Integer Overflow in Adam7 Interlacing

thirumurugan-git:
Issue - #599

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ crc32fast = "1.2.0"
 fdeflate = "0.3.3"
 flate2 = "1.0.35"
 miniz_oxide = { version = "0.8", features = ["simd"] }
+rand = "0.8.4"
 
 [dev-dependencies]
 approx = "0.5.1"
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -86,10 +86,9 @@ jobs:
   test_all_32:
     runs-on: ubuntu-latest
     container:
-      image: i386/debian
+      image: i386/rust
     steps:
     - uses: actions/checkout@v4
-    - run: rustup default stable
     - name: test
       run: >
         cargo test -v adam7::test_32_bit_machine
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -88,7 +88,6 @@ jobs:
     container:
       image: i386/rust
     steps:
-    - name: install node
     - run: apt update && apt install nodejs -y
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ crc32fast = "1.2.0"
 fdeflate = "0.3.3"
 flate2 = "1.0.35"
 miniz_oxide = { version = "0.8", features = ["simd"] }
+rand = "0.8.4"
```
fintelia, line 29: This should be a dev-dependency

197g:
As a note, I'd very much appreciate seeing `wasm32-wasip1` or `wasm32-wasip2` tested anyways. So if we can make that the 32-bit test configuration we get two birds with one stone.

Cargo.toml
Shnatsel: I'd prefer not to pull in the entirety of `rand`'s dependency tree. [`nanorand`](https://crates.io/crates/nanorand) with `default-features = false` seems sufficient.

Or even use this snippet and avoid additional dependencies completely:

```rust
/// A simple pseudo-random number generator (LCG).
pub struct SimpleRng {
    state: u32,
}

impl SimpleRng {
    /// Creates a new `SimpleRng` with a given seed.
    pub fn new(seed: u32) -> Self {
        SimpleRng { state: seed }
    }

    /// Generates the next random `u8` value.
    pub fn next_u8(&mut self) -> u8 {
        // LCG parameters: these are common and produce reasonable results.
        self.state = self.state.wrapping_mul(1664525).wrapping_add(1013904223);
        // Return the upper 8 bits of the state as a `u8`.
        (self.state >> 24) as u8
    }
}
```

197g:
Some pointers:

1. Install wasmtime
2. ```
   # .cargo/config
   [target.wasm32-wasip1]
   runner = ["wasmtime", "--dir", ".::"]
   ```
3. Remove `term` from the dev-dependency tree, it's just for one example. It seems its `home` dependency is confused well with wasip1 having a std library but no defined system environment and doesn't compile.
4. Remove `rayon` from criterion in `dev-dependencies` as we have no threads:
   ```
   criterion = { version = "0.4.0", default-features = false, features = ["cargo_bench_support"] }
   [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
   criterion = "0.4.0"
   ```

197g:
This was superseeded by #606, thank you for the test case sketch..

PR #601, opened 2025-06-14
Allow repeatedly calling `flush` on a `StreamWriter`.

anforowicz:
The new unit test added in this PR would fail without production code changes - calling `flush` repeatedly wasn't supported before.  Before this PR `flush` called into (consuming, only-one-call-possible) `fdeflate::Compressor::finish`.  After this PR:

* `flush` is a no-op for `fdeflate` back-end (because right now `fdeflate::Compressor` doesn't expose a public API for calling the `flush` on the underlying `Write` object).
* `fdeflate::Compressor::finish` call was moved into `png::StreamingWriter::finish`.

The new test would still fail at this point, because `png::StreamingWriter::drop` wouldn't finish and flush all the data.  This PR fixes this by `finish`ing instead of `flush`ing in `fn drop`.

--------------

This PR also opportunistically added a call from `png::StreamingWriter::finish` to `flate2::write::ZlibEncoder::finish`, because this seems like the right thing to do (?).  Ideally we would also add a unit test, but I wasn't able to figure out how to write such test (I probably didn't spend enough time on this...).

This PR does *not* add a call to `finish` from `png::StreamingWriter::finish`'s `Wrapper::Chunk` `match` arm, because:

* We can't consume `wrt.writer` field by value
* Dropping `wrt.writer` field (implicitly, but by the end of `png::StreamingWriter::finish`) should do the right thing (i.e. call `write_iend` and `flush`).

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1599,16 +1599,33 @@ impl<'a, W: Write> StreamWriter<'a, W> {
     /// [`Writer`], then it will also do a check on their correctness. Differently from
     /// [`Writer::finish`], this just `flush`es, returns error if some data is abandoned.
     pub fn finish(mut self) -> Result<()> {
+        self.finish_mut()
+    }
+
+    /// Internal helper that can be called both from `fn finish(mut self)`
+    /// and from `fn drop(&mut self)`.
+    fn finish_mut(&mut self) -> Result<()> {
         if self.to_write > 0 {
             let err = FormatErrorKind::MissingData(self.to_write).into();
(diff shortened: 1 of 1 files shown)


fintelia:
Fun fact: Because flate2's `flush` method does a [sync flush](https://www.bolet.org/~pornin/deflate-flush.html) while the codepath for fdeflate doesn't do anything, the test for "high" compression produces a PNG that's nearly twice as large as the test for "fastest" compression does.

fintelia (approved):

PR #607, opened 2025-06-23
Ignore `acTL` chunks with `num_frames == 0`.

anforowicz:
PTAL?

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -482,7 +482,7 @@ impl<R: BufRead + Seek> Reader<R> {
     /// Returns the next processed row of the image (discarding `InterlaceInfo`).
     ///
     /// See also [`Reader.read_row`], which reads into a caller-provided buffer.
-    pub fn next_row(&mut self) -> Result<Option<Row>, DecodingError> {
+    pub fn next_row(&mut self) -> Result<Option<Row<'_>>, DecodingError> {
         self.next_interlaced_row()
             .map(|v| v.map(|v| Row { data: v.data }))
     }
@@ -490,7 +490,7 @@ impl<R: BufRead + Seek> Reader<R> {
     /// Returns the next processed row of the image.
     ///
     /// See also [`Reader.read_row`], which reads into a caller-provided buffer.
-    pub fn next_interlaced_row(&mut self) -> Result<Option<InterlacedRow>, DecodingError> {
+    pub fn next_interlaced_row(&mut self) -> Result<Option<InterlacedRow<'_>>, DecodingError> {
         let mut output_buffer = mem::take(&mut self.scratch_buffer);
         output_buffer.resize(self.output_line_size(self.info().width), 0u8);
         let result = self.read_row(&mut output_buffer);
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -1139,6 +1139,11 @@ impl StreamingDecoder {
                 num_frames: buf.read_be()?,
                 num_plays: buf.read_be()?,
             };
+            // The spec says that "0 is not a valid value" for `num_frames`.
+            // So let's ignore such malformed `acTL` chunks.
+            if actl.num_frames == 0 {
+                return Ok(Decoded::Nothing);
+            }
             self.info.as_mut().unwrap().animation_control = Some(actl);
             Ok(Decoded::AnimationControl(actl))
         }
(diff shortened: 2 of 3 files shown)


src/decoder/mod.rs
```diff
@@ -482,15 +482,17 @@ impl<R: BufRead + Seek> Reader<R> {
     /// Returns the next processed row of the image (discarding `InterlaceInfo`).
     ///
     /// See also [`Reader.read_row`], which reads into a caller-provided buffer.
-    pub fn next_row(&mut self) -> Result<Option<Row>, DecodingError> {
+    pub fn next_row<'a>(&'a mut self) -> Result<Option<Row<'a>>, DecodingError> {
```
197g: Nit: If these are with regards to (clippy) lints about hidden unexpected lifetimes, the more common option should be using the unnamed lifetime where possible.

```suggestion
    pub fn next_row(&mut self) -> Result<Option<Row<'_>>, DecodingError> {
```

There is a subtle difference in function signature with an explicit lifetime generic parameter. (And slightly more pleasant to read).

197g (commented):
Main code itself LGTM, nits as per comment.

Conceptually strictness vs. practicality is always a balance. I think we can make the argument that these being Ancillary Chunks explicitly allows ignoring them per spec; only then this is not consistent with our treatment of `fcTL` (e.g. BlendOp).

We don't need to resolve all these choices the same way but I'd like to understand what the motivation is coming from, here, so that we can bulid a more coherent policy for other instances of ignorable lint-level file problems we come across.

If libpng compatibility, LGTM, and we open a separate issue for aligning other non-fatal error cases.

src/decoder/mod.rs
anforowicz: Ok - done in all 4 cases.

anforowicz:
> Conceptually strictness vs. practicality is always a balance. I think we can make the argument that these being Ancillary Chunks explicitly allows ignoring them per spec; only then this is not consistent with our treatment of `fcTL` (e.g. BlendOp).
> 
> We don't need to resolve all these choices the same way but I'd like to understand what the motivation is coming from, here, so that we can bulid a more coherent policy for other instances of ignorable lint-level file problems we come across.
> 
> If libpng compatibility, LGTM, and we open a separate issue for aligning other non-fatal error cases.

I am not sure if I have a good, cohesive story to tell wrt reporting errors as fatal-vs-benign.  Let me try to share some notes and see if that helps:

* `libpng` compatibility is not a concern here, because `libpng` doesn't handle APNG chunks.
* I think for `num_frames == 0` it would be equally valid to report this as a fatal error, because IIUC previously there was either an `assert!`-related crash, or a fatal error (`ParameterErrorKind::PolledAfterEndOfImage`).
* I don't have a good justification for going with a benign error.  Trying to avoid introducing a new variant of `FormatErrorInner` is probably one reason.
* The spec [recommends](https://www.w3.org/TR/png-3/#13Decoders.Errors:~:text=It%20is%20strongly%20recommended%20that%20when%20any%20error%20is%20encountered%20decoders%20should%20discard%20all%20subsequent%20frames%2C%20stop%20the%20animation%2C%20and%20revert%20to%20displaying%20the%20static%20image) treating `fcTL` and `fdAT` errors as a trigger for falling back to a non-animated `IDAT` image.  This is why the `png` crate can't implement the fallback itself - it has to be done in a higher layer.  OTOH I think this doesn't apply to errors in `acTL` chunk, because this chunk appears before `IDAT`.  And ignoring the `acTL` chunk effectively means reverting to the static image.

So, I think the behavior in the PR is okay.  OTOH, I understand that I don't have a strong justification for this.  If you feel strongly, then I can switch to reporting a fatal error.

kornelski:
Decoder strictness is a prisoner's dilemma. We can't win this.

Once some non-conformance slips through, we'll get users reporting *"this image works in {my browser|my image viewer|my graphics software}, and not in your thing, so your thing is bad"*. If we try to explain why, all users are going to hear is that we *could* make the image work, we just choose not to.

We could have a optional opt-in strict mode (*not* by default), or report warnings about spec violations while still decoding the image. But in general, consumers of the library have no use for it. Users just want their images.

197g (approved):
> Anomalous situations other than syntax errors shall be treated as follows:
> 1. Encountering an unknown ancillary chunk is never an error. The chunk can simply be ignored.
> […]
>
> It is strongly recommended that when any error is encountered decoders should discard all subsequent frames, stop the animation, and revert to displaying the static image.
>
> -- https://www.w3.org/TR/png-3/#13Decoders.Errors

Thank you for the reminder, that is sufficient motivation for me. (Also a good basis for making judgment calls on other error sites).

Issue #603, opened 2025-06-18
`output_buffer_size()` calculation may overflow


Shnatsel:
This code uses completely unchecked arithmetic and may overflow on 32-bit systems: 

https://github.com/image-rs/image-png/blob/9294c26dc3ca7622f791e880810b575193fb6c29/src/decoder/mod.rs#L630-L636

This also causes issues when trying to enforce memory limits: https://github.com/image-rs/image/issues/2111

Now that we're shipping a breaking release, it is an excellent time to fix this and make the function return an `Option` or something along those lines.

Shnatsel:
I've taken a stab at converting this to `image`-like API that saturates to `u64::MAX` on overflow. It's harder than it looked on the surface: those calculations are used extensively throughout the crate, so changing the type causes an avalanche of type changes everywhere else, and I'm still not done with them:
https://github.com/image-rs/image-png/compare/master...Shnatsel:image-png:fix-dimensions-overflow?expand=1

Worse, those `u64`s have to be converted into `usize` for allocating buffers eventually, so we have to add a checked conversion there anyway; which is why it was convenient to have them as `usize`. And users of public API would have to know convert the sizes correctly with `usize::try_from` and not `as usize`, so we're creating a potential footgun with reporting these parameters as a `u64`.

At this point I think returning a `Result<usize, _>` instead of `u64` might be a better idea from the API standpoint.

fintelia:
I think it is worth stepping back and thinking about what it means for these values to overflow. The maximum slice size in Rust is `isize::MAX` which is approximately half of `usize::MAX`. Thus there's no way to decode a PNG with `usize::MAX` bytes into memory any more than there's a way to decode one exceeding that many bytes. And while there's no theoretical reason that a slice of `u64::MAX` bytes isn't possible, no system today is anywhere close to supporting that. Even if one did, the time taken to decode an image that large would be measured in decades if not centuries.

If we encounter an image that doesn't fit into `usize::MAX` bytes, our options are either to fail immediately when seeing the header or to continue decoding in a degraded mode. We won't be able to execute `Reader::next_frame`, but we could still give access to the metadata and maybe the row-by row decoding functionality.

The encoding path has similar options. There's a good case to be made that we should outright fail when trying to encode such images or alternatively we could allow only steaming encode.

Shnatsel:
[...]

PR #608, opened 2025-06-24
Check output size overflow when reading frames

197g:
Closes: #603 

I'm not entirely certain this fix is complete but it is a good shot. Crucially, we only check the condition at the start of reading a frame. This ensure the viral effect is minimal, assuming we only require the buffer size to be address space representable when *actually* writing into the output buffer. For instance, `StreamReader` is unaffected. This is by design as any reader path not interacting with image data can proceed working regardless. (Since the `UnfilterBuf` argument is optional this is a blessed use case by our API design).

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -402,11 +402,17 @@ impl<R: BufRead + Seek> Reader<R> {
             self.read_until_image_data()?;
         }
 
-        if buf.len() < self.output_buffer_size() {
+        // Note that we only check if the buffer size calculation holds in a call to decoding the
+        // frame. Consequently, we can represent the `Info` and frameless decoding even when the
+        // target architecture's address space is too small for a frame. However reading the actual
+        if self
+            .checked_output_buffer_size()
+            .map_or(true, |size| buf.len() < size)
+        {
             return Err(DecodingError::Parameter(
                 ParameterErrorKind::ImageBufferSize {
-                    expected: buf.len(),
-                    actual: self.output_buffer_size(),
+                    expected: self.output_buffer_size(),
+                    actual: buf.len(),
                 }
                 .into(),
             ));
@@ -633,6 +639,21 @@ impl<R: BufRead + Seek> Reader<R> {
         let (width, height) = self.info().size();
         let size = self.output_line_size(width);
         size * height as usize
+    }
+
+    /// Return the number of bytes required to hold a deinterlaced image frame, if it fits into the
+    /// memory space of the machine.
+    pub fn checked_output_buffer_size(&self) -> Option<usize> {
+        let (width, height) = self.info().size();
+        let (color, depth) = self.output_color_type();
+        // The subtraction should always work, but we do this for consistency. Also note that by
+        // calling `checked_raw_row_length` the row buffer is guaranteed to work whereas if we
+        // ran other function that didn't include the filter byte that could later fail on an image
+        // that is `1xN`...
+        let size = color.checked_raw_row_length(depth, width)?.checked_sub(1)?;
+        let height = usize::try_from(height).ok()?;
+        size.checked_mul(height)
+            .filter(|&n| n < isize::MAX as usize)
     }
 
     /// Returns the number of bytes required to hold a deinterlaced row.
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -365,7 +365,7 @@ impl<R: BufRead + Seek> Reader<R> {
         self.unfiltering_buffer.reset_all();
 
         // Allocate output buffer.
-        let buflen = self.output_line_size(self.subframe.width);
+        let buflen = self.unguarded_output_line_size(self.subframe.width);
         self.decoder.reserve_bytes(buflen)?;
 
         Ok(())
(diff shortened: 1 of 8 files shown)


src/decoder/mod.rs
```diff
@@ -635,6 +641,21 @@ impl<R: BufRead + Seek> Reader<R> {
...
+        let (width, height) = self.info().size();
+        let (color, depth) = self.output_color_type();
+        // The subtraction should always work, but we do this for consistency. Also note that by
+        // calling `checked_raw_row_length` the row buffer is guaranteed to work whereas if we
+        // ran other function that didn't include the filter byte that could later fail on an image
+        // that is `1xN`...
+        let size = color.checked_raw_row_length(depth, width)?.checked_sub(1)?;
+        let height = usize::try_from(height).ok()?;
+        size.checked_mul(height)
+            .filter(|&n| n < isize::MAX as usize)
```
Shnatsel: I find this line really confusing. Could you express it as an `if let` or a `match` for clarity?

src/decoder/mod.rs
```diff
@@ -635,6 +641,21 @@ impl<R: BufRead + Seek> Reader<R> {
         size * height as usize
     }
 
+    /// Return the number of bytes required to hold a deinterlaced image frame, if it fits into the
+    /// memory space of the machine.
+    pub fn checked_output_buffer_size(&self) -> Option<usize> {
```
197g: The function could also return an error, LimitExceeded is arguably correct. @Shnatsel any preference on this? (Aside: I'll do the same transform to output_line_size afterall, might as well strive for consistency in the API break).

Shnatsel: `DecodingError` is too broad because there is only one failure condition and making a new error type just for this doesn't seem to be worth the trouble.

So I think it's best to return an `Option`. It is consistent with `checked_mul` from std, too.

197g: Alright, that seems to make sense. Then I'll use the `LimitsExceeded` only for diagnostics in the usage sites.

src/decoder/mod.rs
197g: I opted for `bool::then_some`, is that alright to read?

fintelia:
I've been increasingly thinking we should just return an error when creating the `Reader` if the output buffer size would exceed `usize::MAX`.

Like realistically when is that case ever going to come up for a real image that wasn't crafted by a fuzzer / someone trying to crash the library?  It pretty much comes down to a 32-bit system -- which are very rare these days -- trying to decode a 500+ megapixel image. On a 64-bit system, a PNG would have to have a compressed size of 16 petabytes(!) in order to hit that limit.

Rather than creating an API that says to users "this is a case you should worry about" we should instead just return an error at the start and then not think about it throughout the rest of the code-base.

src/decoder/mod.rs
Shnatsel: Yeah that's clear enough. Thanks!

197g:
Images exceeding `2GiB` bytes of image data are not _that_ unrealistic to occur in non-Web environments, an RgbA square of 16 * 1024 edge length is enough which is easily in reach of rendered maps of the Mandelbrot set. I think it's quite useful to be able to read the `Info` of such images or even iterate over chunks _without_ the decoded image data. Blowing through the limit of `line_size` is .. less convincingly useful but that is a matter of consistency then.

fintelia:
What 32-bit systems are around that would want to work with that kind of data?

197g:
Arduino boards are still 32-bit. ARM instruction sets before ARMv9-A (2021!) are allowed to be implemented without aarch64 support so I expect a bunch of cheap industrial SBC sets to be 32-bit as well. Plus wasm32 is a thing... so. I don't see the upper-bound estimate of 32-bit prevalence being _reliable_ enough to drop support for it, particular over such a somewhat small maintenance issue.

fintelia:
I certainly don't think we should drop support for 32-bit. But I also don't want to devote too much time and energy to edge cases that might not even impact anybody.

* An easy option is to say if you are running on 32-bit **and** try to load an image whose uncompressed size exceeds 4 GB, then you'd get an error immediately after reading the metadata. That would guarantee we'd hit no overflow bugs later, and if anyone complains, we could always change the implementation later. But if no one hits this case then we'd avoid having to think about it.
* Another idea is to return an error if the output size exceeds 2^64 bytes and make `output_buffer_size` saturate to `usize::MAX`. That guarantees correct behavior on 64-bit, and you only have to think about the gotcha of output_buffer_size being wrong if you target 32-bit systems and deal with giant images. 
* We could also always saturate to `usize::MAX` or switch the signature and saturate to `u64::MAX`.

Shnatsel (approved):
This fixes the immediate problem I was having with the overflows, and perfect is the enemy of good, so I'm OK with landing this as-is.

We can explore alternative approaches in follow-up PRs; we still have plenty of time before shipping v0.18 on stable.

197g:
Good catch. I think we ought to remove that check then for the rowline-by-rowline case in a follow-up. I'll try to come up with a testcase that exercises that kind of usage and run it on wasm32. For now, let's simply modify the check to use the new method and add a note.

kornelski (approved):

PR #609, opened 2025-06-27
Implement Splat interlacing

197g:
Closes: #576 by replacement.

**Note that the constructor for Adam7PassInfo::new was modified** but the new parameter is type compatible. This is public and should be part of patch notes.

Differences to the previous PR:
- The width information is passed via the Adam7Iterator into the pass information
- Introduces a field to the `Reader` which stores the active interlacing method. The configuration for exposing this are TBD in the PR but the enum could just be made public with a setter in `DecodeOptions`. Its `Default` impl marks the default behavior.
- More extensive testing

### DIFF
--- a/src/adam7.rs
+++ b/src/adam7.rs
@@ -305,331 +305,337 @@ pub fn expand_pass(
     }
 }
 
-#[test]
-fn test_adam7() {
-    /*
-        1646
-        7777
-        5656
-        7777
-    */
-    let it = Adam7Iterator::new(4, 4);
-    let passes: Vec<_> = it.collect();
-    assert_eq!(
-        &*passes,
-        &[
-            Adam7Info {
-                pass: 1,
-                line: 0,
-                width: 1
-            },
-            Adam7Info {
-                pass: 4,
-                line: 0,
-                width: 1
-            },
-            Adam7Info {
-                pass: 5,
-                line: 0,
-                width: 2
-            },
-            Adam7Info {
-                pass: 6,
-                line: 0,
-                width: 2
-            },
-            Adam7Info {
-                pass: 6,
-                line: 1,
-                width: 2
-            },
-            Adam7Info {
-                pass: 7,
-                line: 0,
-                width: 4
-            },
-            Adam7Info {
-                pass: 7,
-                line: 1,
-                width: 4
-            }
-        ]
-    );
-}
+#[cfg(test)]
+mod tests {
+    use super::*;
 
-#[test]
-fn test_subbyte_pixels() {
-    const BIT_POS_1: [u8; 8] = [7, 6, 5, 4, 3, 2, 1, 0];
-
-    let scanline = &[0b10101010, 0b10101010];
-    let pixels = subbyte_values(scanline, BIT_POS_1, 1).collect::<Vec<_>>();
-
-    assert_eq!(pixels.len(), 16);
-    assert_eq!(pixels, [1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0]);
-}
-
-#[test]
-fn test_expand_adam7_bits() {
-    let width = 32;
-    let bits_pp = 1;
-    let stride = width / 8;
-    let info = |pass, line, img_width| create_adam7_info_for_tests(pass, line as u32, img_width);
-
-    let expected = |offset: usize, step: usize, count: usize| {
-        (0..count)
-            .map(move |i| step * i + offset)
-            .map(|i| BitPostion {
-                byte: i / 8,
-                bit: (i % 8) as u8,
-            })
-            .collect::<Vec<_>>()
-    };
-
-    for line_no in 0..8 {
-        let start = 8 * line_no * width;
-
+    #[test]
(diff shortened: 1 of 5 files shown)


src/decoder/mod.rs
```diff
@@ -306,6 +307,8 @@ pub struct Reader<R: BufRead + Seek> {
     scratch_buffer: Vec<u8>,
     /// Whether `ImageEnd` was already reached by `fn finish`.
     finished: bool,
+    /// How to expand interlaced data we encounter.
+    interlace: Adam7Variant,
```
197g: This should be settable, and there should be a method of selecting `Adam7Variant::Splat`. The simple choice would be `DecodeOptions` but if this would be useful to change each frame, discuss.

src/adam7.rs
```diff
@@ -88,16 +162,16 @@ impl Adam7Iterator {
...
-            7 => (w, (h - 1.0) / 2.0),
-            _ => unreachable!(),
-        };
+
+        let info = PassConstants::PASSES[self.current_pass as usize - 1];
+        // Note: integer arithmetic `div_ceil` would not handle the cases `w < 8` where the
+        // intermediate value becomes negative. Since offset <= 0.5 * sampling we get the correct
+        // ceiling value with floating point arithmetic however. Also note the float arithmetic is
+        // exact since all divisions are pure exponent offsets and the mantissa holds any possible
+        // `[-8; 0) + u32[]` value.
```
fintelia: I think it should work if we use `saturating_sub` and `div_ceil` together: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=17f92086d15d1db125435913458c5c02

197g: Since this isn't a performance sensitive path I'd rather prefer the readability. Just ran it on the whole range of `0..=u32::MAX` just to be sure, and it checks out. I'll change that.

197g: Note: I found why we're not using that yet, but we can easily bump the MSRV to `1.73`, as required, in the `0.18` release series.

src/decoder/mod.rs
anforowicz: `DecodeOptions` seems like a reasonable choice.  I don't see a scenario where changing this setting per-frame would be useful.

anforowicz:
> @anforowicz If you'd be so kind to confirm if the implementation addresses the [original issue](https://issues.chromium.org/issues/379305021), e.g. via a code review.

Thanks!  This is a step toward addressing https://crbug.com/379305021, but to unblock progress on Chromium side, we would need a way to expose "splatting" or _individual rows_ through the public API.  This is because in Chromium case, decoding weaves in and out of the `png` crate as follows:

* First `png::Reader::next_interlaced_row` decodes a single row
* Then Chromium/Skia may translate the row (e.g. alpha-premultiply, do `cICP`-chunk-based color-space-correction, do RGBA<->BGRA conversion, etc).  For non-interlaced images this translation is done into the final image buffer, but in case of interlaced images, this translation is done [into an intermediate buffer](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/experimental/rust_png/decoder/impl/SkPngRustCodec.cpp;l=655;drc=d350ca68909171a740a8e33c43fea86ed0574a05) which later needs to be "splat" or expanded into the final image (which is something that the `png` crate know how to do), so:
* Finally [`png::expand_interlaced_row`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/experimental/rust_png/ffi/FFI.rs;l=688;drc=70253db1ecfe261003756f0d81ae30929cc77ee4) is called [from here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/experimental/rust_png/decoder/impl/SkPngRustCodec.cpp;l=661-664;drc=d350ca68909171a740a8e33c43fea86ed0574a05)

So I think the `DecodeOptions`-level setting is a great way to expose the "splatting" behavior, but we'd also need a way to choose between "sparse" vs "splat" when calling [`expand_interlaced_row`](https://docs.rs/png/0.18.0-rc.2/png/fn.expand_interlaced_row.html).  "a way" = maybe this function just needs a new parameter?

src/adam7.rs
```diff
@@ -126,6 +200,39 @@ impl Iterator for Adam7Iterator {
...
+/// from a slow IO medium while optimizing time-to-first-meaningful-paint and then replacing the
+/// presented data as it is streamed in.
+///
+/// There are trade-offs to make here. The strictly necessary requirement for an implementation is
+/// that the exact image is recovered after all passes are applied. However the intermediate states
+/// of the output are left to the implementation, as long as it follows the restriction of
+/// resulting in the intended image when all passes have been applied.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Adam7Variant {
```
197g, line 218: While no longer strictly necessary in the API this is a nice documentation entry point. We can also make future use of this type when either the decoder is changed or we provide a wrapper around an image buffer (e.g. support for reading `read_frame` into an explicitly strided output buffer would be nice and somewhat painless to provide).

PR #610, opened 2025-07-01
Release notes and metadata for v0.18.0-rc.3

197g:
Given that all reported Chromium issues have been addressed, I would cut a release. We can also do version `v0.18` proper if your encoder API concerns do not feel critical.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,10 @@
 * The decoding of Adam7 interlaced data is now much faster.
 * The `acTL` chunk is now ignored when it is invalid, instead of producing
   errors while reading or decoding the following APNG chunks.
+* The requirement of the `fcTL` chunk for the default image to match the IHDR's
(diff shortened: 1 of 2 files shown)


197g:
I just came across #417 and would like to look into fixing that API difference before `0.18`. So let's keep it an rc for now.

CHANGES.md
```diff
@@ -9,12 +9,28 @@
 * `StreamingDecoder::update` now takes a structured `UnfilterBuf` argument
   instead of a direct reference to a vector. This allows in-place
   decompression. There is a public constructor for `UnfilterBuf`.
+* The methods `Decoder::output_buffer_size` and `output_line_size` now return
+  `Option<usize>` to reflect that these calculations can overflow on some
+  targets and the required buffers can not be represented in the address space.
```
fintelia: nit: cannot

PR #615, opened 2025-07-02
Add buffer to `StreamWriter` to allow reuse between rows

okaneco:
Fix a TODO by extracting the temporary `Vec` from `Write::write` for `StreamWriter` to a field in `StreamWriter`, avoiding an allocation for every filtered row.

This seemed like low-hanging perf fruit but I don't have any benchmark numbers.
Not sure if this wasn't done in the past because of some borrow checker limitations.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1380,6 +1380,7 @@ pub struct StreamWriter<'a, W: Write> {
     writer: Wrapper<'a, W>,
     prev_buf: Vec<u8>,
     curr_buf: Vec<u8>,
+    filtered_buf: Vec<u8>,
     /// Amount of data already written
     index: usize,
     /// length of the current scanline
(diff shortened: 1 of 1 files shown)


src/encoder.rs
```diff
@@ -1380,6 +1380,7 @@ pub struct StreamWriter<'a, W: Write> {
     writer: Wrapper<'a, W>,
     prev_buf: Vec<u8>,
     curr_buf: Vec<u8>,
+    filtered_buf: Vec<u8>,
```
okaneco, line 1383: Wondering if the three buffers should be encoded as `Box<[u8]>` to ensure that the lengths are never mutated?

Then `vec![0; in_len].into_boxed_slice()` in the constructor.

fintelia (approved):

src/encoder.rs
197g: Rather not. Actually there should be length information associated with the line buffer. If we consider encoding of APNG frames we will have to change the length at runtime (soonish) for functionality anyways. I don't think the gains of a `usize` in type size warrant the readability change.

PR #620, opened 2025-07-06
Remove `State::ParseChunkData` and inline `reserve_current_chunk`

fintelia:
This simplifies reading of chunk data by combining it into a single state. Previously, a second state was used to signal that the buffer needed to grow or more input data was required. Eliminating the `reserve_current_chunk` method in favor of directly inlining also makes the code a bit easier to read.

I did change the memory limits logic: it now will allocate up to the full amount of the `self.limits.bytes` rather than subtracting the current buffer capacity first. I don't really understand the rationale for the previous approach, so if anyone thinks the old behavior is correct, please say so!

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -61,11 +61,8 @@ enum State {
         accumulated_count: usize,
     },
     /// In this state we are reading chunk data from external input, and appending it to
-    /// `ChunkState::raw_bytes`.
+    /// `ChunkState::raw_bytes`. Then if all data has been read, we parse the chunk.
     ReadChunkData(ChunkType),
-    /// In this state we check if all chunk data has been already read into `ChunkState::raw_bytes`
-    /// and if so then we parse the chunk.  Otherwise, we go back to the `ReadChunkData` state.
-    ParseChunkData(ChunkType),
     /// In this state we are reading image data from external input and feeding it directly into
     /// `StreamingDecoder::inflater`.
     ImageData(ChunkType),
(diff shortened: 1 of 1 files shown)


src/decoder/stream.rs
```diff
@@ -758,25 +741,36 @@ impl StreamingDecoder {
...
                     } = &mut self.current_chunk;
 
+                    if raw_bytes.len() == raw_bytes.capacity() {
+                        if self.limits.bytes == 0 {
+                            return Err(DecodingError::LimitsExceeded);
+                        }
+
+                        // Double the size of the Vec, but not beyond the allocation limit.
+                        debug_assert!(raw_bytes.capacity() > 0);
+                        let reserve_size = raw_bytes.capacity().min(self.limits.bytes);
```
kornelski, line 751: I think the total size of the chunk should also be available at this point. You could include that in the calculation, to avoid the doubling overshooting the chunk size.

fintelia: I think the overshooting is partially intentional to avoid later re-allocations since this same buffer is used for staging all chunks. Though there's also the question of how justified it is to repeatedly double the buffer size rather than growing it to the needed length right away.

We might want to think more about chunk size handling overall. Right now if an image contains an sBIT chunk that's 1 GB, we'll first read that entire 1 GB into memory before going to parse it and noticing that the chunk exceeds the maximum allowed size of 4 bytes.

For now, I'm going to merge this PR and then think about better size handling for a followup

Issue #622, opened 2025-07-07
CRC checking should happen before chunk parsing


fintelia:
The default behavior in libpng for auxiliary chunks is to only use them if their CRCs are valid. In fact, we even have a method that [claims to enable this behavior](https://docs.rs/png/latest/png/struct.DecodeOptions.html#method.set_skip_ancillary_crc_failures). However, at the moment we cannot actually do so because we parse and save the info from each chunk before checking the CRC. That means by the time we've checked the checksum it is too late to do anything but return an error and stop decoding. We should instead delay parsing until after the checksum has been validated.

(This is in addition to fully ignoring CRCs if that is requested)

fintelia:
That's already mostly how it works. At the moment we (1) incrementally is compute the CRC as the chunk is read, (2) call `parse_chunk`, and (3) read the 4-byte CRC value at the end of the chunk and compare it to the computed CRC. All we really need to do is swap the last two steps

PR #623, opened 2025-07-07
Simplify the set of states returned from `StreamingDecoder::update`

fintelia:
Instead of having dedicated states like `Decoded::Header`, the caller is now expected to just look for the appropriate `Decode::ChunkComplete(...)` return code. This also means that all chunk parsing results in `Decode::Nothing` so the return values from the various parsing functions have all been removed.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -278,6 +278,7 @@ pub(crate) enum FormatErrorInner {
         /// The type of the unrecognized critical chunk.
         type_str: ChunkType,
     },
+    BadGammaValue,
 }
 
 impl error::Error for DecodingError {
@@ -420,6 +421,7 @@ impl fmt::Display for FormatError {
             UnrecognizedCriticalChunk { type_str } => {
                 write!(fmt, "Unrecognized critical chunk: {:?}", type_str)
             }
+            BadGammaValue => write!(fmt, "Bad gamma value."),
         }
     }
 }
@@ -1385,10 +1387,10 @@ impl StreamingDecoder {
         } else {
             let mut buf = &self.current_chunk.raw_bytes[..];
             let source_gamma: u32 = buf.read_be()?;
-            // The spec says that "A gAMA chunk containing zero is meaningless".
-            // So let's ignore such `gAMA` chunks.
             if source_gamma == 0 {
-                return Ok(Decoded::Nothing);
+                return Err(DecodingError::Format(
+                    FormatErrorInner::BadGammaValue.into(),
+                ));
             }
 
             let source_gamma = ScaledFloat::from_scaled(source_gamma);


examples/pngcheck.rs
```diff
@@ -251,6 +246,9 @@ fn check_image<P: AsRef<Path>>(c: Config, fname: P) -> io::Result<()> {
                             fctl.blend_op,
                         );
                     }
+                    ImageData => {
+                        //println!("got {} bytes of image data", data.len())
```
kornelski, line 250: println

src/decoder/stream.rs
```diff
@@ -83,20 +83,20 @@ impl State {
 pub enum Decoded {
     /// Nothing decoded yet
     Nothing,
-    Header(u32, u32, BitDepth, ColorType, bool),
```
197g, line 86: Yay for not having _data_ in these variants but I'm wondering if we should still emit an event different from `Nothing` in case we have been modifying (or supposed to be modifying) the `info()` struct. At least for the decoder that seems interesting information (as per the modifications necessary to the example).

I suppose it can be inferred from the related `ChunkComplete` of the individual pieces of information when early information is crucial. But that may get .. complicated due to the entanglement with the CRC treatment and us ignoring some invalid auxiliary chunks; so these are not precisely the same event. In some cases we have a completion event, but _no_ change to to the info struct. That isn't exactly critical but may be worth considering.

All that said, with the current control flow the `Nothing` events allow us to continue in the stream's decoding loop while other events bubble up all the way to the `Decoder` and incur the whole function call chain to continue. So maybe it's for the better?

197g: To clarify: all in all I'm in favor of simplifying the state and event enum. The control flow does become cleaner with this changes.

fintelia: How would you feel about folding #622 into this PR? Then we could make `Decoded::ChunkComplete` mean that the chunk was parsed successfully with a valid CRC (and thus stored in `info()`), and add a `DecodedChunkSkipped` to signal that the chunk was corrupt or set to ignored.

That would mean that `Decoded::ChunkBegin` would always be paired with an associated complete / skipped state (or a fatal error) rather than sometimes just being missing

197g: A bit mixed since this will be a difference between IDAT and other chunks. However, such difference already exist and generally, this seems like a nice way to encode the relevant parts of the state machine transitions in one single event. So yeah, I'd like to see how it would look.

For Final verdict however I would need to try writing a stream processor simplifying PNGs into validated / normalized parts as part of an example.

fintelia: I ended up adding two additional states. `SkippedAncillaryChunk` for when the chunk type is unrecognized or specifically requested to be skipped, and `BadAncillaryChunk` for when there's some problem with the chunk.

I've also been thinking about whether to include some error information in the `BadAncillaryChunk` variant. That would give us a way of exposing benign errors without terminating the decoding process, which is something we've discussed a few times.

197g: I would consider _not_ stick the actual error information into the BadAncillaryChunk variant. At worst the error info will affect the enum's size and while there is a 'fault' it's arguably not supposed to be fatal. So really you should be expecting to handle a backlog of these at a time, not one-by-one. Also, you shouldn't be needing the precise reader state as a caller, `png` still needs that consistent to parse further, so buffering of that information should be expected and possible.

src/decoder/stream.rs
```diff
@@ -964,34 +981,53 @@ impl StreamingDecoder {
...
             chunk::sRGB => self.parse_srgb(),
-            chunk::cICP => Ok(self.parse_cicp()),
-            chunk::mDCV => Ok(self.parse_mdcv()),
-            chunk::cLLI => Ok(self.parse_clli()),
-            chunk::eXIf => Ok(self.parse_exif()),
-            chunk::bKGD => Ok(self.parse_bkgd()),
+            chunk::cICP => self.parse_cicp(),
+            chunk::mDCV => self.parse_mdcv(),
+            chunk::cLLI => self.parse_clli(),
+            chunk::bKGD => self.parse_bkgd(),
```
197g, line 1014: I wonder if we should instead treat *all* parse errors in non-critical chunk, including their positions in the chunk stream, as `SkippedAncillaryChunk` as well. Well, not really blocking.

src/decoder/stream.rs
```diff
@@ -1378,145 +1411,162 @@ impl StreamingDecoder {
...
-                chromaticities,
-                max_luminance,
-                min_luminance,
-            })
-        }
+    fn parse_mdcv(&mut self) -> Result<(), DecodingError> {
+        let info = self.info.as_mut().unwrap();
 
         // The spec requires that the mDCV chunk MUST come before the PLTE and IDAT chunks.
         // Additionally, we ignore a second, duplicated mDCV chunk (if any).
```
197g: The comment here still says `ignore` but it now errors.

src/decoder/stream.rs
```diff
@@ -1378,145 +1411,162 @@ impl StreamingDecoder {
...
+    fn parse_cicp(&mut self) -> Result<(), DecodingError> {
+        let info = self.info.as_mut().unwrap();
 
-            // RGB is currently the only supported color model in PNG, and as
-            // such Matrix Coefficients shall be set to 0.
-            if matrix_coefficients != 0 {
-                return Err(std::io::ErrorKind::InvalidData.into());
-            }
+        // The spec requires that the cICP chunk MUST come before the PLTE and IDAT chunks.
+        // Additionally, we ignore a second, duplicated cICP chunk (if any).
```
197g: Similar `ignore` comment.

src/decoder/stream.rs
```diff
@@ -1763,33 +1811,46 @@ impl StreamingDecoder {
...
 
-        Ok(Decoded::Nothing)
+        Ok(())
     }
 
     // NOTE: This function cannot return `DecodingError` and handles parsing
     // errors or spec violations as-if the chunk was missing.  See
     // https://github.com/image-rs/image-png/issues/525 for more discussion.
-    fn parse_bkgd(&mut self) -> Decoded {
+    fn parse_bkgd(&mut self) -> Result<(), DecodingError> {
```
197g: Also no longer accurate with regards to the return type.

src/decoder/stream.rs
```diff
@@ -1763,33 +1811,46 @@ impl StreamingDecoder {
...
+                FormatErrorInner::AfterIdat { kind: chunk::bKGD }.into(),
+            ));
+        }
+
+        let expected = match info.color_type {
+            ColorType::Indexed => {
+                if info.palette.is_none() {
+                    return Err(DecodingError::IoError(
+                        std::io::ErrorKind::InvalidData.into(),
+                    ));
```
197g, line 1835: I'd keep the current behavior of skipping the chunk. This is semantic evaluation similar to a zeroed `gAMA` chunk.

fintelia:
The `parse_chunk` method contains logic that converts any `DecodingError::FormatError` or `ErrorKind::UnexpectedEof` from an ancillary chunk into an `Ok(Decoded::BadAncillaryChunk)`:

https://github.com/image-rs/image-png/blob/0a50c02a4f7b344e7815de82042c48fb526fde49/src/decoder/stream.rs#L1033-L1063

PR #630, opened 2025-08-24
Support skipping chunks without loading them into memory

fintelia:
This PR changes the decoder to skip chunks that are unrecognized or have the wrong size, without fulling loading them into memory first. It also means that a couple limits checks can be removed because the maximum memory that might be allocated is small and bounded.

Previously, the decoder would accept chunks that were too large and simply ignore any trailing bytes. After this PR, any such chunks are treated as corrupt. Before merging, we'll want to investigate whether overly large chunks are common in the wild.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -64,6 +64,8 @@ enum State {
     /// In this state we are reading chunk data from external input, and appending it to
     /// `ChunkState::raw_bytes`. Then if all data has been read, we parse the chunk.
     ReadChunkData(ChunkType),
+    /// Skip past chunk without processing it and then return the given value for the decoding result.
+    SkipChunk(Decoded),
     /// In this state we are reading image data from external input and feeding it directly into
     /// `StreamingDecoder::inflater`.
     ImageData(ChunkType),
@@ -270,8 +272,8 @@ pub(crate) enum FormatErrorInner {
     UnexpectedRestartOfDataChunkSequence {
         kind: ChunkType,
     },
-    /// Failure to parse a chunk, because the chunk didn't contain enough bytes.
-    ChunkTooShort {
+    /// Failure to parse a chunk, because the chunk had the wrong number of bytes.
+    ChunkLengthWrong {
         kind: ChunkType,
     },
     UnrecognizedCriticalChunk {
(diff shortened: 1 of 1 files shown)


src/decoder/stream.rs
```diff
@@ -988,6 +1004,63 @@ impl StreamingDecoder {
...
+            }
+            _ => {
+                self.current_chunk.remaining += 4; // Also skip the CRC
+                return Ok(State::SkipChunk(Decoded::SkippedAncillaryChunk(type_str)));
+            }
+        };
+
+        if !target_length.contains(&length) {
+            // Uncomment to detect unexpected chunk lengths during testing.
+            // panic!("chunk type_str={type_str:?} has length={length}, target_length={target_length:?}");
```
fintelia, line 1059: @Shnatsel if you have a chance to check whether this panic ever happens on your test corpus, that would be super helpful. Either to detect that I messed up one of the chunk size ranges or because there's some images with wrongly sized chunks

Shnatsel: Is this bit of code sufficient to exercise this codepath or do I need to invoke something else?

    let mut reader = Decoder::new(Cursor::new(image)).read_info().unwrap();
    reader.next_frame(output).unwrap();

fintelia: That should be mostly sufficient, however adding a `finish` call will also check any chunks after the image data:
```
...
reader.finish().unwrap();
```

Shnatsel: I don't see a single occurrence of that in the wild on none of my corpora. Only on test files and fuzzer-generated files.

This is very unusual. Something is clearly selecting against them. I suspect libpng just rejects such files outright, that's why none of them made it into my corpora.

Shnatsel: This is the code I've used:

```rust
use std::io::Cursor;

pub type BoxResult<T> = Result<T, Box<dyn std::error::Error + Send + Sync>>;

fn main() -> BoxResult<()> {
    let path = std::env::args_os().skip(1).next().expect("Please specify the input file");
    eprintln!("{}", path.display());
    let image = std::fs::read(path)?;
    let mut reader = png::Decoder::new(Cursor::new(image)).read_info()?;
    let mut output = vec![0; reader.output_buffer_size().unwrap()];
    reader.next_frame(&mut output)?;
    reader.finish()?;
    Ok(())
}
```

fintelia: That's especially weird because the PNG spec recommends ignoring invalid ancillary chunks rather than rejecting the files

src/decoder/stream.rs
```diff
@@ -988,6 +1004,63 @@ impl StreamingDecoder {
...
+
+            _ if is_critical(type_str) => {
+                return Err(DecodingError::Format(
+                    FormatErrorInner::UnrecognizedCriticalChunk { type_str }.into(),
+                ));
+            }
+            _ => {
+                self.current_chunk.remaining += 4; // Also skip the CRC
+                return Ok(State::SkipChunk(Decoded::SkippedAncillaryChunk(type_str)));
+            }
```
BenBE, line 1054: This allows for hiding CRC failures for auxiliary chunks, which previously were detected.

fintelia: Good catch. We should be considering `skip_ancillary_crc_failures` here

PR #631, opened 2025-08-26
Release information for 0.18.0

197g:
With the state machine change resolved and good feedback from the Chromium integration process, let's cut the full release for `0.18`. This branches also updates our dev-dependencies a bit to cut the depencency tree. (Most of the ~99 compiled crates are in criterion and clap, fine with me but if someone takes issue we can reduce that further).


197g:
Change notes were drafted along the release candidates. I _think_ they are up to date except for the state machine changes. Feel free to push notes into the branch, otherwise I'll update as soon as we agree on doing the release.

PR #632, opened 2025-08-28
perf: portable_simd 3bpp paeth filter

Sentimentron:
With `--features=unstable`, this implementation will decode up to 16 pixels at once.

Reduces the number of cycles required to decode a 768px row by 20-30% on Arm systems, and around 50-70% on x64 systems ([see discussion](https://github.com/image-rs/image-png/discussions/416)

Results on an AMD EPYC 7B13 system
    $  taskset -c 64-68 cargo +nightly bench --features=benchmarks,unstable -- filter=Paeth/bpp=3
    unfilter/filter=Paeth/bpp=3
        time:   [18.987 µs 18.994 µs 19.001 µs]
        thrpt:  [616.74 MiB/s 616.96 MiB/s 617.19 MiB/s]

    $ taskset -c 64-68 cargo +nightly bench --features=benchmarks,unstable -- filter=Paeth/bpp=3
        unfilter/filter=Paeth/bpp=3
            time:   [15.355 µs 15.361 µs 15.367 µs]
            thrpt:  [762.59 MiB/s 762.89 MiB/s 763.18 MiB/s]
        change:
            time:   [-19.180% -19.087% -19.007%] (p = 0.00 < 0.05)
            thrpt:  [+23.468% +23.589% +23.732%]


Sentimentron:
AI disclosure: I wrote an original sliding-window 3bpp Paeth portable_simd implementation and optimized it for best performance on the Cortex A520. I then used the Gemini family of LLMs to automatically refine the code to achieve the best possible code-generation and performance across all other micro-architectures. The PR is derived from that work, but includes documentation and other cleanups.

src/filter.rs
```diff
@@ -2,6 +2,175 @@ use core::convert::TryInto;
...
+        let (simd_prev_row, remainder_prev_row) = prev.split_at(chunks * STRIDE_BYTES);
+        let row_iter = simd_row.chunks_exact_mut(STRIDE_BYTES);
+        let prev_iter = simd_prev_row.chunks_exact(STRIDE_BYTES);
+        let combined_iter = row_iter.zip(prev_iter);
+
+        for (chunk, prev_chunk) in combined_iter {
+            let mut x: Simd<u8, STRIDE_BYTES> = Simd::<u8, STRIDE_BYTES>::from_slice(chunk);
+            let b: Simd<u8, STRIDE_BYTES> = Simd::<u8, STRIDE_BYTES>::from_slice(prev_chunk);
+
+            // Process the chunk using the SIMD helper, passing the initial 'c' (vlast)
```
Sentimentron: Remove (`vlast` was renamed to just `c`)

src/filter.rs
```diff
@@ -2,6 +2,175 @@ use core::convert::TryInto;
...
+            a = process_paeth_chunk_bpp3_s48(a, &b, c, &mut x);
+
+            // Update `vlast` for the next chunk: it's the upper-left of the next chunk,
```
Sentimentron: Ditto

src/filter.rs
```diff
@@ -2,6 +2,175 @@ use core::convert::TryInto;
...
+        process_pixel!(39);
+        process_pixel!(42);
+        process_pixel!(45);
+
+        *x_out += Simd::from_array(preds);
+        current_a
+    }
+
+    // Apply Paeth unfiltering in 16 pixel chunks (3bpp).
+    pub fn paeth_3bpp(current: &mut [u8], prev: &[u8]) {
```
Sentimentron: Revise to something more descriptive (e.g. `paeth_unfilter_3bpp`)

examples/unfilter_testbench.rs
```diff
@@ -0,0 +1,111 @@
+// Microbenchmark for all the various filters
```
Shnatsel: We already have a benchmark for unfiltering, https://github.com/image-rs/image-png/blob/master/benches/unfilter.rs

I think it would be better to unify them so that both are runnable through `cargo bench`

Sentimentron: I think I should remove this before merging: it's only really useful for the micro-architectural simulation (since Criterion.rs is not designed to run in such a restricted environment), but it's here in the PR for context and if you wish to request clarifications and changes.

examples/unfilter_testbench.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use png::Filter;
+use std::hint::black_box;
+
+const CURRENT_ROW: [u8; 771] = [
+    30, 134, 150, 121, 53, 45, 57, 186, 248, 150, 4, 5, 212, 192, 104, 161, 238, 94, 32, 89, 199,
```
197g: Style nit: `include_bytes!` of a file with those bytes achieves the same result and a more readable module.  (Giving me strong C #embed vibes).

src/filter.rs
```diff
@@ -2,6 +2,185 @@ use core::convert::TryInto;
...
+    use core::convert::TryInto;
+    use core::simd::cmp::{SimdOrd, SimdPartialOrd};
+    use core::simd::num::{SimdInt, SimdUint};
+    use core::simd::{LaneCount, Simd, SupportedLaneCount};
+
+    // Import the fastest arch-specific scalar implementations from the outer crate.
+    #[cfg(target_arch = "aarch64")]
+    use crate::filter::filter_paeth as filter_paeth_decode;
+    #[cfg(target_arch = "x86_64")]
+    use crate::filter::filter_paeth_stbi as filter_paeth_decode;
```
fintelia: We should share this with the scalar logic (replacing the current `if` statement). And we'll want to pick something for non-x86_64 and non-aarch64 architectures to avoid a compile error.

Also, `filter_paeth_chosen` might be a good name?

src/filter.rs
```diff
@@ -2,6 +2,185 @@ use core::convert::TryInto;
...
+
+    // AArch64 Paeth predictor. Ported from the libpng implementation at
+    // https://github.com/pnggroup/libpng/blob/master/arm/filter_neon_intrinsics.c
```
fintelia: It is better to use [doc comments](https://doc.rust-lang.org/rust-by-example/meta/doc.html#doc-comments) (starting with `///`) to document methods since they are treated specially by Rust tooling.

src/filter.rs
```diff
@@ -574,25 +753,32 @@ pub(crate) fn unfilter(
...
-                        *c0 = c0.wrapping_add(filter_paeth_decode(a_bpp[0], *p0, c_bpp[0]));
-                        *c1 = c1.wrapping_add(filter_paeth_decode(a_bpp[1], *p1, c_bpp[1]));
-                        *c2 = c2.wrapping_add(filter_paeth_decode(a_bpp[2], *p2, c_bpp[2]));
-
-                        a_bpp = [*c0, *c1, *c2];
-                        c_bpp = [*p0, *p1, *p2];
+                        simd::paeth_unfilter_3bpp(current, previous);
+                    }
+                    #[cfg(not(feature = "unstable"))]
+                    {
```
fintelia: We went through a couple variations of cfg dispatch, and IMO the cleanest / least error-prone turned out to be: 
```rust
#[cfg(feature = "unstable")]
{
    simd::paeth_unfilter_3bpp(current, previous);
    return;
}

// Fallback case
let mut a_bpp = [0; 3];
let mut c_bpp = [0; 3];
...
```

fintelia (commented):
These are quite significant improvements! On my Zen 3 machine, I see a ~11% speed improvement compiling with the default x86-64 target, and a ~30% improvement when compiling for  `x86-64-v2` or `x86-64-v3`

src/filter.rs
Sentimentron: Fixed in subsequent commit

src/filter.rs
Sentimentron: Fixed in subsequent commit

src/filter.rs
Sentimentron: Fixed with new name in subsequent commit

examples/unfilter_testbench.rs
Sentimentron: Now removed from latest tip

Shnatsel:
That is correct, `portable_simd` does not do runtime feature selection. 

It can be added as a layer on top of `portable_simd` but that will require some `unsafe` in the implementation. The [`multiversion`](https://crates.io/crates/multiversion) crate provides a safe wrapper around that using a proc macro.

I suggest exploring multiversioning separately from this PR, since it is a mostly orthogonal concern.

Shnatsel:
We never really got 3bpp Paeth to vectorize, so it makes sense that large gains are possible here.

On a desktop Zen4 the difference is under 5%, but at least it seems to be stable and not noise. "My CPU is too good for optimization work" is a weird problem to have.

Shnatsel:
Bad news: this is a 10% regression on an Apple M4 laptop.

Measured with `cargo +nightly bench --features=benchmarks,unstable -- filter=Paeth/bpp=3`

Sentimentron:
Yes, looks like I get a similar result on my personal M2 laptop:

```
unfilter/filter=Paeth/bpp=3
                        time:   [19.663 µs 19.667 µs 19.671 µs]
                        thrpt:  [595.75 MiB/s 595.87 MiB/s 595.98 MiB/s]
                 change:
                        time:   [+11.301% +11.404% +11.494%] (p = 0.00 < 0.05)
                        thrpt:  [-10.309% -10.236% -10.154%]
                        Performance has regressed.
``` 

I don't know if the LLVM cost model for the M1 is in any way accurate, but I think we can gate this off for Mac OS for now.

Shnatsel:
Apple uses LLVM for all their C/C++/Objective-C/Swift compilation, so I'd be very much surprised if the LLVM cost model for Apple was way off.

I am curious what your microarchitecture simulator (llvm-mca?) has to say about this code on Apple systems. If it still shows a performance improvement on Apple, that's a major red flag flag for its reliability.

Sentimentron:
Looking at the public LLVM cost model, seems like all the Apple cores are modelled using this SchedCyclone def:

https://github.com/llvm/llvm-project/blame/main/llvm/lib/Target/AArch64/AArch64SchedCyclone.td

Doesn’t seem to have received any significant updates in the last 11 years or so. It’s possible that they do this properly in their proprietary Clang, but I don’t think the LLVM MCA-based simulator that I can run will accurately model this processor. This work is mostly about lifting up the Arm low-end (Cortex A520/A35/A53) which are all relatively simple to model - any extra things that benefit are just a bonus!

Shnatsel:
I wonder if this is at all beneficial in practice on lower-end ARM, since older SoCs are heavily bottlenecked by memory bandwidth, so microarchitectural optimizations might be entirely irrelevant.

It's a hassle to get benchmarks from Android, but maybe we could get this benchmarked on a Raspberry Pi or some such?

awxkee:
Making effective intrinsics on NEON for such a small series like `[u8; 3]` is incredible complicated even with direct intrinsics, I almost never had success with this. NEON instructions are power hungry so instead of cheap scalar instructions on big cores they consule roughly 2 times more energy to execute and cause more heat, to process such a small series. As a result, this can lead to throttling.

> Does PNG decoding get scheduled onto little cores often? I would imagine that interactive workloads like web browsing would run mostly on the big cores, with the little ones mostly getting non-interactive background processes.

It starts on the little cores, and if the workload does not decrease, it switches to the big cores (if not pinned). After that, it may throttle.

Sentimentron:
> ~I'm curious what the microarchitecture simulator has to say about #635 which significantly improves benchmarks for x86_64, Apple M3 and gives an especially huge boost to Graviton 3.~ I misread the benchmarks, it actually regresses ARM a lot. Never mind.

I was curious too. Testing #635, the micro-architecture simulator indicates this:
Filter | Arm Cortex A520 | Arm Cortex X4 | Intel Arrow Lake | AMD Zen 5 | Arm Cortex A520 (baseline cycles) | Arm Cortex X4 (baseline cycles) | Intel Arrow Lake (baseline cycles) | AMD Zen 5 (baseline cycles)
-- | -- | -- | -- | -- | -- | -- | -- | --
Paeth (3bpp) | -13.92% | -18.52% | 1.35% | 0.20% | 1073543 | 520265 | 1158373 | 505037
Paeth (4bpp) | 9.53% | -7.49% | 6.28% | -4.25% | 885251 | 288701 | 413955 | 259222

The PR doesn't say which CPU it was measured on, but the results indicate a 6% uplift on both bpp=3 and bpp=4. I've measured it on my Epyc system and it yields about a 2% uplift. It's not surprising that the micro-arch simulation is a little off for the reasons I've discussed (it assumes basically a "perfect" path to main memory with no latency), this is what LLVM's cost-model _thinks_ the code will cost to execute. A theoretically worse solution can be better in practise as long as it decreases stalled cycles.

awxkee:
FWIW 
To make this PR effective on NEON, it’s necessary to figure out a way to utilize the pipeline and instruction scheduling, to compensate introduced latency. On AArch64, you can’t just generate nearly the same amount of code with more power hungry instructions and expect guaranteed performance improvements. Some gains might appear sometimes, but they often come at the cost of shifting work from one bottleneck to another.
For example, instead of processing `pixel(row, 0), pixel(row, 1), pixel(row, 2)`, process `process_pixels(row0, row1, row2, row3, 0), process_pixels(row0, row1, row2, row3, 1)` to give LLVM chance to schedule instructions on the NEON pipeline, since multiple rows will be processed for the same latency ( as they are not blocking each other ) if there is more than one SIMD unit on a chip.

Sentimentron:
Yes, you've very succinctly highlighted the main problem with optimizing anything for AArch64: there are some wildly different micro-architectures, and it's not always possible to get an uplift on everything in all circumstances. There are very big out-of-order cores like the X4 where instruction scheduling doesn't matter so much, and some that are effectively in-order (though dual-issue) like the Cortex-A53/Cortex-A35/Cortex-A520 (little family), where instruction scheduling matters an awful lot. The little cores tend to be 30-40% of the performance of the big cores. I'm intrigued at breaking the data-dependency vertically, but I think it would probably work best for really small images due to the data locality. 

<details>
<summary>Disclosure</summary>
I worked at Arm for around 10 years on AArch64 software optimization. I currently work in Google's Pixel software division on optimizing Chrome's performance on AArch64. This contribution is on behalf of my current employer.
</details>

Sentimentron:
Although, @awxkee's comment does have some truth to it. Switching off SLP auto-vectorization entirely (i.e. no neon generated):

```
$ RUSTFLAGS="-C llvm-args=--slp-threshold=30000" cargo +nightly bench --features=benchmarks -- filter=Paeth/bpp=3
unfilter/filter=Paeth/bpp=3
                        time:   [14.587 µs 14.619 µs 14.675 µs]
                        thrpt:  [798.57 MiB/s 801.59 MiB/s 803.35 MiB/s]
                 change:
                        time:   [-17.228% -17.138% -16.971%] (p = 0.00 < 0.05)
                        thrpt:  [+20.440% +20.683% +20.813%]
                        Performance has improved.
```

This indicates that... yes, someone should definitely try and model Apple Silicon a bit better in upstream LLVM: it's a very unusual CPU.

Sentimentron:
@awxkee thank you for the discussion.

Doing this (or not) is a strategic trade-off. The Paeth filter's unbreakable top-to-bottom, left-to-right data dependency and its many formulations make it really challenging to optimize (especially the 3bpp version). With that constraint in mind, this implementation succeeds in its primary goal: delivering a significant performance improvement for low-end Arm cores and most x64 systems, which will benefit many users. By explicitly writing a SIMD-style solution, it delivers reasonable performance irrespective of whether LLVM's SLP auto-vectorization is working effectively, without the maintenance overhead of architecture-specific intrinsics.

To handle the regressions, my preferred solution is to gate the new code behind a `cfg` flag and switch it off on platforms like Apple Silicon. Let me know your thoughts on this approach.

Sentimentron:
@197g that's an interesting idea, and it brings up an important point about how we should handle performance features.

The `unstable` feature flag, in my opinion, should serve as a well-managed "expert mode" for sophisticated users who are willing to use a nightly compiler to get bleeding-edge performance. It's not a guarantee of universal improvement, but it's an opportunity to gain access to something potentially faster. However, we shouldn't put something in that mode if it's definitely known to regress a given platform: it breaks the implicit good-faith promise that the `unstable` feature makes. I think that `cfg(not(any(target_os="ios", target_os="macos")))` is OK for this: the code can still be tested on MacOS in this config, it just won't be used when decoding images.

Should I proceed with gating the code off for Apple Silicon?

Shnatsel:
Apple Silicon also has big and little cores. I've only benchmarked with the built-in harness which runs in a loop for a long time, so I assume it gets placed on the big core. It's possible that on the little cores this would be beneficial on Apple Silicon; I think we should benchmark with core pinning before gating it off.

Sentimentron:
I found a couple of ways to try and measure this, the first is using the `taskpolicy` command:

```
$ taskpolicy -b cargo +nightly bench --features=benchmarks -- filter=Paeth/bpp=3
unfilter/filter=Paeth/bpp=3
                        time:   [70.397 µs 70.568 µs 70.719 µs]
                        thrpt:  [165.71 MiB/s 166.06 MiB/s 166.47 MiB/s]
                 change:
                        time:   [+0.0841% +2.6611% +5.2874%] (p = 0.04 < 0.05)
                        thrpt:  [−5.0219% −2.5921% −0.0841%]
Found 14 outliers among 100 measurements (14.00%)
  9 (9.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe

$ taskpolicy -b cargo +nightly bench --features=benchmarks,unstable -- filter=Paeth/bpp=3
unfilter/filter=Paeth/bpp=3
                        time:   [83.191 µs 83.257 µs 83.331 µs]
                        thrpt:  [140.63 MiB/s 140.75 MiB/s 140.87 MiB/s]
                 change:
                        time:   [+18.000% +19.620% +21.775%] (p = 0.00 < 0.05)
                        thrpt:  [−17.881% −16.402% −15.254%]
                        Performance has regressed.
Found 11 outliers among 100 measurements (11.00%)
  3 (3.00%) low severe
```

That's at background QoS level (so barely idle) - it's actually quite comparable to the performance on an Cortex A520, despite being a vastly more powerful chip. Playing with the latency and throughput tiers did not reveal any interesting behaviour. The other option seems to be putting the device on low power mode:

```
$ taskpolicy -a cargo +nightly bench --features=benchmarks -- filter=Paeth/bpp=3
                        time:   [31.426 µs 31.440 µs 31.459 µs]
                        thrpt:  [372.51 MiB/s 372.73 MiB/s 372.90 MiB/s]
                 change:
                        time:   [−0.9927% −0.6168% −0.3670%] (p = 0.00 < 0.05)
                        thrpt:  [+0.3683% +0.6206% +1.0026%]

$ taskpolicy -a cargo +nightly bench --features=benchmarks,unstable -- filter=Paeth/bpp=3
unfilter/filter=Paeth/bpp=3
                        time:   [35.013 µs 35.018 µs 35.024 µs]
                        thrpt:  [334.59 MiB/s 334.65 MiB/s 334.70 MiB/s]
                 change:
                        time:   [+11.368% +11.424% +11.485%] (p = 0.00 < 0.05)
                        thrpt:  [−10.302% −10.253% −10.208%]
                        Performance has regressed.
  2 (2.00%) high mild
  4 (4.00%) high severe
```

I'm not sure whether this means background QoS = efficiency core at the lowest frequency and low power mode = efficiency core at the higher frequency, or low power mode = performance core at the lowest frequency. 

But basically, seems whatever issue is affecting Apple Silicon, it's worse on the efficiency core.

Sentimentron:
`cfg` gate has been implemented, PR is ready for the next round of review

src/filter.rs
```diff
@@ -573,6 +774,11 @@ pub(crate) fn unfilter(
                         }
                     }
                     BytesPerPixel::Three => {
+                        #[cfg(all(feature = "unstable", not(target_vendor = "apple")))]
```
197g, line 775: Could you comment here linking to the benchmark tables in the PR thread? Maybe with some of the key numbers but it does not duplicate all the details.

Sentimentron: Addressed in next round.

PR #492, opened 2024-08-15
Use SIMD to speed up Paeth unfiltering for bpp=4, bpp=8

okaneco:
- Add SIMD-accelerated unfiltering for `BytesPerPixel::{Four, Eight}`
- Add `paeth_predictor_u8`, `paeth_step_u8`, and `unfilter_paeth_u8` for calculating the Paeth predictor entirely in `Simd<u8, N>` without converting to `Simd<i16, N>`
- Parameterize `PaethState` by `T`, the `SimdElement` type

Calculating the filter entirely in `u8` avoids penalties which result from unpacking to and packing from `Simd<i16, N>`.
`pa`, `pb`, and `pc` can all be calculated using absolute differences and reusing the logic of `filter::filter_paeth`.

---

This PR duplicates the existing logic for unfilter3 and unfilter6. I tried to apply this approach to those cases but it was not a complete win: staying in `u8` was faster for the SSE2 baseline profile but slower for AVX2.

Benchmark numbers
```bash
# 1. bpp=4, autovectorization
$ RUSTFLAGS="-Ctarget-cpu=x86-64-v3" cargo +nightly bench --features=benchmarks -- filter=Paeth/bpp=4            
                        time:   [12.895 µs 12.900 µs 12.905 µs]
                        thrpt:  [1.1824 GiB/s 1.1829 GiB/s 1.1833 GiB/s]

# 2. bpp=4, portable_simd
$ RUSTFLAGS="-Ctarget-cpu=x86-64-v3" cargo +nightly bench --features=benchmarks,unstable -- filter=Paeth/bpp=4
                        time:   [11.143 µs 11.162 µs 11.195 µs]
                        thrpt:  [1.3630 GiB/s 1.3671 GiB/s 1.3694 GiB/s]

# 3. bpp=8, autovectorization
$ RUSTFLAGS="-Ctarget-cpu=x86-64-v3" cargo +nightly bench --features=benchmarks -- filter=Paeth/bpp=8
                        time:   [18.209 µs 18.226 µs 18.246 µs]
                        thrpt:  [1.6725 GiB/s 1.6744 GiB/s 1.6760 GiB/s]

# 4. bpp=8, portable_simd
$ RUSTFLAGS="-Ctarget-cpu=x86-64-v3" cargo +nightly bench --features=benchmarks,unstable -- filter=Paeth/bpp=8
                        time:   [11.172 µs 11.180 µs 11.187 µs]
[...]

PR #633, opened 2025-08-29
perf: portable_simd 4bpp filtering

Sentimentron:
Results on an AMD EPYC 7B13 system with `-C target-cpu=native` are basically unchanged, though micro-architectural simulation indicates a 40% reduction in cycles on a 768 pixel row. This could indicate the system is bottlenecked somewhere else (e.g. memory). 

```
$ RUSTFLAGS="-C target-cpu=native" taskset -c 64-68 cargo +nightly bench --features=benchmarks -- unfilter/filter=Paeth/bpp=4
    unfilter/filter=Paeth/bpp=4
                        time:   [11.241 µs 11.246 µs 11.252 µs]
                        thrpt:  [1.3561 GiB/s 1.3568 GiB/s 1.3574 GiB/s]
                 change:
                        time:   [-0.3785% -0.2463% -0.1349%] (p = 0.00 < 0.05)
                        thrpt:  [+0.1351% +0.2469% +0.3800%]

$ RUSTFLAGS="-C target-cpu=native" taskset -c 64-68 cargo +nightly bench --features=benchmarks,unstable -- unfilter/filter=Paeth/bpp=4
    unfilter/filter=Paeth/bpp=4
                        time:   [11.243 µs 11.253 µs 11.263 µs]
                        thrpt:  [1.3548 GiB/s 1.3560 GiB/s 1.3572 GiB/s]
                 change:
                        time:   [-0.0754% +0.0009% +0.0761%] (p = 0.98 > 0.05)
                        thrpt:  [-0.0760% -0.0009% +0.0754%]
```

The micro-arch simulation also indicates a 21.56% improvement on Cortex A520, an 8.5% improvement on the Cortex X4, and a 56.61% improvement on x64 Arrow Lake.

Marking as draft until PR #632 is completed.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -895,7 +895,7 @@ pub(crate) fn unfilter(
                     BytesPerPixel::Three => {
                         #[cfg(all(feature = "unstable", not(target_vendor = "apple")))]
                         {
-                            // Results in PR: https://github.com/image-rs/image-png/pull/63
+                            // Results in PR: https://github.com/image-rs/image-png/pull/632
                             // Approximately 30% better on Arm Cortex A520, 7%
                             // regression on Arm Cortex X4. Switched off on Apple
                             // Silicon due to 10-12% regression.
@@ -1052,7 +1052,7 @@ pub(crate) fn unfilter(
                     BytesPerPixel::Three => {
                         #[cfg(feature = "unstable")]
                         {
-                            // Results in PR: https://github.com/image-rs/image-png/pull/63
+                            // Results in PR: https://github.com/image-rs/image-png/pull/632
                             // 23% better on an Epyc 7B13, 10% on a Zen 3 part.
                             // ~30% when targeting x86-64-v2.
                             simd::paeth_unfilter_3bpp(current, previous);


okaneco:
In https://github.com/image-rs/image-png/pull/492, I changed the implementation at the time to keep the Paeth filtering in `u8` lanes.
On x86, this should be faster than unpacking to i16 and packing back to u8.
I don't know the effect on arm64.

This was also before the stb filter was added, so that may be faster than the `filter_paeth_fpnge` method in that PR.

Shnatsel:
By default we assume that our users are NOT using `-C target-cpu=native`, and instead optimize for the baseline instruction sets available on each platform. On x86_64 that means SSE2. Chromium ships with SSE3 enabled but that doesn't really add anything meaningful.

At least on ARM we can always assume NEON because it's part of the baseline for Aarch64 and Chromium ships with it enabled even on 32-bit ARM.

Sentimentron:
No meaningful change in perf for Apple Silicon (M2):
```
unfilter/filter=Paeth/bpp=4
                        time:   [20.200 µs 20.226 µs 20.255 µs]
                        thrpt:  [771.42 MiB/s 772.52 MiB/s 773.53 MiB/s]
                 change:
                        time:   [-0.1253% -0.0189% +0.0894%] (p = 0.74 > 0.05)
                        thrpt:  [-0.0893% +0.0189% +0.1254%]
```

Sentimentron:
Benchmark results on real silicon (Pixel 10): the Cortex-A520 improves from 220.7 MiB/s to 313.2 MiB/s (41.92%), the Cortex X4 improves from 850.8 MiB/s to 938.1 MiB/s (10.26%).

src/filter.rs
```diff
@@ -290,6 +290,127 @@ mod simd {
...
+                    a_bpp[2].into(),
+                    b_bpp[2].into(),
+                    c_bpp[2].into(),
+                )),
+                chunk[3].wrapping_add(filter_paeth_chosen(
+                    a_bpp[3].into(),
+                    b_bpp[3].into(),
+                    c_bpp[3].into(),
+                )),
+            ];
```
fintelia, line 408: Is there a benefit of unrolling this versus the loop version that the scalar implementation uses?

Sentimentron: Replacing with this fragment:

```rust
let mut a_bpp = a.to_array();
        let mut c_bpp = c.to_array();
        for (chunk, b_bpp) in remainder_row
            .chunks_exact_mut(BPP)
            .zip(remainder_prev_row.chunks_exact(BPP))
        {
            for i in 0..4 {
                chunk[i] = chunk[i].wrapping_add(filter_paeth_chosen(
                    a_bpp[i].into(),
                    b_bpp[i].into(),
                    c_bpp[i].into(),
                ));
            }
            a_bpp.copy_from_slice(chunk);
            c_bpp = b_bpp.try_into().unwrap();
        }
 ```

Seems to result in a 0.48% increase in cycles under the simulation - exactly why is not clear to me just yet.

197g:
@Sentimentron Thanks, if your employer lets you grind this then I'm glad for all improvements you can eke out. We _may_ want to revisit how the unfilter implementation to run is selected. At the moment that branch is based on a `cfg` and chosen inside the `unfilter` function. As the number of possibilities increases, the selection logic may be more wieldy as a separate utility which returns a function pointer or table of such instead ([this style](https://github.com/etemesi254/zune-image/blob/5460142430a699be884f53e61b39fefbc772b0ee/crates/zune-jpeg/src/idct.rs#L51)). If the current one impedes testability we should switch, I trust you'll see for yourself when you get it.

PR #636, opened 2025-08-31
Revert "perf: Improve Paeth unfilter auto-vectorization by changing how loop state is maintained"

Shnatsel:
Reverts image-rs/image-png#635 because it regressed ARM and I failed to notice the flipped sign :face_exhaling: 

It still looks absolutely worth it for x86.

PR #637, opened 2025-08-31
Revert ARM performance regression

Shnatsel:
A less invasive alternative to #636 that only reverts the part that regressed ARM. 

@okaneco planned for this possibility and split the changes into two separate commits; this lets us keep x86 gains while undoing the damage done to ARM.

This reverts commit a7b6e1cf7b1e56c29647e0d4922ab737a15e221c.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -93,6 +93,31 @@ impl RowFilter {
     }
 }
 
+// This code path is used on non-x86_64 architectures but we allow dead code
+// for the test module to be able to access it.
+#[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


fintelia (approved):

197g:
Given that we dispatch on `bytes_per_pixel`, which is constant over a frame, it may make sense to write these indivual blocks as individual functions that we combine at runtime into a vtable of only the compression methods? I'll test at some point the next week if that is more readable without losing performance. That way we may be able to make much more fine grained decisions per architecture, e.g. choosing different for bpp=3 etc.

Issue #639, opened 2025-09-03
Unexpected change of behavior when reading partial data


anforowicz:
A Skia test [decodes the first 0x8D bytes of a PNG image](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/tests/SkPngRustDecoderTest.cpp;l=365-421;drc=4703976a4daee1b35b7f865c9791047688c6c4c6) and [expects the first 10 rows of the images to successfully decode before getting `UnexpectedEof` error](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/tests/SkPngRustDecoderTest.cpp;l=422;drc=4703976a4daee1b35b7f865c9791047688c6c4c6).  This roughly translates to the following ad-hoc test in the `png` crate:

```
#[test]
fn test_partial_decode() {
    // The first 0x8D bytes from the following test image from Skia:
    // resources/images/apng-test-suite--dispose-ops--none-basic.png
    let partial_png: &[u8] = &[
        0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 0x00, 0x00, 0x00, 0x0d, 0x49, 0x48,
        0x44, 0x52, 0x00, 0x00, 0x00, 0x80, 0x00, 0x00, 0x00, 0x40, 0x08, 0x06, 0x00, 0x00,
        0x00, 0xd2, 0xd6, 0x7f, 0x7f, 0x00, 0x00, 0x00, 0x08, 0x61, 0x63, 0x54, 0x4c, 0x00,
        0x00, 0x00, 0x03, 0x00, 0x00, 0x00, 0x01, 0xb9, 0xea, 0x8a, 0x56, 0x00, 0x00, 0x00,
        0x1a, 0x66, 0x63, 0x54, 0x4c, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x80, 0x00,
        0x00, 0x00, 0x40, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x0a, 0x00,
        0x64, 0x00, 0x01, 0x26, 0x12, 0x2f, 0xe0, 0x00, 0x00, 0x00, 0x93, 0x49, 0x44, 0x41,
        0x54, 0x78, 0x9c, 0xed, 0xd2, 0xa1, 0x01, 0x00, 0x30, 0x10, 0x84, 0xb0, 0xdb, 0x7f,
[...]

PR #640, opened 2025-09-04
Allow full unfiltering for partial data

197g:
Closes: #639

With the in-place unfilter buffer we forced unfiltering to trail behind the deflate window whereas previously it could copy data that was still used for back-reference and filter on it. That resulted in better performance but for _partial_ images we lost out on 32kB of available image data. This relaxes the `UnfilteringBuffer` to allow both: use in-place for the common path but add a fallback for unfiltering the back reference portion of the data when we hit `UnexpectedEof`.

Performance with this patch is _mostly_ good but I'm hitting some erratic, weird fluctuations on the noncompressed tests (slower) and paletted-zune (faster) and Exoplanet_Phase_Curve_(Diagram).png (also much _faster_??). I'd very much like to understand the individual causes here given that we do just two extra bounds checks and an integer-min that shouldn't be in extremely tight loops.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -589,7 +589,7 @@ impl<R: BufRead + Seek> Reader<R> {
     ) -> Result<(), DecodingError> {
         self.next_raw_interlaced_row(rowlen)?;
         let row = self.unfiltering_buffer.prev_row();
-        let row = &row[..rowlen - 1];
+        debug_assert_eq!(row.len(), rowlen - 1);
 
         // Apply transformations and write resulting data to buffer.
         let transform_fn = {
--- a/src/decoder/unfiltering_buffer.rs
+++ b/src/decoder/unfiltering_buffer.rs
@@ -26,6 +26,9 @@ pub(crate) struct UnfilteringBuffer {
     available: usize,
     /// The number of bytes before we shift the buffer back.
     shift_back_limit: usize,
+    /// Size of the previously unfiltered row. For interlaced images this changes over time, from
+    /// pass to pass. Also if we get other subframes.
+    unfiltered_len: usize,
 }
 
 impl UnfilteringBuffer {
@@ -45,6 +48,9 @@ impl UnfilteringBuffer {
         debug_assert!(self.available <= self.filled);
         // The logically filled region is always within bounds of the data stream.
         debug_assert!(self.filled <= self.data_stream.len());
+        // Note: `unfiltered_len`, `prev_start` and `current_start` do not always hold, for
+        // instance at the end of an image, since we update some of this lazily. We do not care
+        // about that invariant right now.
     }
 
     /// Create a buffer tuned for filtering rows of the image type.
@@ -98,6 +104,7 @@ impl UnfilteringBuffer {
             filled: 0,
             available: 0,
             shift_back_limit,
+            unfiltered_len: 0,
         };
 
         result.debug_assert_invariants();
@@ -117,17 +124,13 @@ impl UnfilteringBuffer {
         self.current_start = 0;
         self.filled = 0;
         self.available = 0;
+        self.unfiltered_len = 0;
         self.debug_assert_invariants();
     }
 
     /// Returns the previous (already `unfilter`-ed) row.
     pub fn prev_row(&self) -> &[u8] {
-        &self.data_stream[self.prev_start..self.current_start]
-    }
-
-    /// Returns how many bytes of the current row are present in the buffer.
-    pub fn curr_row_len(&self) -> usize {
-        self.available - self.current_start
+        &self.data_stream[self.prev_start..self.current_start][..self.unfiltered_len]
     }
 
     /// Returns how many bytes of the current row are available in the buffer in the mutable region.
@@ -233,6 +236,7 @@ impl UnfilteringBuffer {
 
         self.prev_start = self.current_start + 1;
         self.current_start += rowlen;
+        self.unfiltered_len = row.len();
 
         self.debug_assert_invariants();
 
@@ -265,7 +269,7 @@ impl UnfilteringBuffer {
         // so just put it right before the minimum of `current_start` and `available`. In this case
         // however we should also pass an empty row to `unfilter`.
         if self.prev_start == self.current_start {
-            let potential_end_of_new_prev = self.current_start.min(self.available);
+            let potential_end_of_new_prev = self.prev_start.min(self.available);
             // Insert free space between what we treat as the previous row and the current row.
             // NOTE: this is because of the decoder's `commit` function which will take a fixed
             // window of data back from its filled state before it is done. But we move that region
@@ -302,6 +306,7 @@ impl UnfilteringBuffer {
 
             prev.copy_from_slice(&row[1..rowlen]);
             unfilter(filter, bpp, &[], prev);
+            self.unfiltered_len = prev.len();
 
             self.prev_start = start_of_new_prev;
             self.current_start += rowlen;
@@ -333,6 +338,7 @@ impl UnfilteringBuffer {
 
             // Do NOT modify prev_start
             self.current_start += rowlen;
+            self.unfiltered_len = prev.len();
             self.debug_assert_invariants();
 
             Ok(true)


197g:
Performance seems to have been an artifact of the known problems llvm. Here's the results from Rust `1.89` locally:

```
test result: ok. 0 passed; 0 failed; 82 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running benches/decoder.rs (target/release/deps/decoder-bbb4e9ee3bf79167)
decode/Lohengrin_-_Illustrated_Sporting_and_Dramatic_News.png
                        time:   [13.993 ms 14.008 ms 14.026 ms]
                        thrpt:  [761.35 MiB/s 762.31 MiB/s 763.14 MiB/s]
                 change:
                        time:   [+0.3531% +0.5748% +0.8034%] (p = 0.00 < 0.05)
                        thrpt:  [−0.7970% −0.5715% −0.3518%]
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high severe
decode/Transparency.png time:   [48.908 µs 48.935 µs 49.007 µs]
                        thrpt:  [6.8414 GiB/s 6.8514 GiB/s 6.8553 GiB/s]
                 change:
                        time:   [+0.1524% +0.2277% +0.3460%] (p = 0.00 < 0.05)
                        thrpt:  [−0.3448% −0.2272% −0.1522%]
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
decode/kodim02.png      time:   [2.0754 ms 2.0758 ms 2.0765 ms]
                        thrpt:  [541.79 MiB/s 541.97 MiB/s 542.07 MiB/s]
                 change:
                        time:   [−3.9275% −2.3384% −1.3331%] (p = 0.00 < 0.05)
                        thrpt:  [+1.3511% +2.3944% +4.0881%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
decode/kodim07.png      time:   [2.5150 ms 2.5179 ms 2.5218 ms]
                        thrpt:  [446.11 MiB/s 446.80 MiB/s 447.31 MiB/s]
                 change:
                        time:   [+0.1683% +0.2903% +0.4104%] (p = 0.00 < 0.05)
                        thrpt:  [−0.4087% −0.2894% −0.1680%]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
decode/kodim17.png      time:   [2.0609 ms 2.0614 ms 2.0621 ms]
                        thrpt:  [545.57 MiB/s 545.74 MiB/s 545.88 MiB/s]
                 change:
                        time:   [+0.7159% +0.7368% +0.7580%] (p = 0.00 < 0.05)
                        thrpt:  [−0.7523% −0.7314% −0.7108%]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
decode/kodim23.png      time:   [1.9937 ms 1.9940 ms 1.9942 ms]
                        thrpt:  [564.15 MiB/s 564.20 MiB/s 564.27 MiB/s]
                 change:
                        time:   [+0.6497% +0.6881% +0.7242%] (p = 0.00 < 0.05)
                        thrpt:  [−0.7190% −0.6834% −0.6455%]
decode/Exoplanet_Phase_Curve_(Diagram).png
                        time:   [14.449 ms 14.462 ms 14.491 ms]
                        thrpt:  [1.5992 GiB/s 1.6024 GiB/s 1.6039 GiB/s]
                 change:
                        time:   [−1.5108% −1.0287% −0.6139%] (p = 0.00 < 0.05)
                        thrpt:  [+0.6177% +1.0394% +1.5340%]
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
decode/Exoplanet_Phase_Curve_(Diagram)_indexed_gimp.png
                        time:   [5.1186 ms 5.1201 ms 5.1210 ms]
                        thrpt:  [4.5253 GiB/s 4.5262 GiB/s 4.5275 GiB/s]
                 change:
                        time:   [+0.2168% +0.2882% +0.3530%] (p = 0.00 < 0.05)
                        thrpt:  [−0.3517% −0.2874% −0.2163%]
Found 3 outliers among 10 measurements (30.00%)
  1 (10.00%) low mild
  2 (20.00%) high mild
decode/Fantasy_Digital_Painting.png
                        time:   [12.341 ms 12.355 ms 12.371 ms]
                        thrpt:  [511.36 MiB/s 512.02 MiB/s 512.60 MiB/s]
                 change:
                        time:   [+0.9006% +1.0055% +1.1120%] (p = 0.00 < 0.05)
                        thrpt:  [−1.0998% −0.9954% −0.8926%]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
decode/lorem_ipsum_oxipng.png
                        time:   [432.52 µs 435.01 µs 438.94 µs]
                        thrpt:  [3.1781 GiB/s 3.2068 GiB/s 3.2253 GiB/s]
                 change:
                        time:   [+0.1332% +1.4542% +2.7337%] (p = 0.06 > 0.05)
                        thrpt:  [−2.6610% −1.4334% −0.1330%]
decode/lorem_ipsum_screenshot.png
                        time:   [660.77 µs 661.09 µs 661.31 µs]
                        thrpt:  [2.8126 GiB/s 2.8135 GiB/s 2.8149 GiB/s]
                 change:
                        time:   [+0.6805% +0.7392% +0.7972%] (p = 0.00 < 0.05)
                        thrpt:  [−0.7909% −0.7338% −0.6759%]
decode/paletted-zune.png
                        time:   [6.0131 ms 6.0136 ms 6.0143 ms]
                        thrpt:  [2.1435 GiB/s 2.1438 GiB/s 2.1439 GiB/s]
                 change:
                        time:   [+1.0916% +1.1040% +1.1169%] (p = 0.00 < 0.05)
                        thrpt:  [−1.1046% −1.0919% −1.0799%]
                        Performance has regressed.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low mild
  1 (10.00%) high mild
decode/tango-icon-address-book-new-128-rsvg-convert.png
                        time:   [56.076 µs 56.077 µs 56.079 µs]
                        thrpt:  [1.0884 GiB/s 1.0884 GiB/s 1.0884 GiB/s]
                 change:
                        time:   [−0.2360% −0.1647% −0.1115%] (p = 0.00 < 0.05)
                        thrpt:  [+0.1116% +0.1650% +0.2365%]
decode/tango-icon-address-book-new-16.png
                        time:   [4.7660 µs 4.7695 µs 4.7743 µs]
                        thrpt:  [204.55 MiB/s 204.75 MiB/s 204.90 MiB/s]
                 change:
                        time:   [+0.7646% +0.8426% +0.9170%] (p = 0.00 < 0.05)
                        thrpt:  [−0.9087% −0.8355% −0.7588%]
decode/tango-icon-address-book-new-32.png
                        time:   [8.0016 µs 8.0396 µs 8.1525 µs]
                        thrpt:  [479.15 MiB/s 485.88 MiB/s 488.18 MiB/s]
                 change:
                        time:   [−0.0613% +0.6959% +2.1059%] (p = 0.42 > 0.05)
                        thrpt:  [−2.0625% −0.6911% +0.0613%]
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe

generated-noncompressed-4k-idat/8x8.png
                        time:   [685.14 ns 685.40 ns 685.71 ns]
                        thrpt:  [356.04 MiB/s 356.20 MiB/s 356.34 MiB/s]
                 change:
                        time:   [−0.0314% +0.0248% +0.0823%] (p = 0.41 > 0.05)
                        thrpt:  [−0.0822% −0.0248% +0.0314%]
  1 (1.00%) low mild
  4 (4.00%) high mild
  3 (3.00%) high severe
generated-noncompressed-4k-idat/128x128.png
                        time:   [8.9758 µs 8.9768 µs 8.9778 µs]
                        thrpt:  [6.7984 GiB/s 6.7992 GiB/s 6.8000 GiB/s]
                 change:
                        time:   [+2.4892% +2.6191% +2.7237%] (p = 0.00 < 0.05)
                        thrpt:  [−2.6514% −2.5523% −2.4288%]
                        Performance has regressed.
  1 (1.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
generated-noncompressed-4k-idat/2048x2048.png
                        time:   [1.9937 ms 2.0081 ms 2.0235 ms]
                        thrpt:  [7.7219 GiB/s 7.7809 GiB/s 7.8372 GiB/s]
                 change:
                        time:   [−10.610% −6.6236% −2.8469%] (p = 0.00 < 0.05)
                        thrpt:  [+2.9303% +7.0935% +11.869%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
Benchmarking generated-noncompressed-4k-idat/12288x12288.png: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 5.5s or enable flat sampling.
generated-noncompressed-4k-idat/12288x12288.png
                        time:   [99.363 ms 99.636 ms 100.01 ms]
                        thrpt:  [5.6243 GiB/s 5.6455 GiB/s 5.6610 GiB/s]
                 change:
                        time:   [−0.2676% −0.0491% +0.1835%] (p = 0.71 > 0.05)
                        thrpt:  [−0.1832% +0.0491% +0.2683%]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe

generated-noncompressed-64k-idat/128x128.png
                        time:   [7.6085 µs 7.6092 µs 7.6099 µs]
                        thrpt:  [8.0205 GiB/s 8.0213 GiB/s 8.0220 GiB/s]
                 change:
                        time:   [+0.0696% +0.0869% +0.1040%] (p = 0.00 < 0.05)
                        thrpt:  [−0.1039% −0.0868% −0.0695%]
```

src/decoder/unfiltering_buffer.rs
```diff
@@ -202,6 +237,106 @@ impl UnfilteringBuffer {
...
+            );
+
+            let prev = &mut prev[self.prev_start..][..rowlen - 1];
+
+            let filter = RowFilter::from_u8(row[0]).ok_or(DecodingError::Format(
+                FormatErrorInner::UnknownFilterMethod(row[0]).into(),
+            ))?;
+
+            // Unfilter this in a temporary buffer.
+            let mut row = row[1..rowlen].to_vec();
```
anforowicz, line 335: nit: IIUC `to_vec` will always allocate here.  Would it be possible to avoid doing such an allocation for ~every row in a partially read input?  Can we allocate once and then reuse the allocation?

src/decoder/unfiltering_buffer.rs
```diff
@@ -202,6 +237,106 @@ impl UnfilteringBuffer {
...
+            // upwards so it may erroneously point backwards by more than necessary. That just
+            // freezes data but it might also freeze a portion that we are using as the unfilter
+            // buffer / the space to put the 'previous row'.
+            let padding = rowlen - 1;
+
+            let start_of_new_prev = potential_end_of_new_prev.saturating_sub(padding);
+            let end_of_new_prev = start_of_new_prev + padding;
+
+            // Shift everything up as required.
+            let current_shift = end_of_new_prev - potential_end_of_new_prev;
```
anforowicz, line 285: I wonder if instead of reusing/shifting `self.data_stream` it may be a bit easier/clearer if a separate buffer (or two) would be introduced as a field of `UnfilteringBuffer`?  Say `copied_curr_row: Vec<u8>` and `copied_prev_row: Option<Vec<u8>>` or something like that?

anforowicz: Oh, and if we do that, then I guess the type of the `current_start` and `prev_start` fields would need to change from `usize` to something like `Either<usize>` (where `Either::Left` means the old behavior - data is in `data_stream` and where `Either::Right` means that the rows are in the new buffers).

197g: The type change in `current_start` was the main concern that drove me away from that concept. I wanted to keep influence on the main path as low as possible, having an `Either` would be an extra match / conditional read on every single path accessing `current_start`—even the happy one. I can benchmark it but intuitively, it will be quite visible.

anforowicz:
> I think there might be a complication with this approach: `fdeflate` actually uses the size of the output buffer / the value of output_index to know whether a back-reference is valid or should trigger `DistanceTooFarBack`. If an EOF is encountered and a few rows of data are prepended to the unfiltering buffer, then subsequent calls to decompress may wrongly think that a back-reference is valid when it isn't.

This doesn't seem like a hard blocker - we just need to ensure that `output: &mut [u8]` passed to [`fdeflate::Decompressor::read`](https://docs.rs/fdeflate/0.3.7/fdeflate/struct.Decompressor.html) doesn't include the prepended/unfiltered rows, right?  In other words, I think we would just need to use `&mut buffer[available..output_limit]` rather than `&mut buffer[..output_limit]` in `zlib.rs` here: https://github.com/image-rs/image-png/blob/14496f23c98dba6d52288ddf46e514d75eb2e58d/src/decoder/zlib.rs#L120.

Does that sound reasonable?

src/decoder/unfiltering_buffer.rs
```diff
@@ -111,16 +124,27 @@ impl UnfilteringBuffer {
...
+    /// Why negative? There's a backreference section in the buffer that is reserved for the zlib /
+    /// deflate decoder (or any other for that matter). We must not modify that part of the data
+    /// but if we only *read* the row then the start of our current_start is allowed to run _into_
+    /// that section. So as a less efficient fallback where partial data is crucial we can unfilter
+    /// outside the read area instead of in-place but still indicate how many bytes we have
+    /// consumed. In that case, this number is negative.
+    pub(crate) fn curr_row_len_with_unfilter_ahead(&self) -> isize {
+        // Both are indices into the buffer that is an allocation of bytes. Allocations are at most
+        // isize::MAX in size hence this is valid.
+        self.available as isize - self.current_start as isize
```
anforowicz: I would like to tentatively propose 3 change here:

* Can this function be renamed to `mutable_len_of_curr_row`?  (first of all - does such a name make sense?  and second - is this a good/readable/helpful name?  FWIW when seeing `curr_row_len_with_unfilter_ahead` I don't really know what this means / what this tries to convey)
* Can this keep returning `usize` by clamping at zero / by using `saturating_sub` - see: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2da4b40a9bcac8e9dd6232ab1a427867
* I wonder if the comment can be shortened + moved from the doc comment (`///`) into an implementation comment (`//`), saying something like: "Using `saturating_sub`, because after `fn unfilter_ahead_row` it is possible that `self.available < self.current_row`."

src/decoder/unfiltering_buffer.rs
```diff
@@ -34,11 +37,20 @@ impl UnfilteringBuffer {
     /// Asserts in debug builds that all the invariants hold.  No-op in release
     /// builds.  Intended to be called after creating or mutating `self` to
     /// ensure that the final state preserves the invariants.
+    #[track_caller]
     fn debug_assert_invariants(&self) {
+        // The previous row pointer is always behind the current row pointer.
         debug_assert!(self.prev_start <= self.current_start);
-        debug_assert!(self.current_start <= self.available);
+        // The current row pointer is always into filled bytes (but sometimes may point to
+        // read-only back reference buffer, see `curr_row_len_with_unfilter_ahead`).
```
anforowicz, line 45: Let me propose a minor tweak to address these perceived issues:

* "current row pointer" is a bit ambiguous - it doesn't directly refer to the `current_start` field, so when reading it I find myself thinking: "is this the same as `current_start` field or did the author use a different wording to hint that this means something else?"
* I think that "sometimes" can be succinctly summarized

Proposed tweak:

> `current_start` always points into filled (i.e. decompressed) bytes.  It is possible both for
> 1) `current_start <= available` (when unfiltering in-place via `fn unfilter_curr_row`)
> 2) `current_start > available` (when unfiltering via `fn unfilter_ahead_row`)

src/decoder/unfiltering_buffer.rs
```diff
@@ -197,11 +235,114 @@ impl UnfilteringBuffer {
...
 
         self.debug_assert_invariants();
 
         Ok(())
     }
+
+    /// Unfilter but allow the current_start to exceed `available` at the cost of some compute.
+    /// This will be called when we encounter an `UnexpectedEof` and want to push out all
+    /// interlaced rows that we can, i.e. it is not in the usual critical path of decoding.
+    pub(crate) fn unfilter_ahead_row(
```
anforowicz, line 249: I would also like to propose renaming the `fn unfilter...` functions into:

* `fn unfilter_curr_row_in_place`
* `fn unfilter_curr_row_in_scratch_buffer`

WDYT?  I think those names may convey more directly what those functions do?

src/decoder/unfiltering_buffer.rs
```diff
@@ -111,16 +124,27 @@ impl UnfilteringBuffer {
         self.current_start = 0;
         self.filled = 0;
         self.available = 0;
```
anforowicz, line 126: I wonder if (maybe even as a separate PR, or at least a separate commit) it would help to change how `UnfilteringBuffer` is initialized.  In particular, I wonder if things would be simplified if `UnfilteringBuffer` would **always** have room for 2 mutable rows (2 times `rowlen` bytes).  IMHO this may be beneficial in the following ways:

* This way `prev_row` **always** exists (initially filled with zeros) and we don’t need to worry about special-casing unfiltering of the first row.
* This way there is no need for shifting/shuffling things around in `fn unfilter_ahead_row` / `fn unfilter_curr_row_in_scratch_buffer`, because the "scratch buffer" is **always** / **already** present at the beginning of the buffer.

WDYT?

src/decoder/mod.rs
```diff
@@ -679,17 +679,38 @@ impl<R: BufRead + Seek> Reader<R> {
...
+                    // We only have partial data but we want to make as much data available as
+                    // possible. The immediate problem is now that zlib requires a lookback buffer
+                    // of prior data which must stay available while our filter code will want to
+                    // modify the data in the rowlines (i.e. prediction). So what to do?
+                    //
+                    // A copy of the data is necessary unless all filters are no-op. We must decide
+                    // how to do this copy without interfering with the main code paths. The
+                    // resulting unfilter operations puts the indices in the buffer into a weird
+                    // state at the cost of having to shuffle data around to make the actual data
+                    // valid for the following rows.
```
anforowicz, line 703: I wonder if this comment can be shortened (new/proposed text assumes that `curr_row_len_with_unfilter_header` was renamed to `mutable_len_of_curr_row` and that `fn unfilter...` functions got renamed as proposed elsewhere):

```
// `mutable_len_of_curr_row` is too small to call `unfilter_curr_row_in_place`
// and `UnexpectedEof` says that we can't decompress more data.  But we may have
// enough read-only bytes of the current row to call `unfilter_curr_row_in_scratch_buffer`.
```

---------------------

Related, but kind of separate: I also wonder if things would be cleaner if `UnfilteringBuffer` would not only have `fn mutable_len_of_curr_row` but also `fn readable_len_of_curr_row` (i.e. `filled` - `current_start`).  I think it may be cleaner to call `unfilter_curr_row_in_scratch_buffer` if we know that we have enough decompressed data - this way `unfilter_curr_row_in_scratch_buffer` / `unfilter_ahead_row` wouldn't have to return a `bool` and the code here could look like:

```
Err(DecodingError::IoError(e))
if e.kind() == std::io::ErrorKind::UnexpectedEof && self.unfiltering_buffer.readable_len_of_curr_row() >= rowlen => {
    return self.unfiltering_buffer.unfilter_curr_row_in_scratch_buffer(rowlen, self.bpp);
}
```

src/decoder/stream.rs
```diff
@@ -721,40 +721,87 @@ impl StreamingDecoder {
             U32 {
                 kind,
                 mut bytes,
-                mut accumulated_count,
+                accumulated_count,
```
anforowicz: I didn't understand the changes in `stream.rs` :-(.  I'll try harder on Monday 🙃 .

I am guessing that the `mod.rs` changes may affect how the high-level part of the decoder communicates with `stream.rs` (wrt how many times `self.decoder.decode_image_data` is called?  or maybe wrt what work needs to be done after flushing + if that work can complete without calling into `decode_image_data` again?).

But I also think/observe that if we manage to flush/finalize the zlib stream, then all the bytes should become available (supporting unfiltering in-place).  And this observation seems to indicate that no `stream.rs` changes should be needed.  😕

src/decoder/unfiltering_buffer.rs
```diff
@@ -26,6 +26,9 @@ pub(crate) struct UnfilteringBuffer {
     available: usize,
     /// The number of bytes before we shift the buffer back.
     shift_back_limit: usize,
+    /// Size of the previously unfiltered row. For interlaced images this changes over time, from
+    /// pass to pass. Also if we get other subframes.
+    unfiltered_len: usize,
```
anforowicz, line 31: Would it make sense to call this `rowlen`?

Or maybe this field can be removed, and instead we can add `rowlen` parameter to `fn prev_row`?  Would that work?

PS. I assume that this field wasn't needed in the past, because `curr_start - prev_start` was sufficient before the changes in this PR?  Can you please confirm?

src/decoder/unfiltering_buffer.rs
197g: That could be worth trying out. We have to balance it though against the slowness induced in the construction. The vector is only allocated as needed which for particularly wide images may be after only several IO calls. Alternatively to preserve that we could reserve the space for one row after we've grown to that size, too. Either way, I think a separate PR is appropriate.

src/decoder/unfiltering_buffer.rs
197g: We used `rowlen` to refer to the filter byte plus the byte data, whereas here we store `rowlen - 1`. I think it would be confusing to. But I take away that `unfiltered_len` is not the best name either.

In the one call where we handle `prev_row` there was an additional indexing by `[..rowlen - 1]`. I'm not completely sure if the necessity may have changed over time as well. So hard to confirm completely. It is definitely needed after this PR (and cleaner imo).

src/decoder/unfiltering_buffer.rs
197g: The renaming is very appropriate. I played a bit around and also looked at `checked_sub`. In the end, `saturating_sub` does a better job at being optimized with the comparison use. Also reworded the comment with the semantics for the return value which should be the relevant detail when calling this.

src/decoder/stream.rs
anforowicz: @197g, I need your help please with figuring out if/why the changes in `stream.rs` are required.

I did the following test:

* Check out ToT of `image-png` (0add8a293f2d1c2f2b0fc729b1e5b05fa00bad45)
* Patch in the latest version of https://github.com/image-rs/image-png/pull/640.patch
* Try `cargo test` - all tests passed
* Revert the changes in `stream.rs`: `$ git checkout origin/master -- src/decoder/stream.rs`
* Try `cargo test` again - all tests are still passing

So I think that either:

* If those changes are really needed, then I think the PR should provide a regression test that fails without those changes
* Or, if those changes are not needed after all, then they should be reverted

197g: Alright, let's shelf it and see.

197g:
```
Result #0: Err(Format(FormatError { inner: CorruptFlateStream { err: InsufficientInput } }))
Result #1: Err(Format(FormatError { inner: CorruptFlateStream { err: InsufficientInput } }))
Result #2: Ok("ok")

thread '<unnamed>' (31) panicked at fuzz_targets/buf_independent.rs:245:14:
Inconsistent results - some are Ok(_) and some are Err(_)
```

Definitely finds fuzzer results as conjectured. There is a varying amount of data available in an incomplete stream depending on the granularity of inputs. Each decompression call gets _exactly_ `8kB` of available output buffer and a sending more input at the same time makes it harder to use it up.

You can end up with a partially decompressed backreference as the last symbol, i.e. all input bits are consumed and the output is filled by some part of the backreference still needs to be copied over. Then the decoder state machine goes into flush where it will pass `eof=true` and that will cause it to _error_ instead of yield the remaining input. With smaller inputs (or intermittent EOF) it had more calls to decompress with a fresh buffer and will/may have retrieved the bytes.

That said, we can wait on https://github.com/image-rs/fdeflate/pull/59 as an alternative. That may be cleaner than handling it here.

src/decoder/unfiltering_buffer.rs
anforowicz: Thanks for explaining.  I actually missed/forgot that `rowlen` refers to something slightly different - I agree that using `rowlen` name here would have been confusing.

So, maybe let's just tweak the comment?:

```
/// Size of an unfiltered row (without the `FilterType` byte / `rowlen - 1`).
/// This may vary over time depending on the size of the frame being
/// currently decoded, and on the interlacing pass.
```

src/decoder/stream.rs
anforowicz: Thanks - that fuzzing result helps me to understand what may be going on.  I think that the problem is that after this PR, when `fn next_raw_interlaced_row` in `src/decoder/mod.rs` takes the `if e.kind() == std::io::ErrorKind::UnexpectedEof` branch, then it ignores `self.subframe.consumed_and_flushed` which could result in `FormatErrorInner::NoMoreImageData`.  I think that this problem could be avoided by [one of my earlier suggestions](https://github.com/image-rs/image-png/pull/640#discussion_r2594267282) to change the `match` arm condition so that it only is taken if `unfilter_curr_row_in_scratch_buffer` / `unfilter_ahead_row` is guaranteed to succeed.

src/decoder/unfiltering_buffer.rs
anforowicz: I thought about this some more and I think that I strongly preferred the `enum`-based version, because:

* There are big readability benefits IMO
* I don't have concerns about performance impact:
    - It seems that 1) decompression and actual `unfilter` are much hotter than 2) maintaining `prev_start` and/or `current_start` offsets of `UnfilteringBuffer`.  In particular the latter seems to happen on the order of O(image-height), right?  (this seems to be true in-the-big-O-sense even for interlaced images).
    - I don't see a negative impact in local benchmarks
    - Micro-optimization is the root of all evil

src/decoder/stream.rs
anforowicz: Hmm... actually I can repro a similar fuzzing failure even after augmenting the match arm condition with `readable_len_of_curr_buf` checks.   But I still think that this shouldn't happen / that this shouldn't affect how far into the input stream we call `fn decompress` / whether `finish_compressed_chunks` is called / whether `fdeflate` runs out of input data and reports an error.  I have minimized the fuzzing test input and I have a fix at https://github.com/image-rs/image-png/pull/664 (although now I see that https://github.com/image-rs/image-png/pull/662 may be an alternative fix).

197g: I don't think that's a complete picture. The result in the fuzzer indicates that the `unfilter_ahead_row` did in fact _succeed_! Note that `#2` refers to the byte-by-byte reader and `Ok(_)` is for a successfully decoded image with _all_ image data having been parsed. Note that we stop polling the zlib decompression once all rows have been decoded (from that point on we call `decode_image_data(None)` instead of passing the `UnfilterBuf`). So that also never tries to flush the stream — which means we can not get `InsufficientInput`. The other two consume part of their last output buffer with previous symbols and then have to buffer a partially consumed backreference into the flush path.

And alas, as I'm typing this, fuzzer failure shows up in #664. (Huh, just a python install issue though?)

fintelia: > That said, we can wait on https://github.com/image-rs/fdeflate/pull/59 as an alternative. That may be cleaner than handling it here.

That PR is exactly equivalent to always passing end_of_input=false. There's no need to wait on it

anforowicz:
I think that may be this PR can be closed, because it should be mostly replaced by https://github.com/image-rs/image-png/pull/664 that has landed recently?

197g:
Yep, thanks for the reminder.

PR #643, opened 2025-09-09
portable_simd version of Up

Sentimentron:
Implements a `portable_simd` version of the Up filter - merely adds the previous and current row's 64 bytes simultaneously. This filter is ludicrously fast already on a big out-of-order CPU, but does seem to help the Cortex-A520.

Results from Silicon (3bpp)

| CPU  | Baseline | Result | Speedup |
| ---  | --- | --- | --- |
| Arm  Cortex A520 | 4211.8 MiB/s | 7726.3 MiB/s | 83.44% |
| Arm Cortex X4 | 64949.0 MiB/s | 67270.0 MiB/s | 3.57% |
| Apple Silicon M2 | 65217.0 MiB/s | 65269.0 MiB/s | 0.08% |
| AMD EPYC 7B13 | 37272.0 MiB/s | 37536.0 MiB/s | 0.71% |

The same implementation is shared across all pixel depths.

Draft until the scaffolding in #632 is merged.

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -93,6 +93,207 @@ impl RowFilter {
     }
 }
 
+/// Optional module containing `portable_simd` versions of the most
+/// important unfiltering algorithms. Enable using the `unstable` feature.
+#[cfg(feature = "unstable")]
+mod simd {
+    use core::convert::TryInto;
+    use core::simd::cmp::{SimdOrd, SimdPartialOrd};
+    use core::simd::num::{SimdInt, SimdUint};
+    use core::simd::{LaneCount, Simd, SupportedLaneCount};
+
+    // Import the fastest arch-specific scalar implementations from the outer crate.
+    #[cfg(not(target_arch = "x86_64"))]
+    use crate::filter::filter_paeth as filter_paeth_chosen;
+    #[cfg(target_arch = "x86_64")]
+    use crate::filter::filter_paeth_stbi as filter_paeth_chosen;
+
+    /// Paeth predictor specialized for AArch64 systems. Ported from the libpng
+    /// implementation at
+    /// https://github.com/pnggroup/libpng/blob/master/arm/filter_neon_intrinsics.c
+    #[cfg(target_arch = "aarch64")]
+    #[inline(always)]
+    fn paeth_predictor_simd<const SIZE: usize>(
+        a_i16: Simd<i16, SIZE>,
+        b_i16: Simd<i16, SIZE>,
+        c_i16: Simd<i16, SIZE>,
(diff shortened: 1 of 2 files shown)


Sentimentron:
AI disclosure: I wrote a sliding-window `portable_simd` implementation of the Paeth filter (3bpp). I then used the Gemini family of LLMs provided by my employer to automatically adapt this code to the Up filter and optimize it to achieve the best possible code-generation and performance across all other micro-architectures in simulation. This PR is derived from that output, but includes documentation and other cleanups.

src/filter.rs
```diff
@@ -93,6 +93,232 @@ impl RowFilter {
...
+
+        let current_iter = simd_current.chunks_exact_mut(STRIDE_BYTES);
+        let previous_iter = simd_previous.chunks_exact(STRIDE_BYTES);
+        let combined_iter = current_iter.zip(previous_iter);
+
+        for (current_chunk, previous_chunk) in combined_iter {
+            let mut x: Simd<u8, STRIDE_BYTES> = Simd::<u8, STRIDE_BYTES>::from_slice(current_chunk);
+            let b: Simd<u8, STRIDE_BYTES> = Simd::<u8, STRIDE_BYTES>::from_slice(previous_chunk);
+            x = x + b; // Wrapping addition
+            x.copy_to_slice(current_chunk);
```
fintelia, line 312: It might be possible to write a scalar implementation here that rustc would be able to vectorize

src/filter.rs
```diff
@@ -93,6 +93,232 @@ impl RowFilter {
...
+        const STRIDE_BYTES: usize = 64; // Maximum supported width
+
+        let chunks = current.len() / STRIDE_BYTES;
+
+        let (simd_current, remainder_current) = current.split_at_mut(chunks * STRIDE_BYTES);
+        let (simd_previous, remainder_previous) = previous.split_at(chunks * STRIDE_BYTES);
+
+        let current_iter = simd_current.chunks_exact_mut(STRIDE_BYTES);
+        let previous_iter = simd_previous.chunks_exact(STRIDE_BYTES);
+        let combined_iter = current_iter.zip(previous_iter);
```
fintelia, line 306: I think you should be able to do something like:
```rust
let mut current_iter = current.chunks_exact_mut(STRIDE_BYTES);
let mut previous_iter = previous.chunks_exact(STRIDE_BYTES);

for (current_chunk, previous_chunk) in (&mut current_iter).zip(&mut previous_iter) {
    ...
}

let remainder_current = current_iter.remainder();
let remainder_previous = previous_iter.remainder();
...
```

PR #647, opened 2025-09-21
Add Filter::MinEntropy

fintelia:
Adds a new adaptive filter strategy inspired by the similar one in oxipng. `MinEntropy` is much slower than the current `Adaptive` filter but tends to produce smaller output.

In my experimentation, the tradeoff between better compression and longer encode time started to become worthwhile around deflate level 7 though it is somewhat subjective (and will depend on the input image).

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -20,6 +20,7 @@ pub enum Filter {
     Avg,
     Paeth,
     Adaptive,
+    MinEntropy,
 }
 
 impl Default for Filter {
@@ -88,7 +89,7 @@ impl RowFilter {
             Filter::Up => Some(Self::Up),
             Filter::Avg => Some(Self::Avg),
             Filter::Paeth => Some(Self::Paeth),
-            Filter::Adaptive => None,
+            Filter::Adaptive | Filter::MinEntropy => None,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/filter.rs
```diff
@@ -1195,46 +1196,105 @@ fn filter_internal(
...
+        counts[0][i] += counts[1][i] + counts[2][i] + counts[3][i];
+    }
+
+    // Compute the entropy.
+    let mut entropy = ilog2i(total);
+    for &count in &counts[0] {
+        if count > 0 {
+            entropy = entropy.saturating_sub(ilog2i(count));
+        }
+    }
```
okaneco, line 1298: Suggestion: Should this return early if `entropy == 0`?

Not sure how likely that case is.

src/filter.rs
```diff
@@ -1195,46 +1196,105 @@ fn filter_internal(
...
 
+/// Estimate the value of i * log2(i) without using floating point operations,
+/// implementation originally from oxipng.
+fn ilog2i(i: u32) -> u32 {
+    let log = 32 - i.leading_zeros() - 1;
+    i * log + ((i - (1 << log)) << 1)
+}
+
+fn entropy(buf: &[u8]) -> u64 {
+    let mut counts = [[0_u32; 256]; 4];
```
okaneco, line 1257: Comment: Curious what the speed difference is here for this array being 1, 2, or 4 arrays large.

I believe you that it's faster, but slightly surprised that using 4 arrays and consolidating after performs better (I see that the consolidation loop is auto-vectorized when checked in Compiler Explorer).

fintelia: This blog post goes into more details: https://fastcompression.blogspot.com/2014/09/counting-bytes-fast-little-trick-from.html

src/filter.rs
fintelia: Zero entropy should only happen if all the bytes of the row are a single value after filtering or due to rounding errors. Added a check for all the bytes being zero (constant non-zero seems improbable after filtering).

Issue #649, opened 2025-09-29
Can we make `ScaledFloat::SCALING` public?


d-e-s-o:
What it says on the tin. It's not clear to me why it's private. Is there a reason? The documentation already [mentions it](https://docs.rs/png/0.18.0/png/struct.ScaledFloat.html) so it can't really be changed without being a breaking change. Users may reasonable want to use it instead of introducing a constant of their own.

fintelia:
This sounds reasonable to me given that the constant comes from the PNG spec sections for the cHRM and gAMA chunks

PR #650, opened 2025-09-30
Make `ScaledFloat::SCALING` constant public

d-e-s-o:
Make the `ScaledFloat::SCALING` constant part of the public API, such that users can reference it instead of having to create their own version based on the value mentioned in the documentation.

Closes: #649

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -420,12 +420,12 @@ impl DeflateCompression {
 }
 
 /// An unsigned integer scaled version of a floating point value,
-/// equivalent to an integer quotient with fixed denominator (100_000)).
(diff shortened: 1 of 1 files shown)


d-e-s-o:
I am actually kind of tempted to make this a `u32` instead of a `f32`, as that seems a bit more apt (at least to me). I don't feel terribly strongly about it, though. Any thoughts?

fintelia:
Before checking I was initially assuming that the constant would be a `u32`, but I think using `f32` makes more sense.

32-bit floats can exactly represent all integers up to ~16 million so there aren't any accuracy issues. And if you do want to do arithmetic using the scale factor, you probably do actually need it as an `f32` so there'd be less casting that way.

PR #651, opened 2025-10-14
Ignore duplicate `acTL` chunks.

anforowicz:
PTAL?


197g (approved):
Thank you for the diagnosis and fix. LGTM, error is appropriate and consistent with the current handling strategy.

PR #653, opened 2025-11-18
Require IDAT before fdAT

stoeckmann:
Every PNG file requires at least one IDAT chunk, which is also true for animated PNG. With animated PNG, the IDAT chunk(s) must come before the first fdAT, according to PNG 3 standard (5.6 chunk ordering).

See https://www.w3.org/TR/png-3/#5ChunkOrdering for details.

Proof of Concept:

1. Create a PNG with two frames but only one fdAT and no IDAT
```
$ cat > poc.png.b64 << EOF
iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAACXBIWXMAAAABAAAAAQBPJcTWAAAA
CGFjVEwAAAACAAAAAYSKo+YAAAAaZmNUTAAAAAAAAAABAAAAAQAAAAAAAAAAAAEAHgAAQVA7fAAA
ABBmZEFUAAAAAXicY/z//z8ABgIC/1STQa4AAAAASUVORK5CYII=
EOF
$ base64 -d poc.png.b64 > poc.png
```

2. Open PNG in an application using image-png

You will see the fdAT content (1x1 image with a white pixel).

3. Check PNG with examples/pngcheck, which notices the issue

```
$ cargo run --package png --example pngcheck poc.png
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/examples/pngcheck poc.png`

    2 frames, 1 plays
    sequence #0, 1 x 1 pixels @ (0, 0)
    1/30 s delay, dispose: DISPOSE_OP_NONE, blend: BLEND_OP_SOURCEERROR: poc.png
poc.png: IDAT chunk missing
```

4. Open PNG in an application using libpng

The image is not shown due to an error.

5. Check with libpng's pngfix, which also notices the issue

```
$ pngfix poc.png
IEND ERR 00 libpng Success Success IEND:_out_of_place poc.png
```

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -1094,7 +1094,7 @@ impl StreamingDecoder {
             0
         });
         self.inflater.reset();
-        self.ready_for_fdat_chunks = true;
+        self.ready_for_fdat_chunks = self.have_idat;
         let fc = FrameControl {
             sequence_number: next_seq_no,
             width: buf.read_be()?,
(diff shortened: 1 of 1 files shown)


stoeckmann:
Update: I have added a unit test with given proof of concept file.

stoeckmann:
> Another case we should double check is that an fdAT before any IDATs should not cause decoding to fail

Why should that be allowed? According to standard, IDAT must occur before fdAT. Or is it a typo and you mean that it should fail and the test suite doesn't cover it?

PR #655, opened 2025-11-21
Implement PartialEq for EncodingError

wmedrano:
- I have a library where I would like to test the errors with expect_eq and deriving PartialEq would be convenient
- Did a manual implementation because the trait is not derivable due to IoError not having straightforward eq semantics.
- Feel free to reject the PR, there is some discussion at https://users.rust-lang.org/t/help-understanding-io-error-and-lack-of-partialeq/13212 for why io error equality doesn't make sense. In this PR, I just made it always false.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -923,12 +923,12 @@ impl Default for Transformations {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct ParameterError {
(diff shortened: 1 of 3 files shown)


fintelia:
I think reasoning behind not implementing `PartialEq` for `io::Error` probably also applies to `EncodingError`. For your use case, would calling `.to_string()` and comparing the resulting strings be sufficient?

PR #656, opened 2025-11-26
Only passing `available..` bytes of `UnfilterBuf` to `fdeflate`.

anforowicz:
PTAL?

This PR tries to support the work from https://github.com/image-rs/image-png/pull/640 (/cc @fintelia, @197g).  I wonder if the changes in this PR may help to avoid fuzzing failures reported in https://github.com/image-rs/image-png/pull/640#issuecomment-3253727444 (since passing an incorrect look-back buffer to fdeflate may be one hypothetical difference seen when reading byte-by-byte vs whole-content).

--------------------------

I am not sure if rebasing https://github.com/image-rs/image-png/pull/640 on top of this PR would address remaining concerns about that other PR.  I hope so?  :-)  At any rate, I think that merging this PR should still be desirable and be a step in the right direction.

If there are other remaining issues with https://github.com/image-rs/image-png/pull/640 (limited time available for continuing working on it?  functional or performance concerns?), then I may try to continue this work later.  At the same time, I find it a bit hard to fit all of the complexity of `UnfilteringBuffer` and `UnfilterBuf` into my brain all at once.  At the same time, I don't have any solid/constructive ideas, so maybe I just need to try harder.

One half-baked idea that I had was to:

* Rename `UnfilterBuf` into `DecompressorBufferMutRef`
    - Public API can remain unchanged by renaming in `lib.rs` via `pub use DecompressorBufferMutRef as UnfilterBuf` (I think this is ok, because `zlib` module is not public)
    - Alternatively we could also rename in the public API and introduce a `#[deprecated]` type alias to keep the old name available until 0.19
* Rename `UnfilterRegion` into `DecompressorBufferRegion` (same notes as above)
* Introduce an `DecompressorBuffer` struct with owned (i.e. non-ref) `buffer`, `available`, `filled`
* Somehow (/me waves hands) encapsulate managing the buffers into `DecompressorBuffer...`.  Maybe:
    - Replace `data_stream`, `filled`, and `available` fields of `UnfilteringBuffer` with a single `buf: DecompressorBuffer`
    - Move shifting code from `UnfilteringBuffer` into `DecompressorBuffer` (while also letting `UnfilteringBuffer` appropriately shift/managed `prev_start` and `current_start`).
    - Make the fields of `DecompressorBuffer` and `DecompressorBufferMutRef` private (exposing accessors if needed - e.g. maybe `fn available_bytes_mut(&mut self) -> &mut [u8]`)

But I am not sure if the above would really help, or if I would just be shuffling the chairs around.  (And in the process making the code more familiar and therefore maybe easier to understand for me [I am not even sure about that part :-)], but not necessarily for others.)

### DIFF
--- a/src/decoder/zlib.rs
+++ b/src/decoder/zlib.rs
@@ -211,6 +211,11 @@ impl UnfilterBuf<'_> {
         if decompressor.is_done() {
             *self.available = *self.filled;
         } else if let Some(new_available) = self.filled.checked_sub(Self::LOOKBACK_SIZE) {
+            // The decompressed data may have started in the middle of the buffer,
+            // so ensure that `self.available` never goes backward.  This is needed
+            // to avoid miscommunicating the size of the "look-back" window when calling
+            // `fdeflate::Decompressor::read` a bit earlier and passing
+            // `&mut self.buffer[*self.available..output_limit]`.
             if new_available > *self.available {
                 *self.available = new_available;
             }


src/decoder/zlib.rs
```diff
@@ -194,20 +170,65 @@ impl UnfilterRegion {
...
+            Some(input) => (input, false),
+            None => ([].as_slice(), true),
+        };
+
+        let output_limit = (*self.filled + UnfilteringBuffer::GROWTH_BYTES).min(self.buffer.len());
+        let (in_consumed, out_consumed) = decompressor
+            .read(
+                input,
+                &mut self.buffer[*self.available..output_limit],
+                *self.filled - *self.available,
```
197g, line 203:
```suggestion
                &mut self.buffer[..output_limit],
                *self.filled,
```

Avoiding the pivot here shows me quite some benchmarks improvements. I think it's the combined fact that 1) we avoid the read from `self.available` until we check if it needs to be updated, actually sometimes only write 2) the optimizers sees that `..output_limit` needs no bounds-check due to the `min` above but when we add an unbounded start the access could panic.

Was there a particular motivation for this change?

src/decoder/zlib.rs
```diff
@@ -194,20 +170,65 @@ impl UnfilterRegion {
...
+                DecodingError::Format(FormatErrorInner::CorruptFlateStream { err }.into())
+            })?;
+
+        *self.filled += out_consumed;
+        if decompressor.is_done() {
+            *self.available = *self.filled;
+        } else if let Some(new_available) = self.filled.checked_sub(Self::LOOKBACK_SIZE) {
+            if new_available > *self.available {
+                *self.available = new_available;
+            }
```
197g, line 221:
```suggestion
            // The zlib stream may have started in the middle of the buffer, but even
            // then we must not cover previously made available data.
            *self.available = new_available.max(*self.available);
```

I'd find this more readable so unless it has performance implications. The comment would be much appreciated either way.

197g (changes requested):
I'm growing more fond of the approach the deeper I look. Decompression becomes much cleaner, the `grow_buffer_if_needed_for_final_flushing` is just much better. (And rewards us with a good bunch of perf sometimes, I think?)

There are a few remarks I'd like to see discussed. We are going to go this direction. All the extra documentation is greatly appreciated.

src/decoder/zlib.rs
anforowicz: From the commit message:

> Before this commit, `UnsafeBuf::decompress` would pass
> `&mut self.buffer[..output_limit]` to `fdeflate`.  After this commit it
> instead passes `&mut self.buffer[*self.available..output_limit`.  This
> helps to support future usage of `UnsafeBuf` where the `0..available`
> bytes may not necessarily represent a valid "look back" window.
> In particular, this should support the suggestion made here:
> https://github.com/image-rs/image-png/pull/640#issuecomment-3283518271

src/decoder/zlib.rs
anforowicz: Done.

src/decoder/zlib.rs
anforowicz: > Avoiding the pivot here shows me quite some benchmarks improvements.

FWIW I tried running the benchmarks and it seemed that there is no change.  (I note that some benchmarks are rather noisy - I did see a handful of supposed regressions after a single run, but rerunning those particular benchmarks didn't repro the regression.)

197g: Ah thanks, that slipped by.

I do get a consistent 1.5% on AMD Ryzen 7900X but not on all benchmarks. And decently little noise by running everything via `taskset -c 1` and the right scheduler selection; enough for me to even trust some noise below what criterion shows them as statistically significant and those benchmarks clear that statistical threshold clearly.

Alas, if it's not widespread I appreciate the code clarity and consistency above such a small number. Moving it into the `UnfilterBuf` (sic! commit message says `UnsafeBuf) we should be able to make use of that invariant better, will get a fixup if need be.

PR #657, opened 2025-11-28
Allow seeking back to read skipped chunks

fintelia:
With #630 in place, we now have most of the pieces to load PNG metadata without consuming an unbounded amount of space.

The key here is that we track just the start position of each unbounded-size chunk we might care about. Which means that we can track thousands of text chunks with even a tiny space limit. When it comes time to read the chunks, we can seek back to them and provide the desired info. But if the chunk info is never requested then we don't pay the overhead of storing them. When in recording-mode, I believe that the only non-constant memory usage should be recording text chunk positions and reserving two rows of space for the unfiltering buffer.

This is particularly useful for the `image` crate which needs to know whether there's a tRNS chunk (to determine whether the color type is RGB or RGBA) before it knows whether the EXIF or ICC profile will be requested. By doing things this way, it neatly sidesteps the question of setting decoding limits before reading metadata. The main thing decoding limits were being used for was prevent zip-bombs in the compressed iCCP chunks and to a lesser degree preventing unreasonably sized EXIF or text chunks.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -467,8 +467,10 @@ pub struct DecodeOptions {
     ignore_adler32: bool,
     ignore_crc: bool,
     ignore_text_chunk: bool,
+    ignore_exif_chunk: bool,
     ignore_iccp_chunk: bool,
     skip_ancillary_crc_failures: bool,
+    record_skipped_chunks: bool,
 }
 
 impl Default for DecodeOptions {
@@ -477,8 +479,10 @@ impl Default for DecodeOptions {
             ignore_adler32: true,
             ignore_crc: false,
             ignore_text_chunk: false,
+            ignore_exif_chunk: false,
             ignore_iccp_chunk: false,
             skip_ancillary_crc_failures: true,
+            record_skipped_chunks: false,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/decoder/stream.rs
```diff
@@ -1074,6 +1128,33 @@ impl StreamingDecoder {
...
+        match self.current_chunk.type_ {
+            chunk::tEXt | chunk::zTXt | chunk::iTXt => {
+                if self.limits.reserve_bytes(8).is_ok() {
+                    self.text_chunk_positions.push(chunk_start);
+                }
+            }
+            chunk::eXIf => {
+                if self.exif_position.is_none() {
+                    self.exif_position = Some(chunk_start);
+                }
```
197g, line 1147: The error treatment here differs from parsing the `eXIf` and `iCCP` chunk. In the latter case we error first with a `Format` and then later, probably, downgrade that to `BadAncillaryChunk` in most cases. Here however we do nothing.

src/decoder/stream.rs
```diff
@@ -1074,6 +1128,33 @@ impl StreamingDecoder {
...
 
+    fn record_ancillary_chunk(&mut self, stream_position: u64) {
+        if !self.decode_options.record_skipped_chunks {
+            return;
+        }
+
+        let chunk_start = stream_position - 8 - self.current_chunk.raw_bytes.len() as u64;
+
+        match self.current_chunk.type_ {
+            chunk::tEXt | chunk::zTXt | chunk::iTXt => {
```
197g, line 1139: Wouldn't we still want to know the exact kind of text chunk? `zTXt` is handled very different from the other chunks. We can of course parse it back when we seek back and re-read the chunk anew but it's fixed-size metadata—so we might as well.

src/decoder/stream.rs
197g: Also do we consider the option _changing_ through `set_ignore_exif_chunk` while it is read, i.e. first reading one fully and then only recording a position would still be a duplicate chunk. That may not happen but without the interface really assuring it we should know if that is in-scope of the implementation or not.

src/decoder/stream.rs
fintelia: Yeah, I wasn't sure whether to save the chunk size/type or just re-read it when requested. Easy enough to track it alongside the chunk start position.

PR #662, opened 2025-12-08
Have UnfilteringBuffer track number of remaining bytes in the frame

### LINKED ISSUE #605: CorruptFlateStream { err: InsufficientInput } (works elsewhere)
This is loading fine with GdkPixbuf and Firefox

Image

from https://ip.bieringer.net/cgn-test.html

```rust
Format(FormatError { inner: CorruptFlateStream { err: InsufficientInput } })
```

fintelia:
This is an attempt to fix #605 by tracking precisely how many bytes there are left in a frame so that we don't decompress more zlib data than we need.

~~Doesn't fully fix the `buf_independent` fuzz test, but gets us closer. Will investigate more when I have a chance...~~ Fuzzing seems to pass now.

### DIFF
--- a/src/decoder/zlib.rs
+++ b/src/decoder/zlib.rs
@@ -133,11 +133,21 @@ impl ZlibStream {
     /// Returns `Ok(true)` if all data has been decompressed and there is no
     /// more data that will be produced, or `Ok(false)` if there's potentially
     /// more output.
+    ///
+    /// Returns `Err` if the zlib stream is corrupt or truncated too early.
     pub(crate) fn finish(
         &mut self,
         image_data: &mut UnfilterBuf<'_>,
     ) -> Result<bool, DecodingError> {
-        if !self.started || self.state.is_done() || *image_data.filled == image_data.buffer.len() {
+        if !self.started || self.state.is_done() {
+            return Ok(true);
+        }
+
+        // If the zlib stream isn't done but we've already output all the pixel
+        // data needed, then either there's too much compressed data or the
+        // checksum is missing. Those aren't allowed by the spec, but libpng
+        // generally doesn't treat them as fatal.
+        if *image_data.filled == image_data.buffer.len() {
             return Ok(true);
         }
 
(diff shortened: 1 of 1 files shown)


src/decoder/zlib.rs
```diff
@@ -124,29 +124,23 @@ impl ZlibStream {
...
     /// within the decompressor and flushing additional data which may have been kept back in case
     /// more data were passed to it.
-    pub(crate) fn finish_compressed_chunks(
+    pub(crate) fn finish(
         &mut self,
         image_data: &mut UnfilterBuf<'_>,
-    ) -> Result<(), DecodingError> {
-        if !self.started {
-            return Ok(());
+    ) -> Result<bool, DecodingError> {
```
anforowicz, line 141: Can you please describe the meaning of `bool` in the doc comment?  e.g.:

```
///
/// Returns `Ok(true)` if the end of the zlib stream has been reached.
/// Returns `Ok(false)` if the zlib stream can produce more output.
/// Returns `Err(_)` if the zlib stream is corrupted/malformed.
```

anforowicz (approved):

src/decoder/zlib.rs
```diff
@@ -210,6 +204,13 @@ impl UnfilterBuf<'_> {
         *self.filled += out_consumed;
         if decompressor.is_done() {
             *self.available = *self.filled;
+        } else if end_of_input && out_consumed == 0 {
+            return Err(DecodingError::Format(
+                FormatErrorInner::CorruptFlateStream {
+                    err: fdeflate::DecompressionError::InsufficientInput,
+                }
+                .into(),
+            ));
```
197g: I'm not sure I understand the implications of this. If we pass an empty `buffer` then surely `out_consumed = 0`. But that doesn't really say that we were or were not actually expecting any data? Previously there'd always be some buffer available as long as we passed an `UnfilterBuf` but now that is—at least superficially—no longer true.

197g: As a note, in #664 passes the unfilter buffer in more situations that seem to also include after we have all the actual image data. In fact the argument is no longer optional. It's just _super confusing_ in combination.

anforowicz: I think this is fine - I don't think there are any cases where the output buffer would be full without decompressing all the raw rows?

There is a minor question of what exact error should be returned here (maybe `FormatErrorInner::NoMoreImageData` would be more appropriate), but it probably doesn't matter that much, because error details are an internal implementation detail.

anforowicz: > As a note, in #664 passes the unfilter buffer in more situations that seem to also include after we have all the actual image data. In fact the argument is no longer optional. It's just _super confusing_ in combination.

After _this_ PR lands, then #664 can (and should) revert those particular changes.

fintelia:
I've thought about the `buf_independent` test some more, and I think there's still and additional case this PR doesn't handle. The problem is when the final bytes of the zlib stream contains an error after all of the pixel data. Decoding byte-by-byte will complete the `next_frame` call successfully, while the all-at-once decoder will notice the error in the same operation as reading the final row of pixels.

Though honestly the best fix might be to just adjust `buf_independent` to call both `next_frame` and `finish` consecutively, and not distinguish which one returns the error. (We'd still need to make sure that `finish` surfaced any errors that were skipped, but that would be less complicated/invasive than other alternatives)

src/decoder/zlib.rs
fintelia: Ok, I refactored this code some and added a bunch more comments. Hopefully it should be clearer now.

anforowicz:
@fintelia / @197g - are there any remaining concerns or action items for this PR?  Merging this PR will unblock/simplify the changes in another PR at https://github.com/image-rs/image-png/pull/664

PR #590, opened 2025-06-02
Unfilter buffer for decoded stream

197g:
An alternative to #589 where we preserve having an argument in `StreamDecoder::update`, so roughly keeping the previous API intact. Instead, the internals for the `UnfilterBuffer` are modified such that a reference to it and its state tracking are being passed into `zlib`. The interface is even kept flexible in the sense that it does not know the details of the required preserved data length (which it must not modify in `unfilter`, I did that on accident and that was a hard bug to discover :smiley:).

As an additional tweaking parameter, we pass the maximum rowlength (in bytes) to that buffer. This way we can pre-allocate and tune the resizing steps. I think I've improperly tuned for very large widths for now (see performance logs) but this additional information should enable serving all sizes more reliably.

<details>
<summary>i7-8565U / 16GB</summary>

```
decode/Lohengrin_-_Illustrated_Sporting_and_Dramatic_News.png
                        time:   [53.514 ms 56.304 ms 58.575 ms]
                        thrpt:  [182.31 MiB/s 189.66 MiB/s 199.55 MiB/s]
                 change:
                        time:   [-0.9990% +3.4599% +7.5301%] (p = 0.16 > 0.05)
                        thrpt:  [-7.0028% -3.3442% +1.0091%]
                        No change in performance detected.
decode/Transparency.png time:   [208.74 µs 209.25 µs 209.94 µs]
                        thrpt:  [1.5970 GiB/s 1.6023 GiB/s 1.6062 GiB/s]
                 change:
                        time:   [-20.036% -19.583% -19.174%] (p = 0.00 < 0.05)
                        thrpt:  [+23.723% +24.352% +25.057%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
decode/kodim02.png      time:   [7.1647 ms 7.1879 ms 7.2262 ms]
                        thrpt:  [155.68 MiB/s 156.51 MiB/s 157.02 MiB/s]
                 change:
                        time:   [-4.0359% -3.5946% -3.1558%] (p = 0.00 < 0.05)
                        thrpt:  [+3.2587% +3.7286% +4.2057%]
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) high mild
decode/kodim07.png      time:   [8.9415 ms 8.9682 ms 9.0092 ms]
                        thrpt:  [124.87 MiB/s 125.44 MiB/s 125.82 MiB/s]
                 change:
[...]

PR #664, opened 2025-12-10
Allow full unfiltering for partial data (`enum`-based approach)

anforowicz:
Hello!

This PR fixes https://github.com/image-rs/image-png/issues/639.   This is an alternative approach to @197g's https://github.com/image-rs/image-png/pull/640 - the two main differences are:

* `prev_start: usize` is replaced with `prev_row: PrevRow` where `PrevRow` is an `enum` with separate `None`, `InPlace`, and `Scratch` variants.  I think this results in easier to read code then imposing a special meaning on certain values of offsets (e.g. `self.prev_start == self.curr_start` => no previous row;  or having `self.prev_start` sometimes trail `self.curr_start` by `rowlen-1` and sometimes point to mutable scratch space in the same buffer as `curr_start`).
* There is a test and a fix for consistently detecting errors in the tail portion of zlib stream (see the discussion at https://github.com/image-rs/image-png/pull/640#discussion_r2594273419).  This part may conflict with https://github.com/image-rs/image-png/pull/662 where @fintelia proposes to always ignore the tail zlib bytes (rather than as in this PR, to always decode them to detect errors).  I think I should wait until @fintelia's PR lands, and then rebase on top of it (removing changes in _this_ PR that won't be necessary afterwards).

PTAL?

### DIFF
--- a/src/decoder/unfiltering_buffer.rs
+++ b/src/decoder/unfiltering_buffer.rs
@@ -8,12 +8,7 @@ use crate::Info;
 pub(crate) struct UnfilteringBuffer {
     /// Vec containing the uncompressed image data currently being processed.
     data_stream: Vec<u8>,
-    /// Index in `data_stream` where the previous row starts.
-    /// This excludes the filter type byte - it points at the first byte of actual pixel data.
-    /// The pixel data is already-`unfilter`-ed.
-    ///
-    /// If `prev_start == current_start` then it means that there is no previous row.
-    prev_start: usize,
+    prev_row: PrevRow,
     /// Index in `data_stream` where the current row starts.
     /// This points at the filter type byte of the current row (i.e. the actual pixel data starts at `current_start + 1`)
     /// The pixel data is not-yet-`unfilter`-ed.
@@ -28,6 +23,8 @@ pub(crate) struct UnfilteringBuffer {
     shift_back_limit: usize,
     /// How many bytes are left to decompress into this buffer for the current frame.
     remaining_bytes: u64,
+    /// To avoid always allocating a new vector in `fn unfilter_curr_row_using_scratch_buffer`.
+    scratch_buffer: Vec<u8>,
 }
 
 impl UnfilteringBuffer {
@@ -36,9 +33,14 @@ impl UnfilteringBuffer {
     /// Asserts in debug builds that all the invariants hold.  No-op in release
     /// builds.  Intended to be called after creating or mutating `self` to
     /// ensure that the final state preserves the invariants.
+    #[cfg(not(debug_assertions))]
+    fn debug_assert_invariants(&self) {}
+    #[cfg(debug_assertions)]
     fn debug_assert_invariants(&self) {
-        debug_assert!(self.prev_start <= self.current_start);
-        debug_assert!(self.current_start <= self.available);
+        if let PrevRow::InPlace(prev_start) = &self.prev_row {
+            debug_assert!(*prev_start <= self.current_start);
+        }
+        debug_assert!(self.current_start <= self.filled);
         debug_assert!(self.available <= self.filled);
         debug_assert!(self.filled <= self.data_stream.len());
     }
@@ -89,12 +91,13 @@ impl UnfilteringBuffer {
 
         let result = Self {
             data_stream: Vec::with_capacity(data_stream_capacity),
-            prev_start: 0,
+            prev_row: PrevRow::None,
             current_start: 0,
             filled: 0,
             available: 0,
             shift_back_limit,
             remaining_bytes: u64::MAX,
+            scratch_buffer: Vec::new(),
         };
 
         result.debug_assert_invariants();
@@ -104,13 +107,19 @@ impl UnfilteringBuffer {
     /// Called to indicate that there is no previous row (e.g. when the current
     /// row is the first scanline of a given Adam7 pass).
     pub fn reset_prev_row(&mut self) {
-        self.prev_start = self.current_start;
+        // Stash a previously allocated buffer (for potential reuse later)
+        // rather than throwing it away when resetting `self.prev_row`.
+        if let PrevRow::Scratch(buf) = &mut self.prev_row {
+            self.scratch_buffer = std::mem::take(buf);
+        }
+
+        self.prev_row = PrevRow::None;
         self.debug_assert_invariants();
     }
 
     pub fn start_frame(&mut self, frame_bytes: u64) {
         self.data_stream.clear();
-        self.prev_start = 0;
+        self.prev_row = PrevRow::None;
         self.current_start = 0;
         self.filled = 0;
         self.available = 0;
@@ -123,12 +132,28 @@ impl UnfilteringBuffer {
 
     /// Returns the previous (already `unfilter`-ed) row.
     pub fn prev_row(&self) -> &[u8] {
-        &self.data_stream[self.prev_start..self.current_start]
+        self.prev_row
+            .as_slice(&self.data_stream[..self.current_start])
     }
 
-    /// Returns how many bytes of the current row are present in the buffer.
-    pub fn curr_row_len(&self) -> usize {
-        self.available - self.current_start
+    /// Returns how many bytes of the current row are present in the mutable
+    /// part of the buffer (32kB most recently decompressed bytes are read-only
+    /// to retain the "lookback" window as needed for inflate algorithm).  If a
+    /// full row is mutable, then it may be unfiltered using
+    /// `unfilter_curr_row_in_place`.
+    ///
+    /// See also `readable_len_of_curr_row`.
+    pub fn mutable_len_of_curr_row(&self) -> usize {
+        self.available.saturating_sub(self.current_start)
+    }
+
+    /// Returns how many bytes of the current row have been already
+    /// decompressed.  If a full row is available, then it may be unfiltered
+    /// using `unfilter_curr_row_using_scratch_buffer`.
+    ///
+    /// See also `mutable_len_of_curr_row`.
+    pub fn readable_len_of_curr_row(&self) -> usize {
+        self.filled - self.current_start
     }
 
     /// Runs `f` on the underlying buffer.
@@ -143,7 +168,11 @@ impl UnfilteringBuffer {
         F: FnOnce(&mut UnfilterBuf<'_>) -> T,
     {
         // Potentially shift the buffer left to avoid unbounded growth.
-        if self.prev_start >= self.shift_back_limit
+        let discard_size = self.available.min(match &self.prev_row {
+            PrevRow::None | PrevRow::Scratch(_) => self.current_start,
+            PrevRow::InPlace(prev_start) => *prev_start,
+        });
+        if discard_size >= self.shift_back_limit
             // Avoid the shift back if the buffer is still very empty. Consider how we got here: a
             // previous decompression filled the buffer, then we unfiltered, we're now refilling
             // the buffer again. The condition implies, the previous decompression filled at most
@@ -151,7 +180,7 @@ impl UnfilteringBuffer {
             // attempt will not yet be limited by the buffer length.
             && self.filled >= self.data_stream.len() / 2
         {
-            self.shift_buffer_left(self.prev_start);
+            self.shift_buffer_left(discard_size);
         }
 
         if self.filled + Self::GROWTH_BYTES > self.data_stream.len() {
@@ -206,7 +235,10 @@ impl UnfilteringBuffer {
         self.current_start -= discard_size;
         self.available -= discard_size;
         self.filled -= discard_size;
-        self.prev_start -= discard_size;
+        match &mut self.prev_row {
+            PrevRow::None | PrevRow::Scratch(_) => (),
+            PrevRow::InPlace(prev_start) => *prev_start -= discard_size,
+        }
     }
 
     fn curr_row_filter(&self) -> Result<RowFilter, DecodingError> {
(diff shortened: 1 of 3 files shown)


src/decoder/unfiltering_buffer.rs
```diff
@@ -204,6 +295,45 @@ impl UnfilteringBuffer {
...
+    /// `buf` should refer to the `..current_start` portion of
+    /// `UnfilteringBuffer::data_stream`.
+    fn as_slice<'a>(&'a self, buf: &'a [u8]) -> &'a [u8] {
+        match self {
+            PrevRow::None => &[],
+            PrevRow::InPlace(prev_start) => &buf[*prev_start..],
+            PrevRow::Scratch(buf) => buf.as_slice(),
+        }
+    }
+}
```
197g, line 362: > micro optimization is the root of all evil

"Yet we should not pass up our opportunities in that critical 3%", to complete the quote. There's a bit of cost I'm can accept to accommodate best-efforts on broken images and incremental encoding but given that this isn't the use case for a lot of our other users, not too much. This shouldn't even really appear in the executed code path for any of our benchmarks that supply the full images in memory.

I'm also sorry to say it is far from micro on non-compressed streams in so far as I can tell. Granted the majority is rather neutral but this is all end-to-end and the zlib decompression masks a _lot_ of other costs when it is involved.

<details>

```
Benchmarking decode/tango-icon-address-book-new-16.png
Benchmarking decode/tango-icon-address-book-new-16.png: Warming up for 3.0000 s
Benchmarking decode/tango-icon-address-book-new-16.png: Collecting 10 samples in estimated 5.0001 s (1.1M iterations)
Benchmarking decode/tango-icon-address-book-new-16.png: Analyzing
decode/tango-icon-address-book-new-16.png
                        time:   [4.5412 µs 4.5419 µs 4.5426 µs]
                        thrpt:  [214.98 MiB/s 215.01 MiB/s 215.04 MiB/s]
                 change:
                        time:   [−1.6796% −1.6476% −1.6182%] (p = 0.00 < 0.05)
                        thrpt:  [+1.6449% +1.6752% +1.7083%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low mild
Benchmarking decode/tango-icon-address-book-new-32.png
Benchmarking decode/tango-icon-address-book-new-32.png: Warming up for 3.0000 s
Benchmarking decode/tango-icon-address-book-new-32.png: Collecting 10 samples in estimated 5.0002 s (652k iterations)
Benchmarking decode/tango-icon-address-book-new-32.png: Analyzing
decode/tango-icon-address-book-new-32.png
                        time:   [7.6328 µs 7.6330 µs 7.6333 µs]
                        thrpt:  [511.74 MiB/s 511.76 MiB/s 511.77 MiB/s]
                 change:
                        time:   [−3.8710% −3.8472% −3.8219%] (p = 0.00 < 0.05)
                        thrpt:  [+3.9738% +4.0011% +4.0269%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild

Benchmarking generated-noncompressed-4k-idat/8x8.png
Benchmarking generated-noncompressed-4k-idat/8x8.png: Warming up for 3.0000 s
Benchmarking generated-noncompressed-4k-idat/8x8.png: Collecting 100 samples in estimated 5.0024 s (7.3M iterations)
Benchmarking generated-noncompressed-4k-idat/8x8.png: Analyzing
generated-noncompressed-4k-idat/8x8.png
                        time:   [687.55 ns 687.70 ns 687.86 ns]
                        thrpt:  [354.93 MiB/s 355.01 MiB/s 355.09 MiB/s]
                 change:
                        time:   [−6.9506% −6.8972% −6.8453%] (p = 0.00 < 0.05)
                        thrpt:  [+7.3483% +7.4082% +7.4698%]
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
Benchmarking generated-noncompressed-64k-idat/128x128.png
Benchmarking generated-noncompressed-64k-idat/128x128.png: Warming up for 3.0000 s
Benchmarking generated-noncompressed-64k-idat/128x128.png: Collecting 100 samples in estimated 5.0117 s (672k iterations)
Benchmarking generated-noncompressed-64k-idat/128x128.png: Analyzing
generated-noncompressed-64k-idat/128x128.png
                        time:   [7.4723 µs 7.4730 µs 7.4737 µs]
                        thrpt:  [8.1667 GiB/s 8.1674 GiB/s 8.1682 GiB/s]
                 change:
                        time:   [−3.8230% −3.7973% −3.7721%] (p = 0.00 < 0.05)
                        thrpt:  [+3.9199% +3.9472% +3.9750%]
                        Performance has improved.
  1 (1.00%) low severe
  1 (1.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe
Benchmarking generated-noncompressed-64k-idat/2048x2048.png
Benchmarking generated-noncompressed-64k-idat/2048x2048.png: Warming up for 3.0000 s
Benchmarking generated-noncompressed-64k-idat/2048x2048.png: Collecting 10 samples in estimated 5.0242 s (3190 iterations)
Benchmarking generated-noncompressed-64k-idat/2048x2048.png: Analyzing
generated-noncompressed-64k-idat/2048x2048.png
                        time:   [1.5710 ms 1.5718 ms 1.5727 ms]
                        thrpt:  [9.9354 GiB/s 9.9410 GiB/s 9.9459 GiB/s]
                 change:
                        time:   [−12.291% −11.690% −11.025%] (p = 0.00 < 0.05)
                        thrpt:  [+12.391% +13.237% +14.013%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
Benchmarking generated-noncompressed-2g-idat/2048x2048.png
Benchmarking generated-noncompressed-2g-idat/2048x2048.png: Warming up for 3.0000 s
Benchmarking generated-noncompressed-2g-idat/2048x2048.png: Collecting 10 samples in estimated 5.0518 s (3080 iterations)
Benchmarking generated-noncompressed-2g-idat/2048x2048.png: Analyzing
generated-noncompressed-2g-idat/2048x2048.png
                        time:   [1.6381 ms 1.6393 ms 1.6410 ms]
                        thrpt:  [9.5217 GiB/s 9.5314 GiB/s 9.5382 GiB/s]
                 change:
                        time:   [−2.3397% −1.9448% −1.5494%] (p = 0.00 < 0.05)
                        thrpt:  [+1.5738% +1.9833% +2.3957%]
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
Benchmarking row-by-row/128x128-4k-idat
Benchmarking row-by-row/128x128-4k-idat: Warming up for 3.0000 s
Benchmarking row-by-row/128x128-4k-idat: Collecting 100 samples in estimated 5.0212 s (606k iterations)
Benchmarking row-by-row/128x128-4k-idat: Analyzing
row-by-row/128x128-4k-idat
                        time:   [8.2700 µs 8.2706 µs 8.2712 µs]
                        thrpt:  [7.3792 GiB/s 7.3798 GiB/s 7.3803 GiB/s]
                 change:
                        time:   [−11.145% −11.125% −11.105%] (p = 0.00 < 0.05)
                        thrpt:  [+12.493% +12.518% +12.543%]
                        Performance has improved.
  2 (2.00%) high mild
  6 (6.00%) high severe
```

</details>

197g (commented):
Maybe the fuzzer finding of panic in zlib is fixed by a rebase on #662 although I don't understand how it is getting triggered in the first place. It'd be ironic if the enum variants introduced for following the state actually lead to more possible state corruption.

src/decoder/unfiltering_buffer.rs
```diff
@@ -200,30 +211,109 @@ impl UnfilteringBuffer {
         ret
     }
 
-    /// Runs `unfilter` on the current row, and then shifts rows so that the current row becomes the previous row.
+    /// Shifts the contents of `self.data_stream` left,
+    /// discarding the first `discard_size` bytes.
+    fn shift_buffer_left(&mut self, discard_size: usize) {
+        // Violating this assertion will clobber the immutable "lookback"
+        // window that needs to be maintained for decompresor.
```
telecos: Nit: decompressor

src/decoder/unfiltering_buffer.rs
```diff
@@ -200,30 +211,109 @@ impl UnfilteringBuffer {
...
+    /// Before running `unfilter`, the contents of the current row will be
+    /// copied into a scratch buffer.  This allows unfiltering to happen even
+    /// if `mutable_len_of_curr_row < rowlen` (e.g. when handling partial
+    /// or not-yet-complete input streams).
+    pub fn unfilter_curr_row_using_scratch_buffer(
+        &mut self,
+        rowlen: usize,
+        bpp: BytesPerPixel,
+    ) -> Result<(), DecodingError> {
+        debug_assert!(rowlen >= 2); // 1 byte for `FilterType` and at least 1 byte of pixel data.
```
telecos: Nit: `RowFilter` for consistency with comment from L262?

src/decoder/unfiltering_buffer.rs
```diff
@@ -232,6 +322,45 @@ impl UnfilteringBuffer {
...
+impl PrevRow {
+    /// Returns the previous unfiltered row as a slice of bytes.
+    ///
+    /// `buf` should refer to the `..current_start` portion of
+    /// `UnfilteringBuffer::data_stream`.
+    fn as_slice<'a>(&'a self, buf: &'a [u8]) -> &'a [u8] {
+        match self {
+            PrevRow::None => &[],
+            PrevRow::InPlace(prev_start) => &buf[*prev_start..],
+            PrevRow::Scratch(buf) => buf.as_slice(),
```
telecos: Nit: buf parameter shadowing. Rename for clarity?

src/decoder/unfiltering_buffer.rs
anforowicz: Good catch - thanks!  The actual enum is indeed called `RowFilter`: https://github.com/image-rs/image-png/blob/3c577ede1c85a6ae059832dfcdd77e926b2ec96e/src/filter/mod.rs#L62-L70

src/decoder/unfiltering_buffer.rs
anforowicz: Good point.  Done.

src/decoder/unfiltering_buffer.rs
anforowicz: Done

anforowicz:
> On the QOI bench corpus measured using [`corpus-bench`](https://github.com/fintelia/corpus-bench/tree/png-0.18.0), I see an approximately 1% slowdown. Even though Chromium is blocked on this PR, I'm not sure that's a tolerable cost.

The functional regression was introduced in https://github.com/image-rs/image-png/pull/590 which (based on Ryzen results mentioned in https://github.com/image-rs/image-png/pull/590#issuecomment-2939603571) resulted in an average performance improvement of -2.94% ((1.1781 - 14.535 - 1.1326 - 1.3190 - 0.7238 - 2.0269 - 2.0258) / 7).  IMO we should accept losing some of that performance improvement, because it stems from a PR that introduced a functional regression.  (In theory, we could get a big performance improvement if we always decoded to all-black-pixels regardless of input, but this would be an obvious functional regression.  The functional regression fixed by this PR is more subtle and before this PR lacked automated regression tests, but I think the same high-level reasoning should apply.)

Shnatsel:
We should cut a new release to make this change easier for Chromium to pick up. I'll run a regression test on my corpora and then open a release PR.

fintelia:
I should probably add that I'm pretty angry about [this fiasco](https://issues.chromium.org/issues/477318798) and don't plan on collaborating/helping the Chromium folks going forward.

197g:
Sigh. Not a week in FOSS without license drama. I'm also disappointed to find out the license wasn't checked during the trial period. Quitting collaboration on the entire staff though, that seems a disproportionate response though to what a regex with the right intent. I would appreciate an explanation on how they intend to verify licenses better to avoid repeat occurrence of similar issues going forward.

fintelia:
It isn't just the regex. It is also the lack of acknowledgement/apology. It is choosing not to fast-track the correction into last week's release. It is reaching out privately asking when PRs might land. It is seeing Google's treatment of ffmpeg and libxml2. And it is taking a step back and wondering why I'm providing free labor for the benefit of multiple multi-trillion dollar corporations.

All that said, I don't have any desire to impose my feelings on the other community members here. If others want to continue collaborating, they should feel free to do so. It really is awesome to see that Chromium has a memory-safe PNG decoder

anforowicz:
> I would appreciate an explanation on how they intend to verify licenses better to avoid repeat occurrence of similar issues going forward.

The test added in https://crrev.com/c/7514149/4/components/resources/about_ui_credits_unittests.cc added verification of the contents of the UI text that is displayed in chrome://credits.

> It is also the lack of acknowledgement/apology.

I apologize for the lack of mention of `png` and other Rust crates in Chromium's chrome://credits page.  I don't have a good explanation for why this happened - I think we just didn't think about testing this particular aspect of adding Rust toolchain and crates to Chromium.  We assumed that the info in `README.chromium` files flows into all the right places and didn't think what specific places we should manually verify.

> It is choosing not to fast-track the correction into last week's release.

Chromium ships a new milestone to the Stable channel around every 4 weeks.  Stable respins are avoided and typically reserved for security issues - this is done to minimize the impact to end users (download bandwidth, having to restart the browser) and this way ensuring that updates that get shipped to the Stable channel get downloaded and applied quickly.

That said, this is not a valid excuse for not considering a merge to the M145 Beta channel.  It was fairly late in the M145 release cycle, but a merge was probably still possible at the end of January.  I'll try to do better going forward.

anforowicz:
> We should cut a new release to make this change easier for Chromium to pick up. I'll run a regression test on my corpora and then open a release PR.

Thank you for publishing the new version - it has now been imported into Chromium (starting with version [147.0.7701.0](https://chromiumdash.appspot.com/commit/322252102f422b266f1e8a3dabcee16ed2707f60)).

PR #665, opened 2025-12-19
Fix, improve, and add intra-doc links in doc comments

noritada:
This PR fixes, improves, and adds intra-doc links in doc comments.

Most of the changes are minor and do not involve alterations to the text.
There are only 2 changes to the text.

One is as follows.
I searched the repository but couldn't find any method or field named `Reader::frames` throughout the entire history,
so I made the change based on my own speculation.
Please let me know if it's wrong.

```
     /// Returns the color type and the number of bits per sample
-    /// of the data returned by [`Reader::next_row`] and Reader::frames`.
+    /// of the data returned by [`Reader::next_row`] and [`Reader::next_frame`].
     pub fn output_color_type(&self) -> (ColorType, BitDepth) {
```

Another one is as follows.

```
     Fastest,
:
     ///
-    /// Like `Compression::Fast` this can currently produce files larger than `NoCompression` in
+    /// Like [`Compression::Fastest`] this can currently produce files larger than `NoCompression` in
     /// streaming mode when given incompressible data. This may change in the future.
     Fast,
```

Thanks!

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -338,7 +338,7 @@ pub enum Compression {
     /// implementation tuned for PNG](https://crates.io/crates/fdeflate), while still providing
     /// better compression ratio than the fastest modes of other encoders.
     ///
-    /// Like `Compression::Fast` this can currently produce files larger than `NoCompression` in
+    /// Like [`Compression::Fastest`] this can currently produce files larger than `NoCompression` in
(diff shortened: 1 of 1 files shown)


fintelia (approved):

fintelia:
Thanks!

PR #670, opened 2026-01-17
Add `encoder` and `decoder` features

linkmauve:
This allows applications which only want to decode PNGs, or to encode PNGs, to not embed code for the direction they don’t need, with a compile-time error if neither are enabled.

On my i5-8350U, in release mode, current master builds in 1.55s, with only the `decoder` feature it builds in 1.13s, and with only the `encoder` feature in 0.68s.

On my iBook G4, still in release mode, the improvements are much more dramatic, a full build of `png` that would previously take 2:13 and 11 units, goes down to 1:16 and 8 units with only the `decoder` feature enabled, since it doesn’t build the `flate2` crate and its dependencies.  With just the `encoder` feature it takes 1:40, which is also much better on this slow CPU. :)

I’ve also cleaned up extra imports already present in the prelude since edition 2021, and removed the now-indirect `miniz_oxide` dependency.

This PR is best reviewed commit-per-commit.

Edit: I still need help to get the docstring tests to only run if the relevant feature is enabled, I couldn’t figure out how to do that so far.

Edit 2: Do we want to also test for only-encoder and only-decoder configurations now, and disable no-default-features checks as that means nothing?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         rust: ["1.73.0", stable, beta, nightly]
         os: [ubuntu-latest, windows-latest, macos-latest]
-        features: [""]
+        features: ["", "decoder", "encoder", "decoder,encoder"]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
@@ -32,7 +32,7 @@ jobs:
   feature_check:
     strategy:
       matrix:
-        features: ["", "unstable", "benchmarks"]
+        features: ["", "decoder", "encoder", "decoder,encoder", "decoder,encoder,unstable", "decoder,encoder,benchmarks"]
         os: [ubuntu-latest, macos-latest] # macos-latest is ARM
     runs-on: ${{ matrix.os }}
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,8 @@ rand = "0.9.2"
 criterion = "0.7.0"
 
 [features]
-default = ["encoder"]
+default = ["decoder", "encoder"]
+decoder = []
 encoder = ["flate2"]
 # Use nightly-only features for a minor performance boost in PNG decoding
 unstable = ["crc32fast/nightly"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ bitflags = "2.0"
 crc32fast = "1.2.0"
 fdeflate = "0.3.3"
 flate2 = { version = "1.0.35", optional = true }
-miniz_oxide = { version = "0.8", features = ["simd"] }
 
 [dev-dependencies]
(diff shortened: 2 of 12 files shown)


.github/workflows/rust.yml
```diff
@@ -32,7 +32,7 @@ jobs:
   feature_check:
     strategy:
       matrix:
-        features: ["", "unstable", "benchmarks"]
+        features: ["decoder", "encoder", "decoder,encoder", "decoder,encoder,unstable", "decoder,encoder,benchmarks"]
```
fintelia: We should probably also keep a check somewhere that empty features works. (I don't expect it to do anything useful, but we wouldn't want there to be compile errors either)

linkmauve: Default features do work, but CI always passes `--no-default-features`. Or do you mean building a crate which does absolutely nothing useful (that is, with no features) should be supported?

fintelia: Passing no features at all should be supported. Or at least it shouldn't produce a compile error. (I have no reason to expect that it wouldn't work now, but good to keep a CI job to confirm)

linkmauve: What do you expect the crate to do with no decoder and no encoder enabled?  Export no symbol whatsoever?

fintelia: Yes

fintelia:
One other question. Does this qualify as an API breaking change? I'm not sure if anyone is currently running without default features, but if so their code would probably stop working after this PR.

.github/workflows/rust.yml
linkmauve: I’ve changed the commit adding `compile_error!()` into one which hides everything and disables all dependencies, making the crate a noop when no features are enabled.

I still think that can’t be what users want, and at least a compile-time warning would be better (but there is no `compile_warning!()` in std…).  If you depend on a crate but disable all of its features, why do you depend on it in the first place?  An error would certainly let the user think about what they are doing.

Cargo.toml
```diff
@@ -38,6 +37,9 @@ rand = "0.9.2"
 criterion = "0.7.0"
 
 [features]
+default = ["decoder", "encoder"]
+decoder = ["bitflags", "crc32fast", "fdeflate"]
+encoder = ["bitflags", "crc32fast", "fdeflate", "flate2"]
```
fintelia, line 42: Since `bitflags`, `crc32fast`, and `fdeflate` are needed regardless of whether encoding or decoding is used, I don't think they should be optional dependencies

linkmauve: Since both decoder and encoder are optional now, we can build with no dependency at all. (I still don’t think this configuration should be possible though.)

nicoburns:
Should `fdeflate` not *also* gain feature flags for `encoding` and `decoding`? Needing only decoding is pretty common, and while it might be a small win for a single image codec, it adds up over several.

PR #679, opened 2026-04-09
Expose iCCP name field if present

### LINKED ISSUE #678: Consider exposing the iCCP block name
Hi,

I was tracking down some leftover compatibility bugs in my implementation of an image-rs based GStreamer plugin, and the last I found was that there's no way at present to read the `iCCP` chunk's name field. This is exposed in stock libpng, but is skipped over in image-png.

Would it be possible to add a getter for it?

amyspark:
Hi all,

This MR adds exposure of the iCCP chunk's name field. I implemented this with a variation of the `zTXt` chunk decoding, for which I needed to make `decode_iso_8859_1` pub-crate available.

All feedback is appreciated.

Fixes #678

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -15,7 +15,7 @@ use crate::common::{
     Info, MasteringDisplayColorVolume, ParameterError, ParameterErrorKind, PixelDimensions,
     ScaledFloat, SourceChromaticities, Unit,
 };
-use crate::text_metadata::{ITXtChunk, TEXtChunk, TextDecodingError, ZTXtChunk};
+use crate::text_metadata::{decode_iso_8859_1, ITXtChunk, TEXtChunk, TextDecodingError, ZTXtChunk};
 use crate::traits::ReadBytesExt;
 use crate::{CodingIndependentCodePoints, Limits};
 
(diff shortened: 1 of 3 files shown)


src/decoder/stream.rs
```diff
@@ -1700,32 +1700,28 @@ impl StreamingDecoder {
...
-            let raw: u8 = buf.read_be()?;
-            if (raw == 0 && len == 0) || (raw != 0 && len == 80) {
-                return Err(DecodingError::from(TextDecodingError::InvalidKeywordSize));
-            }
-            if raw == 0 {
-                break;
-            }
+        let (keyword_slice, value_slice) = Self::split_keyword(buf)?;
+
+        if keyword_slice.is_empty() || keyword_slice.len() > 79 {
```
197g: Both cases seem already handled by `split_keyword`. It will reject nul-bytes after the `79` position as well as an empty keyword. (There's a possibility for optimization in a separate PR, the search via `position()` is unbounded but results after `79` are all equivalent).

src/decoder/stream.rs
```diff
@@ -1700,32 +1700,28 @@ impl StreamingDecoder {
...
-                ))
-            }
+        let compression_method = *value_slice
+            .first()
+            .ok_or_else(|| DecodingError::from(TextDecodingError::InvalidCompressionMethod))?;
+
+        if compression_method != 0 {
+            return Err(DecodingError::Format(
+                FormatErrorInner::UnknownCompressionMethod(compression_method).into(),
+            ));
```
197g:
```suggestion
        let compressed_data = match value_slice.split_first() {
            None => {
                return Err(DecodingError::from(TextDecodingError::InvalidCompressionMethod));
            },
            Some((&0, compressed_data)) => compressed_data,
            Some((&method, _)) => {
                return Err(DecodingError::Format(
                FormatErrorInner::UnknownCompressionMethod(method)));
            ));
```

Nit: personal preference, explicit control flow seems more readable and the call in particular avoids the need for justifying the indexing below. `cargo fmt` not applied yet and the call below needs change.

197g (commented):
Both are smaller nits. I hadn't realized there was duplicate code for keyword splitting already, so good find.

197g (approved):

PR #680, opened 2026-04-18
feat: Add option to keep unknown chunks during decoding

ElliotSis:
#### Context
This change is motivated by the integration of the png crate into the Android Skia graphics framework. Android needs to access chunks that are not explicitly handled by the decoder to maintain feature parity with libpng.

#### Feature Parity with libpng
This implementation provides functionality similar to libpng's unknown chunk handling (see [png_set_keep_unknown_chunks](https://linux.die.net/man/3/libpng) and [png_get_unknown_chunks](http://www.libpng.org/pub/png/libpng-manual.txt)).

#### Changes
- Added UnknownChunk struct and unknown_chunks: Vec<UnknownChunk> to Info.
- Added keep_unknown_chunks to DecodeOptions (defaults to false).
- Updated StreamingDecoder to capture and store unrecognized chunks when enabled.
- Added unit tests for preservation logic, memory limits, and critical chunk handling.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -13,7 +13,7 @@ use crate::chunk::{self, ChunkType, IDAT, IEND, IHDR};
 use crate::common::{
     AnimationControl, BitDepth, BlendOp, ColorType, ContentLightLevelInfo, DisposeOp, FrameControl,
     Info, MasteringDisplayColorVolume, ParameterError, ParameterErrorKind, PixelDimensions,
-    ScaledFloat, SourceChromaticities, Unit,
+    ScaledFloat, SourceChromaticities, Unit, UnknownChunk,
 };
 use crate::text_metadata::{decode_iso_8859_1, ITXtChunk, TEXtChunk, TextDecodingError, ZTXtChunk};
 use crate::traits::ReadBytesExt;
@@ -469,6 +469,7 @@ pub struct DecodeOptions {
     ignore_text_chunk: bool,
     ignore_iccp_chunk: bool,
     skip_ancillary_crc_failures: bool,
+    keep_unknown_chunks: bool,
 }
 
 impl Default for DecodeOptions {
@@ -479,6 +480,7 @@ impl Default for DecodeOptions {
             ignore_text_chunk: false,
             ignore_iccp_chunk: false,
             skip_ancillary_crc_failures: true,
+            keep_unknown_chunks: false,
         }
     }
 }
@@ -524,6 +526,13 @@ impl DecodeOptions {
     /// Defaults to `true`
     pub fn set_skip_ancillary_crc_failures(&mut self, skip_ancillary_crc_failures: bool) {
         self.skip_ancillary_crc_failures = skip_ancillary_crc_failures;
+    }
+
+    /// When set, the decoder will keep unknown chunks in the `Info` struct.
+    ///
+    /// Defaults to `false`.
+    pub fn set_keep_unknown_chunks(&mut self, keep_unknown_chunks: bool) {
+        self.keep_unknown_chunks = keep_unknown_chunks;
     }
 }
 
@@ -1068,7 +1077,11 @@ impl StreamingDecoder {
                 ));
             }
             _ => {
-                self.current_chunk.action = ChunkAction::Skip;
+                if self.decode_options.keep_unknown_chunks {
+                    self.current_chunk.action = ChunkAction::Process;
+                } else {
+                    self.current_chunk.action = ChunkAction::Skip;
+                }
                 return Ok(State::ReadChunkData(type_str));
             }
         };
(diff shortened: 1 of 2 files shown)


ElliotSis:
To give some additional context on the design choices made in this PR: it is motivated by integrating the `png` crate into the Android graphics framework (via Skia) as a memory-safe alternative to `libpng`. To maintain feature parity with Android's ecosystem (which relies on custom metadata like NinePatch `npTc`), we need the ability to extract unknown/custom chunks during the decode process.

A few notes on the implementation:

**1. Why only Decoding changes?**

This PR only touches the decoding path because it appears that the `png` crate's `Encoder` already supports writing "unknown" chunks via its existing `write_chunk` method:

``` rust
// Write the "unknown" chunk using the existing API.
writer.write_chunk(ChunkType(*b"uNKn"), b"Hello from unknown chunk!").unwrap();
```

**2. Why not a 1:1 match with libpng's API?**

I initially looked into porting `libpng`'s exact behavior (`png_set_keep_unknown_chunks`), which allows callers to explicitly provide a list of chunk names to keep or discard. However, Skia currently configures `libpng` to blindly keep all unknown chunks (`PNG_HANDLE_CHUNK_ALWAYS` and `num_chunks` set to `0`).

Because the `png` crate already has a globally-enforced `Limits` (which these unknown chunks are tracked against), capturing all unknown chunks using a simple boolean flag seemed reasonable. I opted for this approach as it fulfills our needs while keeping the crate's API surface minimal. The caller can always add filtering on their side if needed.

**Feedback Request:**

Does this approach look acceptable to you? Or would you prefer I implement explicit, name-based chunk filtering (e.g., passing a list of `ChunkType` tags to the decoder) for closer API parity with the C library? 

Happy to adjust the implementation based on your preference.

fintelia:
Capturing only unknown chunks seems like a backwards compatibility hazard. Any time we add a support for a new chunk it'll stop being unknown and won't be captured...

ElliotSis:
Thanks for the feedback! I see your concern… Instead of preserving "unknown" chunks, where the definition of "unknown" may shift from version to version, would you prefer making the caller specify which `ChunkType`s to capture? Something like `DecodeOptions.set_captured_chunks(names: &[ChunkType])`? That would guarantee that these chunks are always skipped and captured, regardless of whether support for them was recently added.

ElliotSis:
I've pushed an update that replaces the `unknown_chunks` with a more explicit `captured_chunks` as per earlier comment. Let me know if it addresses your backwards compatibility concern.

An implementation choice that may need discussion: should these `captured_chunks` be parsed? From a backwards compatibility point of view it’s not immediately clear to me what would be preferable. If a caller handles a specific `ChunkType` in a custom way, perhaps they would want to bypass how the `png` crate later parses them?

Another thing to consider is whether we should allow handling critical chunks.

I went back and forth on this, so I’ve provided both approaches for you to review:
1. The first commit bypasses parsing for captured chunks and allows capturing critical chunks.
2. The second commit restricts capturing to ancillary chunks only and runs the capture alongside normal parsing.

I think the second approach seems better, but let me know what you’d prefer.

Thanks!

ElliotSis:
@fintelia Quick ping on this, curious if you’re ok with exposing these chunks in principle?

src/decoder/stream.rs
```diff
@@ -469,6 +469,7 @@ pub struct DecodeOptions {
     ignore_text_chunk: bool,
     ignore_iccp_chunk: bool,
     skip_ancillary_crc_failures: bool,
+    captured_chunks: Vec<ChunkType>,
```
197g: Should this use `BTreeSet` or `HashSet`? I know that the `ChunkType` isn't ordered so the former is out and the latter would complicate no-std directions but still better than an `O(n)` call to `contains.

ElliotSis: I’m happy to change this to use a `HashSet`. In practice I think this will contain no more than a handful of `ChunkType` though, likely no more than 5, so I wonder if a `Vec` would not be more efficient in this case?

197g: Sure, the small set argument makes sense but it's called once-per-chunk. That's not a hot path, and the difference of querying a set of 5 or so should be small even multiplied along chunks and only a constant factor between Vec/HashSet. On the other hand, the performance gap—or rather pitfall—if one _does_ use the API with a long slice (for whatever reason..) would be quite a lot larger. I'd prefer `HashSet` which also inherently deals with duplicates.

Another direction, however given the complexity better left as a future direction with at most a comment remark: Given that chunk labels are 7-bit, two fields of small, constant sized, array plus a set for the overflow could be combined to have the best properties of both designs. The array can be padded with ChunkType entries that can never occur when the filter list is shorter than the array. However, really sounds more complex than necessary here.

ElliotSis: Sounds good, I went with using a `HashSet` as suggested. Thanks!

197g (approved):
Good point with rejecting critical chunks early. I do like that better than the silent design before. It gives us to flexibility to _maybe_ support it at a later point by splicing those chunks to both internal storage and the capture list.

Issue #604, opened 2025-06-18
API changes required for supporting limits in `image`


Shnatsel:
We have a long-standing ugly API and ugly workarounds in the code of `image` to compensate for missing error reporting in `png` API: https://github.com/image-rs/image/issues/2084

The upcoming semver-breaking release v0.18 is an excellent opportunity to fix that.

fintelia:
The `PngDecoder` struct was changed to returning `ImageResult<bool>` in 0.25.0 so that blocker mentioned in the linked issue has been resolved.

Unfortunately, there's another problem. The `ImageDecoder` trait in the `image` crate requires that the color type and output buffer size to be available before setting the decoding limits (and changing that would make the API much less pleasant to use). But in PNG that is actually somewhat inconvenient because if the IHDR chunk indicates "truecolor" as the color type, then the image is either RGB or RGBA depending on whether there is a tRNS chunk anywhere before the first IDAT. To determine the absence of a tRNS chunk you thus need to consume all the image metadata up to the first IDAT. However, one key reason we wanted decoding limits was to prevent the images with huge amounts of metadata to pull all that data into memory and trigger an OOM.

There's a few ways we could address this, but they pretty much all center around adding a `Seek` bound (which we already did in #558).

fintelia:
Thinking about it, the other change we probably need is change how the ICC profile, EXIF chunk, and text chunks are accessed. Right now they're public fields on the `Info` object, so the API requires them to be eagerly loaded with the rest of the metadata. Other than those three sorts of chunks the rest of the image metadata can be loaded in a constant amount of memory so it is fine to load them before the limits are known.

There can be only one ICC profile and EXIF chunk, so storing their location/size shouldn't be an issue. For text chunks, we already pass `set_ignore_text_chunk` so I guess that handles it.

fintelia:
One puzzle will be how to handle `Encoder::with_info`. Currently it only takes an `Info` struct, so it wouldn't have access to the ICC or EXIF profile if those were switched to loading lazily

197g:
If the Limits are an issue for ICC and EXIF I would propose that we (add an option / switch) to count their memory use towards the usual _buffer_ usage and instead of auxiliary chunks. That is, the caller should be able to decide which set of chunks is important enough for them to allocate persistently and consider part of the necessary return value. Then the eager metadata should not be much of a problem, the fault of applying them at loading time is the `image` trait in my opinion too.

[...]

PR #684, opened 2026-05-05
Provide a way to prevent overshooting past `IEND` (fixes #531)

### LINKED ISSUE #531: Avoid reading beyond the `IEND` chunk
Upfront disclaimer: I think that this issue is a relatively low priority for Chromium - I am mainly reporting it to openly share and discuss some of the more exotic requirements/scenarios that may be imposed on image decoders.

Consider the following scenario (based on what a Skia test does [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/tests/CodecExactReadTest.cpp;l=50-94;drc=46572b4d445f41943059d0e377afc6d6748cd5ca)):

- Main input (e.g. `Read` in Rust or `SkStream` in Skia) contains a PNG image concatenated with some other data
- The input is wrapped in another, secondary `Read` / `SkStream` (which just forwards all calls to the main input;  it is present only to accomodate the fact that the `png::Decoder::new` consumes/takes-ownership of the input).
- The secondary input is passed to `png` crate for decoding the whole image (including the `IEND` chunk via [`Reader.finish`](https://docs.rs/png/latest/png/struct.Reader.html#method.finish) call).

Expected behavior:

- The client code should be able to figure out where the PNG / `IEND` data ends.  In the case of the `Codec_end` test in Skia, this is achieved through the expectations that the main input stream is positioned right after the `IEND` chunk.

Actual behavior:

- `std::io::BufReader` can aggressively read arbitrarily many bytes from the wrapped `Read`, potentially overshooting beyond the `IEND` chunk.

I realize that the `Read` trait doesn't have any APIs to 1) "peek" forward, but 2) only "consume" up to the end of `IEND`.  And therefore I don't quite see a nice way to address this.  Very silly, brainstorming-quality, probably no-go ideas:

* Maybe `Reader` can track and report how many bytes have actually been consumed via `BufRead.consume` (this should be always less then or equal to the number of bytes that have been read via ` Read.read`)
* Or maybe if the input is `Read + Seek` then `BufReader.seek` can be called by `Reader.finish` to position the wrapper reader at the end of the image.

This issue is tracked in https://crbug.com/378936780 from Chromium side.  This issue seems like a relatively low priority from Chromium perspective, but (based on [a test comment](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/tests/CodecExactReadTest.cpp;l=50-52;drc=46572b4d445f41943059d0e377afc6d6748cd5ca)) it may potentially cause issues with adoption of `SkPngRustCodec` in Android.

ElliotSis:
## Buffering issues addressed
1. Overshoot (fixes #531): Standard `BufReader` can overshoot and read past the `IEND` chunk boundary. This fix is needed to support decoding concatenated images from a single stream (simulated in Skia's [Codec_end](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/tests/CodecExactReadTest.cpp;l=50-94) test), where the stream must be left positioned exactly at the end of the first image's `IEND` chunk for subsequent decodes to work.
2. Blocking reads from pipe-like inputs: When decoding PNGs from blocking, non-seekable inputs (like pipes), standard buffering (e.g., standard `BufReader`'s 8KB refills) causes JNI/Skia layers to greedily request more bytes. Once the PNG data is fully consumed, the JNI loop calls `read()` again to satisfy the remaining buffer capacity, causing it to block indefinitely. This is highlighted by a timeout failure in AOSP CTS [testDecodePngFromPipe](https://android.googlesource.com/platform/cts/+/master/tests/tests/graphics/src/android/graphics/cts/BitmapFactoryTest.java#859) test.

## Seek vs limited reads approach
Although issue 1 (overshoot) could theoretically be resolved by seeking backward to the correct boundary after a read that's too wide, this approach does not seem adequate for issue 2 because pipes are not genuinely seekable. 

The AOSP CTS test deadlock occurs because the main test thread retains an open reference to the pipe's `writeFd` even after the `writeThread` closes the `FileOutputStream`, preventing the pipe from naturally closing and returning EOF. Theoretically, we could fix the test by closing `writeFd` in the main thread, but that would only hide a deeper production issue: if a standard decoder eagerly over-reads on an open, blocking stream (where the writer remains alive), the read blocks the entire decode thread indefinitely even after the image has been fully received.

To address both issues with a single solution, a limited buffering approach is implemented that limits reads at the dynamic state boundary, and the currently unused `Seek` bound is removed.

## How it's achieved
- Introduced a new public trait `LimitBufRead` that abstracts controlled buffering, replacing the generic `BufRead` bound on `Decoder` and `Reader` in a backwards-compatible manner. The trait is designed without sealed constraints to potentially be implementable by downstream callers (e.g., potentially a future `SkStream` integration).
- Exposes `LimitBufReader` wrapper which restricts buffer refills to the bytes expected by the decoder state machine.
- Added `expected_read_limit` to `StreamingDecoder` to calculate expected bytes dynamically.
- Added regression tests in `tests/limit_buf_read.rs` verifying concatenated stream decodes and eager-buffered blocking source handling (via `BlockOnLimit` simulating blocking buffering wrappers).

## Testing

To verify the correctness of the new `LimitBufRead` usage and `LimitBufReader` APIs, I have run all existing integration and unit tests by replacing the `BufReader`s with `LimitBufReader`, and wrapping the `Cursor`s in `LimitBufReader`. All tests passed successfully.

## Benchmarks

Additionally, I ran the decoder criterion benchmark suite across four distinct configurations to collect comparative metrics:

1. **Memory standard** (`Cursor<&[u8]>`)
2. **Memory limit** (`Cursor<&[u8]>` wrapped in a `LimitBufReader`)
3. **File standard** (`File` wrapped in a `BufReader`)
4. **File limit** (`File` wrapped in a `LimitBufReader`)

The raw benchmark metrics are compiled below:

<details><summary>Click to expand benchmark results table</summary>
(On my machine)

| Benchmark Image / Case | Memory Standard | Memory Limit | Memory Overhead | File Standard | File Limit | File Overhead |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: |
| **1. Real-world files (`decode`)** | | | | | | |
| `Fantasy_Digital_Painting.png` | `23.75 ms` | `23.99 ms` | **+1.0%** | `24.51 ms` | `24.56 ms` | **+0.2%** |
| `Exoplanet_Phase_Curve_(Diagram).png` | `29.20 ms` | `28.86 ms` | **-1.1%** | `29.32 ms` | `29.48 ms` | **+0.5%** |
| `Lohengrin_Illustrated.png` | `23.96 ms` | `24.90 ms` | **+3.9%** | `26.96 ms` | `27.36 ms` | **+1.5%** |
| `paletted-zune.png` | `11.06 ms` | `11.16 ms` | **+1.0%** | `11.52 ms` | `11.68 ms` | **+1.3%** |
| `kodim02.png` | `3.68 ms` | `3.72 ms` | **+1.1%** | `3.86 ms` | `3.92 ms` | **+1.5%** |
| `kodim07.png` | `4.70 ms` | `4.71 ms` | **+0.4%** | `4.86 ms` | `4.91 ms` | **+1.0%** |
| `kodim17.png` | `3.91 ms` | `3.92 ms` | **+0.3%** | `4.06 ms` | `4.08 ms` | **+0.6%** |
| `kodim23.png` | `3.58 ms` | `3.63 ms` | **+1.4%** | `3.79 ms` | `3.90 ms` | **+2.8%** |
| `Exoplanet_indexed_gimp.png` | `8.89 ms` | `8.93 ms` | **+0.4%** | `9.28 ms` | `9.31 ms` | **+0.4%** |
| `lorem_ipsum_screenshot.png` | `1.22 ms` | `1.23 ms` | **+1.2%** | `1.29 ms` | `1.36 ms` | **+5.8%** |
| `lorem_ipsum_oxipng.png` | `800.91 µs` | `800.60 µs` | **0.0%** | `833.58 µs` | `850.02 µs` | **+2.0%** |
| `tango-icon-new-128.png` | `109.17 µs` | `109.57 µs` | **+0.4%** | `121.41 µs` | `137.98 µs` | **+13.7%** |
| `tango-icon-new-32.png` | `13.87 µs` | `14.16 µs` | **+2.1%** | `23.34 µs` | `37.21 µs` | **+59.4%** |
| `tango-icon-new-16.png` | `7.73 µs` | `8.04 µs` | **+4.1%** | `17.15 µs` | `41.73 µs` | **+143.4%** |
| `Transparency.png` | `87.30 µs` | `88.26 µs` | **+1.1%** | `97.66 µs` | `125.38 µs` | **+28.4%** |
| **2. 4K chunks (Fragmented)** | | | | | | |
| `8x8.png` | `972.78 ns` | `1.18 µs` | **+21.3%** | `10.07 µs` | `20.09 µs` | **+99.5%** |
| `128x128.png` | `13.02 µs` | `14.33 µs` | **+10.1%** | `32.41 µs` | `94.49 µs` | **+191.5%** |
| `2048x2048.png` | `3.99 ms` | `4.48 ms` | **+12.3%** | `8.42 ms` | `21.92 ms` | **+160.3%** |
| `12288x12288.png` | `195.50 ms` | `171.30 ms` | **-12.4%** | `316.29 ms` | `815.29 ms` | **+157.8%** |
| **3. 64K chunks (Stable)** | | | | | | |
| `128x128.png` | `11.27 µs` | `11.99 µs` | **+6.4%** | `29.19 µs` | `43.61 µs` | **+49.4%** |
| `2048x2048.png` | `3.34 ms` | `3.26 ms` | **-2.3%** | `6.89 ms` | `7.89 ms` | **+14.5%** |
| `12288x12288.png` | `133.93 ms` | `141.75 ms` | **+5.8%** | `243.02 ms` | `275.64 ms` | **+13.4%** |
| **4. 2G chunks (Massive-chunk)** | | | | | | |
| `2048x2048.png` | `2.79 ms` | `3.04 ms` | **+8.8%** | `6.51 ms` | `7.25 ms` | **+11.5%** |
| `12288x12288.png` | `131.59 ms` | `142.90 ms` | **+8.6%** | `242.90 ms` | `251.84 ms` | **+3.7%** |
| **5. Incremental row-by-row** | | | | | | |
| `128x128-4k-idat` | `13.81 µs` | `15.16 µs` | **+9.8%** | `33.42 µs` | `92.27 µs` | **+176.1%** |
</details>

Duplicating all the tests and benchmarks on CI felt redundant, and helps keep the PR diff clean and focused. Let me know if you think otherwise.

## Conclusions and BufReader by-default rationale

1. **Negligible memory CPU overhead**: In memory-to-memory decoding (`Cursor`), `LimitBufReader` introduces low overhead (+20% on the tiny fragmented `8x8.png`, otherwise negligible on most files), proving that the limit tracking has little CPU cost.
2. **OS system calls overhead**: Under File I/O, the overhead is dictated by context-switch costs. When chunk structures are highly fragmented (4K chunks), `LimitBufReader` must refill the buffer in smaller increments to respect boundaries safely, resulting in higher overhead (e.g., +160% to +190% regression in pathological cases).
3. **Low overhead on standard real-world files**: For normal real-world photos and diagrams which use larger, continuous `IDAT` chunks, the File I/O overhead of `LimitBufReader` remains low (mostly between 1% and 15%).

This data supports keeping standard `BufReader` as the default decoder baseline for general performance, while exposing `LimitBufReader` as an opt-in wrapper with relatively low overhead for most files.

### DIFF
--- a/src/decoder/input.rs
+++ b/src/decoder/input.rs
@@ -1,0 +1,78 @@
+use std::io::{BufRead, BufReader, Read, Take};
+
+/// Buffered input for the PNG decoder, with optional boundary-limited refills.
+///
+/// This trait is similar to [`BufRead`], but `fill_buf_limited` receives the maximum number of
+/// bytes the decoder currently expects to consume before reaching the next PNG parser boundary.
+///
+/// To avoid reading beyond PNG boundaries, see [`LimitBufReader`] for an implementation that
+/// honors the provided `limit`.
+pub trait LimitBufRead {
+    /// Same as [`BufRead::fill_buf`], but with a hint to cap reads to at most `limit` bytes.
+    ///
+    /// The `limit` boundary serves as a hint to avoid overshooting past `IEND`. Implementors
+    /// are free to discard this value: [`BufRead`] implements this trait discarding `limit`
+    /// entirely.
+    ///
+    /// Implementations that can enforce the `limit` should use it to cap the number of reads, i.e.
+    /// it is valid to return a slice bigger than `limit` if there are more than `limit` unconsumed
+    /// bytes.
+    fn fill_buf_limited(&mut self, limit: usize) -> std::io::Result<&[u8]>;
+
+    /// Same as [`BufRead::consume`].
+    fn consume(&mut self, amt: usize);
+}
+
+/// Blanket implementation for all [`BufRead`] types.
+///
+/// [`LimitBufRead`] replaces the [`BufRead`] trait bound historically required by
+/// [`Decoder`](crate::Decoder) and [`Reader`](crate::Reader) in a backwards-compatible manner.
+///
+/// This implementation deliberately ignores the `limit` dynamic boundary contract, aggressively
+/// pre-fetching data into standard memory buffers to maximize standard I/O caching performance.
+/// However, it does not prevent the underlying reader from being advanced past PNG boundaries.
+impl<T: BufRead> LimitBufRead for T {
+    fn fill_buf_limited(&mut self, _limit: usize) -> std::io::Result<&[u8]> {
+        self.fill_buf()
+    }
+
+    fn consume(&mut self, amt: usize) {
+        BufRead::consume(self, amt);
+    }
+}
+
+/// A wrapper for [`Read`] streams that implements controlled buffering to prevent reading past
+/// `IEND` chunk boundaries.
+///
+/// # Performance note
+///
+/// Because this wrapper caps reads strictly to what the state machine expects, it can trigger more
+/// frequent operating system read syscalls in files characterized by many small chunks. This can
+/// cause context-switching regressions compared to standard large-block pre-fetching (e.g.
+/// `std::io::BufReader`). [`LimitBufReader`] should only be used when boundary safety is required.
+pub struct LimitBufReader<R>(BufReader<Take<R>>);
+
+impl<R: Read> LimitBufReader<R> {
+    /// Creates a new [`LimitBufReader`] wrapping the given [`Read`] stream with the default
+    /// [`BufReader`] capacity.
+    pub fn new(inner: R) -> Self {
+        Self(BufReader::new(inner.take(u64::MAX)))
+    }
+
+    /// Creates a new [`LimitBufReader`] wrapping the given [`Read`] stream with a specified
+    /// `capacity` limit.
+    pub fn with_capacity(capacity: usize, inner: R) -> Self {
+        Self(BufReader::with_capacity(capacity, inner.take(u64::MAX)))
+    }
+}
+
+impl<R: Read> LimitBufRead for LimitBufReader<R> {
+    fn fill_buf_limited(&mut self, limit: usize) -> std::io::Result<&[u8]> {
+        self.0.get_mut().set_limit(limit as u64);
+        self.0.fill_buf()
+    }
+
+    fn consume(&mut self, amt: usize) {
+        BufRead::consume(&mut self.0, amt);
+    }
+}
--- a/src/decoder/read_decoder.rs
+++ b/src/decoder/read_decoder.rs
@@ -1,8 +1,9 @@
 use super::stream::{DecodeOptions, Decoded, DecodingError, FormatErrorInner, StreamingDecoder};
 use super::zlib::UnfilterBuf;
+use super::LimitBufRead;
 use super::Limits;
 
-use std::io::{BufRead, ErrorKind, Read};
+use std::io::ErrorKind;
 
 use crate::chunk;
 use crate::common::Info;
@@ -16,12 +17,12 @@ use crate::common::Info;
 /// * `decode_image_data` - reading from `IDAT` / `fdAT` sequence into `Vec<u8>`
 /// * `finish_decoding_image_data()` - discarding remaining data from `IDAT` / `fdAT` sequence
 /// * `read_until_end_of_input()` - reading until `IEND` chunk
-pub(crate) struct ReadDecoder<R: Read> {
+pub(crate) struct ReadDecoder<R: LimitBufRead> {
     reader: R,
     decoder: StreamingDecoder,
 }
 
-impl<R: BufRead> ReadDecoder<R> {
+impl<R: LimitBufRead> ReadDecoder<R> {
     pub fn new(r: R) -> Self {
         Self {
             reader: r,
@@ -64,8 +65,9 @@ impl<R: BufRead> ReadDecoder<R> {
         image_data: Option<&mut UnfilterBuf<'_>>,
     ) -> Result<Decoded, DecodingError> {
         let (consumed, result) = {
-            let buf = self.reader.fill_buf()?;
-            if buf.is_empty() {
+            let limit = self.decoder.expected_read_limit()?;
+            let buf = self.reader.fill_buf_limited(limit)?;
+            if limit > 0 && buf.is_empty() {
                 return Err(DecodingError::IoError(ErrorKind::UnexpectedEof.into()));
             }
             self.decoder.update(buf, image_data)?
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1,3 +1,4 @@
+pub(crate) mod input;
 mod interlace_info;
 mod read_decoder;
 pub(crate) mod stream;
@@ -10,9 +11,9 @@ use self::stream::{DecodeOptions, DecodingError, FormatErrorInner};
 use self::transform::{create_transform_fn, TransformFn};
 use self::unfiltering_buffer::UnfilteringBuffer;
 
-use std::io::BufRead;
 use std::mem;
 
+pub use self::input::{LimitBufRead, LimitBufReader};
 use crate::adam7::Adam7Info;
 use crate::common::{
     BitDepth, BytesPerPixel, ColorType, Info, ParameterErrorKind, Transformations,
@@ -87,7 +88,7 @@ impl Default for Limits {
 }
 
 /// PNG Decoder
-pub struct Decoder<R: BufRead> {
+pub struct Decoder<R: LimitBufRead> {
     read_decoder: ReadDecoder<R>,
     /// Output transformations
     transform: Transformations,
@@ -122,7 +123,7 @@ impl<'data> Row<'data> {
     }
 }
 
-impl<R: BufRead> Decoder<R> {
+impl<R: LimitBufRead> Decoder<R> {
     /// Create a new decoder configuration with default limits.
     pub fn new(r: R) -> Decoder<R> {
         Decoder::new_with_limits(r, Limits::default())
@@ -291,7 +292,7 @@ impl<R: BufRead> Decoder<R> {
 /// PNG reader (mostly high-level interface)
 ///
 /// Provides a high level that iterates over lines or whole images.
-pub struct Reader<R: BufRead> {
+pub struct Reader<R: LimitBufRead> {
     decoder: ReadDecoder<R>,
     bpp: BytesPerPixel,
     subframe: SubframeInfo,
@@ -327,7 +328,7 @@ struct SubframeInfo {
     consumed_and_flushed: bool,
 }
 
-impl<R: BufRead> Reader<R> {
+impl<R: LimitBufRead> Reader<R> {
     /// Advances to the start of the next animation frame and
     /// returns a reference to the [`FrameControl`] info that describes it.
     /// Skips and discards the image data of the previous frame if necessary.
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -661,6 +661,32 @@ impl StreamingDecoder {
         self.info.as_ref()
     }
 
+    /// Calculates the maximum number of bytes the streaming decoder expects to consume in its
+    /// current state before transitioning to the next state.
+    ///
+    /// This value is used by [`LimitBufReader`](crate::decoder::LimitBufReader) to limit buffer
+    /// refills, avoiding aggressive pre-fetching beyond the current chunk boundary to prevent
+    /// overshoot past the `IEND` chunk.
+    pub(crate) fn expected_read_limit(&self) -> Result<usize, DecodingError> {
+        self.state
+            .as_ref()
+            .map(|state| match state {
+                State::U32 {
+                    accumulated_count, ..
+                } => 4 - *accumulated_count,
+                State::ReadChunkData(_) | State::ImageData(_) => {
+                    if self.current_chunk.remaining == 0 {
+                        4 // For the CRC that we will transition to
+                    } else {
+                        self.current_chunk.remaining as usize
+                    }
+                }
+            })
+            .ok_or_else(|| {
+                DecodingError::Parameter(ParameterErrorKind::PolledAfterFatalError.into())
+            })
+    }
+
     pub fn set_ignore_text_chunk(&mut self, ignore_text_chunk: bool) {
         self.decode_options.set_ignore_text_chunk(ignore_text_chunk);
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,7 +85,9 @@ pub use crate::adam7::{
 pub use crate::adam7::{Adam7Info, Adam7Variant};
 pub use crate::common::*;
 pub use crate::decoder::stream::{DecodeOptions, Decoded, DecodingError, StreamingDecoder};
-pub use crate::decoder::{Decoder, InterlaceInfo, InterlacedRow, Limits, OutputInfo, Reader};
+pub use crate::decoder::{
+    Decoder, InterlaceInfo, InterlacedRow, LimitBufRead, LimitBufReader, Limits, OutputInfo, Reader,
+};
 pub use crate::decoder::{UnfilterBuf, UnfilterRegion};
 pub use crate::encoder::{Encoder, EncodingError, StreamWriter, Writer};
 pub use crate::filter::Filter;
--- a/tests/limit_buf_read.rs
+++ b/tests/limit_buf_read.rs
@@ -1,0 +1,103 @@
+use std::fs;
+use std::io::{self, BufReader, Cursor, Error, ErrorKind, Read};
+
+use png::{Decoder, LimitBufRead, LimitBufReader};
+
+/// A mock [`Read`] stream that rejects reads over the given `limit`.
+///
+/// This replicates integrations where a buffering layer above a blocking pipe attempts to satisfy
+/// the requested size instead of returning a short read.
+struct BlockOnLimit {
+    inner: Cursor<Vec<u8>>,
+    limit: usize,
+    read_bytes: usize,
+}
+
+impl BlockOnLimit {
+    fn new(data: Vec<u8>, limit: usize) -> Self {
+        Self {
+            inner: Cursor::new(data),
+            limit,
+            read_bytes: 0,
+        }
+    }
+}
+
+impl Read for BlockOnLimit {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        let remaining = self.limit - self.read_bytes;
+        if buf.len() > remaining {
+            return Err(Error::new(
+                ErrorKind::WouldBlock,
+                "Blocked on eager overshoot request",
+            ));
+        }
+        let n = self.inner.read(buf)?;
+        self.read_bytes += n;
+        Ok(n)
+    }
+}
+
+fn decode_image<R: LimitBufRead>(reader: R) -> Result<(), png::DecodingError> {
+    let decoder = Decoder::new(reader);
+    let mut reader = decoder.read_info()?;
+    let mut buf = vec![0; reader.output_buffer_size().unwrap()];
+    reader.next_frame(&mut buf)?;
+    reader.finish()?;
+    Ok(())
+}
+
+#[test]
+fn test_buf_reader_overshoots() {
+    let png_data = fs::read("tests/pngsuite/basi0g01.png").unwrap();
+    let mut concat_data = Vec::new();
+    concat_data.extend_from_slice(&png_data);
+    concat_data.extend_from_slice(&png_data);
+
+    let mut cursor = Cursor::new(concat_data);
+    let buf_reader = BufReader::new(&mut cursor);
+    decode_image(buf_reader).expect("Standard BufReader decode failed");
+
+    assert_eq!(cursor.position(), (png_data.len() * 2) as u64);
+}
+
+#[test]
+fn test_buf_reader_blocks_on_blocking_source() {
+    let png_data = fs::read("tests/pngsuite/basi0g01.png").unwrap();
+    let png_len = png_data.len();
+    let blocking_source = BlockOnLimit::new(png_data, png_len);
+
+    let buf_reader = BufReader::new(blocking_source);
+    let res = decode_image(buf_reader);
+
+    assert!(res.is_err());
+    let err = res.unwrap_err();
+    assert!(matches!(
+        err,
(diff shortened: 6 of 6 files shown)


src/decoder/read_decoder.rs
```diff
@@ -64,8 +65,9 @@ impl<R: BufRead + Seek> ReadDecoder<R> {
         image_data: Option<&mut UnfilterBuf<'_>>,
     ) -> Result<Decoded, DecodingError> {
         let (consumed, result) = {
-            let buf = self.reader.fill_buf()?;
-            if buf.is_empty() {
+            let limit = self.decoder.expected_read_limit()?;
```
ElliotSis: FYI: When testing things locally I think I ran into a relatively subtle issue with error handling priority. If the buffer happened to be empty after encountering e.g. a 0-length fdAT, it could start returning `UnexpectedEof` instead of `PolledAfterFatalError`. This change will make `PolledAfterFatalError` take priority.

Shnatsel:
If we're looking at supporting reading directly from pipes, I think we should go ahead and drop the `Seek` bound, otherwise we'll just get runtime errors on non-seekable sources if we ever do make use of it in the future. Right now the `Seek` bound seems to be unused. Removing it would be a non-semver-breaking change as far as I can tell, since we're only relaxing the bounds on the inputs.

The original PR that introduced the `Seek` bound was https://github.com/image-rs/image-png/pull/558. The motivation was the ability to delay extracting metadata and not wasting memory on reading it when it's not needed, specifically https://github.com/image-rs/image-png/issues/604

@197g has just wrapped up and merged a rework of metadata API in `image` in https://github.com/image-rs/image/pull/2672 so I'd like to hear from them about the `image` requirements.

Shnatsel:
@ElliotSis could you document the motivation for making a custom limited reader with custom bookkeeping instead of using `std::io::Read::take()`? Looking at the code, I can't help but wonder if this can be achieved with less complexity by leveraging the standard library.

ElliotSis:
> If we're looking at supporting reading directly from pipes, I think we should go ahead and drop the Seek bound, otherwise we'll just get runtime errors on non-seekable sources if we ever do make use of it in the future. Right now the Seek bound seems to be unused. Removing it would be a non-semver-breaking change as far as I can tell, since we're only relaxing the bounds on the inputs.

Yeah, I’ve considered doing that but I didn’t have the context as to why the `Seek` bound was introduced in the first place. If the bound is necessary to support future functionality I wonder if we could handle that by specializing the implementation later instead?

Since removing it is a non-semver-breaking relaxation and safe to do, I went ahead and updated the PR to drop the `Seek` bound.

> could you document the motivation for making a custom limited reader with custom bookkeeping instead of using std::io::Read::take()? Looking at the code, I can't help but wonder if this can be achieved with less complexity by leveraging the standard library.

To give some context, the requirements I was trying to satisfy were:
1) **Backwards compatibility**: Avoid breaking existing `BufRead` callers, requiring major version bump, or introducing a new API to use the decoder
2) **Opt-in flexibility**: As highlighted in the benchmark results, limiting reads (while still buffering chunk/image data) can cause up to a 200% performance regression in pathological cases (like highly fragmented PNGs), so we may want to make this opt-in
3) **Still buffer large chunks**: The wrapper must still rely on buffering for large chunks to minimize syscalls (keeping the performance regression on real-world images to a negligible 1.5% to 15%)
4) **Buffering capacity flexibility**: Give the caller the flexibility of choosing the buffer capacity, just like `BufReader`  allows today
5) **Buffering implementation flexibility**: Just like with `BufRead`, the caller should be able to implement their own custom buffering layers (e.g., to potentially avoid double-buffering)

Considering the above, I think introducing a new trait and blanket impl for existing `BufRead` would be necessary (if you generally agree with these requirements).

However, I think you’re right that the `LimitBufReader` itself does not have to implement manual bookkeeping. It can likely delegate to `Take` by using a `BufReader<Take<R>>`. This works by allowing dynamic limit updates via `buf_reader.get_mut().set_limit(...)`. I’ve updated the PR to do that too.

ElliotSis:
FYI: I also ended up removing the `Seek` impl from `LimitBufReader` entirely. `Take` implements `Seek` starting Rust 1.80, but the toolchain used here is 1.73. Given we’ve just removed the `Seek` bound from the API, I think that’s reasonable.

197g:
Well, technically we may only need `Seek`-bounds on some methods for going back to the first `IDAT` frame. So if that needs to be dropped from other parts to allow for an easier path

ElliotSis:
@197g the PR currently drops the `Seek` bound from the public interface, I assume this is a comment about future changes when/if we actually introduce `Seek`-based methods? Or did you have a specific method in mind that should keep the bound in this PR?

Thanks both for the feedback!

ElliotSis:
@Shnatsel if this gets merged, wondering if this could potentially be added in 0.18.2 as well? https://github.com/image-rs/image-png/pull/683

Shnatsel:
We should be able cut a release soon after this gets in, no problem.

ElliotSis:
@197g @Shnatsel friendly ping on this: do you think introducing this `LimitBufRead` trait makes sense?

Thanks again for your help with this!

fintelia:
I really don't like this approach. It seems over-complicated and unnecessary.

The concern from #531 about this library doing `Read::read` calls past the `IEND` has already been solved. We switched to directly calling `BufRead::fill_buf` on the provided reader instead, which lets us peek at the stream without consuming past what we should. It sounds like there might be a separate issue of doing a (potentially blocking) `fill_buf` call once we've already read the `IEND`, but that could instead be fixed with a small change to `ReadDecoder`.

There might also be something going on with your code recreating `BufReader`s rather than passing the same one sequentially to multiple decoders, but if so that's something that should be fixed in your code.

The point of the `Seek` bound is flexibility: It could come in handy for skipping and returning to big blocks of metadata if we decide to implement that, for rewinding to the start of looping APNGs, or potentially for implementing future spec additions like the parallel decoding extension that's being worked on.

ElliotSis:
Hey @fintelia, thanks for your input here, it’s much appreciated!
  
> The concern from https://github.com/image-rs/image-png/issues/531 about this library doing Read::read calls past the IEND has already been solved. We switched to directly calling BufRead::fill_buf on the provided reader instead, which lets us peek at the stream without consuming past what we should.
  
My understanding is that it only solves the problem of parsing past `IEND`, it does not prevent pre-fetching past `IEND` from the underlying stream that is being buffered, right?
  
> There might also be something going on with your code recreating BufReaders rather than passing the same one sequentially to multiple decoders, but if so that's something that should be fixed in your code.
  
You’re right that re-using the `BufReader` would solve the data loss, but I’m not sure this is architecturally feasible in Skia / Android. Based on my reading of the code, the `SkCodec` API appears to be stateless: a caller (like the Android framework), passes a `SkStream` as input and can then create another `SkCodec` with the same stream to decode the next part (which is what the `Codec_end` test is doing). If I understand correctly, that would require maintaining a persistent `BufRead` layer across these calls which I’m not sure is technically feasible.
  
> It sounds like there might be a separate issue of doing a (potentially blocking) fill_buf call once we've already read the IEND, but that could instead be fixed with a small change to ReadDecoder.

I’m not sure the issue is with doing too many `fill_buf` calls but rather with the `fill_buf` call itself. My understanding is that `fill_buf` will request as much as it can from the underlying stream, to fill its buffer. If this stream happens to be a blocking pipe, then the [JNI layer](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/libs/hwui/jni/CreateJavaOutputStreamAdaptor.cpp;l=57) will block until this request is fulfilled or it hits EOF. In the `testDecodePngFromPipe` test linked above, the pipe is left open on purpose rather than closing it. This means the `fill_buf` call is blocked on trying to fulfill the bytes that could be past the `IEND` chunk, even though in this case there aren’t any. As mentioned in the PR description, I could technically "fix" the test by closing the pipe but that would just be hiding the underlying issue of blocking even after the image is already received.

> The point of the Seek bound is flexibility […]

As pointed out by @Shnatsel:

> If we're looking at supporting reading directly from pipes, I think we should go ahead and drop the Seek bound […]

So I think while it adds flexibility for future changes, it’s also restrictive in terms of what the caller can pass as input.

> […] It could come in handy for skipping and returning to big blocks of metadata if we decide to implement that, for rewinding to the start of looping APNGs, or potentially for implementing future spec additions like the parallel decoding extension that's being worked on.

I wonder if this could be supported by adding the bound on specific methods (when introduced) instead?

Something like:

```rust
impl<R: LimitBufRead + Seek> Reader<R> {
    pub fn rewind_apng(&mut self) -> Result<(), DecodingError> {
        // …
    }
}
```

> I really don't like this approach. It seems over-complicated and unnecessary.

Yeah, I don’t particularly like adding this new `LimitBufRead` trait as an API surface of the crate. That’s the solution I could find to support backwards compatibility + opt-in mechanism, but if you have other suggestions I’d be happy to implement them.

Thanks again for your help with this!

197g:
Could you provide a PR dropping the unnecessary `Seek` bounds, and then stack the rest of the additions on top? I do agree that we could probably do `rewind_*` separately as tracking an offset from the first `IDAT` chunk coupled with a relative seek should be functionally equivalent to remembering the `Seek::position` (even more efficient in that it requires no initial IO call and relative seeks may have a better implementation).

ElliotSis:
> Could you provide a PR dropping the unnecessary Seek bounds […]

Sounds good, I’ve opened PR #687 to do this. 

> […] and then stack the rest of the additions on top?

I couldn’t quite figure out how to make the PR show the stacked change, but you can see the stacked diff [here](https://github.com/image-rs/image-png/compare/ElliotSis:drop-seek-bounds...ElliotSis:boundary-preserving-mode)

(Edit: now that PR #687 is merged, this PR is showing the right diff)

Thanks @197g !

fintelia:
> My understanding is that fill_buf will request as much as it can from the underlying stream, to fill its buffer. If this stream happens to be a blocking pipe, then the [JNI layer](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/libs/hwui/jni/CreateJavaOutputStreamAdaptor.cpp;l=57) will block until this request is fulfilled or it hits EOF.

This isn't true. The `fill_buf` method is mis-named and doesn't actually fill the buffer. It just guarantees that the buffer is non-empty and thus shouldn't block as long as the buffer is either already non-empty or there's at least 1-byte available to add to the buffer.

In particular, the implementation of `BufReader` in the standard library [uses `read_buf` rather than `read_buf_exact`](https://github.com/rust-lang/rust/blob/968d50ad35115bc2c8c19cb9039f7ed3dfe56a81/library/std/src/io/buffered/bufreader/buffer.rs#L151) (which are unstable equivelents to `read` and `read_exact` respectively) meaning that it'll end up partially filled if getting more bytes required blocking.

While I wasn't around when the trait was created, it does really seem that `BufRead` was basically designed to work sensibly for reading from pipes.

fintelia:
To be clear, peeking past the end of the PNG is allowed by the `BufRead` API. Switching that is a feature request rather than a bugfix.

The trade-off to me is between novice Rust users having a more confusing interface: rather than directly specifying the standard trait they need to fill, we start taking an implementation of our own bespoke trait that is blanket implemented for implementations of a standard trait. Or Skia needs to make some changes to their tests: either to add a framing layer when passing images (i.e. write a `u64` with the size before each image), have a smarter buffering layer that itself watches for PNG chunks and stops when it sees IEND, or refactor the test harness to either share the BufReader or not share the pipe.

My bias is to prioritize having a simpler API when advanced users have reasonable workarounds to get the functionality they need.

src/decoder/input.rs
```diff
@@ -0,0 +1,78 @@
...
+/// A wrapper for [`Read`] streams that implements controlled buffering to prevent reading past
+/// `IEND` chunk boundaries.
+///
+/// # Performance note
+///
+/// Because this wrapper caps reads strictly to what the state machine expects, it can trigger more
+/// frequent operating system read syscalls in files characterized by many small chunks. This can
+/// cause context-switching regressions compared to standard large-block pre-fetching (e.g.
+/// `std::io::BufReader`). [`LimitBufReader`] should only be used when boundary safety is required.
+pub struct LimitBufReader<R>(BufReader<Take<R>>);
```
fintelia, line 53: Even if we do decide to add the new trait, I don't think this should be part of the crate's public API. It could just as well live in Skia and risks confusing people into using it when they should just pass a `Cursor<&[u8]>` or a normal `BufReader`

ElliotSis:
Thanks again for the feedback @fintelia, that makes a lot of sense.

> This isn't true. The fill_buf method is mis-named and doesn't actually fill the buffer. It just guarantees that the buffer is non-empty and thus shouldn't block as long as the buffer is either already non-empty or there's at least 1-byte available to add to the buffer.

Sorry I didn’t express myself well here: I didn’t mean that `fill_buf` itself will block until the whole buffer can be filled, but rather that it will request some bytes from the underlying stream. In the case of AOSP, the JNI adapters may block to fulfill this request even if it crosses the `IEND` boundary (on a blocking pipe where the sender keeps the pipe open but never writes more bytes).

If I recall correctly from debugging this test, it was the very first `fill_buf` request that was causing this hang. This request was asking for 32KB in the case of Skia’s `BufReader` (to match the legacy behavior of 0.17’s internal capacity), which then ended up blocking through these adapter layers:

```
05-07 14:03:53.342 10282 10300 W Skia_FrontBufferedStream: FRONT_BUFFER: read called for size=32768, fOffset=0, fBufferedSoFar=32, fBufferSize=32
05-07 14:03:53.342 10282 10300 W Skia_FrontBufferedStream: FRONT_BUFFER: readFromBuffer requested size=32768, remainder=32, bytesToCopy=32, fOffset=0
05-07 14:03:53.342 10282 10300 W Skia_FrontBufferedStream: FRONT_BUFFER: readDirectlyFromStream requested size=32736, fOffset=32
[…]
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: read called for size=32736, buffer=0xb4000072b3c81f20
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: doRead called for size=32736, capacity=16384
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: doRead calling Java InputStream.read(capacity=16384, offset=0, requested=16384)
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: Java InputStream.read returned n=300
[…]
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: read called for size=32736, buffer=0xb4000072b3c81f20
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: doRead called for size=32736, capacity=16384
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: doRead calling Java InputStream.read(capacity=16384, offset=0, requested=16384)
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: Java InputStream.read returned n=300
[…]
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: doRead loop progress: bytesRead=300, remaining size=32436
05-07 14:03:53.342 10282 10300 W HWUI    : JNI_ADAPTOR: doRead calling Java InputStream.read(capacity=16384, offset=0, requested=16384)
-> Blocking here
```

> Or Skia needs to make some changes to their tests: either to add a framing layer when passing images (i.e. write a u64 with the size before each image) […] or refactor the test harness to either share the BufReader or not share the pipe.

I’m not sure updating the tests is the right thing to do here.

As far as I know, the tests are here specifically because they want to maintain this behavior of avoiding reads beyond the logical end of data, see e.g. the test comments for `Codec_end`:

```
// Test that some SkCodecs do not attempt to read input beyond the logical
// end of the data. Some other SkCodecs do, but some Android apps rely on not
// doing so for PNGs. Test on other formats that work.
```

So I’d rather maintain the same behavior that the tests are asserting against. The same goes for `testDecodePngFromPipe`. I fear that changing the behavior here will likely lead to compatibility issues in Android.

> […] have a smarter buffering layer that itself watches for PNG chunks and stops when it sees IEND […]

I actually initially used this approach, but that felt relatively hacky as it required re-implementing some of the PNG chunk-parsing logic that is already in the crate. This is what led me to send this PR.

> My bias is to prioritize having a simpler API when advanced users have reasonable workarounds to get the functionality they need.

+1, I totally see where you’re coming from and I wasn’t a fan of introducing this trait either.

I can think of a few alternatives:

1) Expose `expected_read_limit` without passing it around when filling the buffer

I wonder if we could potentially achieve the same thing by instead only exposing the `expected_read_limit` somehow, e.g. in a thread local variable? Skia could create and manage the `struct LimitBufReader<R>(BufReader<Take<R>>)`,  implementing `fill_buf` to read from this thread local variable and update its limit.

I don’t quite like using static state for this (re-entrancy risk?) so perhaps it could be exposed some other way, but I also wonder if that could lead to cyclic dependencies (e.g. the buffer struct depending on the decoder and vice-versa).

2) Make this the default behavior

That would require asking the caller to provide a `Read` instead, and then wrapping it in a `BufReader<Take<…>>`. There are quite a few downsides with this:

- Goes back to taking a `Read` as input, which IIRC was changed on purpose to let the caller provide their own `BufRead` (e.g. they can specify the `BufReader` capacity or use another implementation like `Cursor`)
- Breaks backwards compatibility (or introduces double-buffering)
- Introduces a (relatively minor but perhaps non-negligible?) performance regression

3) Duplicate the PNG chunk parsing logic in Skia

Revert back to the initial approach of doing this in Skia. I’m not against it if you don’t want to pollute the crate with this change, as mentioned I totally see your points about keeping the public facing API of this crate as simple as possible.

Do you have any other suggestion? Let me know what you think is the best path forward here!

fintelia:
If your goal is to do at least one syscall per PNG chunk in the image, then implementing that as part of the buffering layer seems like the most direct way to me. But it is your decision. At the end of the day you're being paid to work on this and I'm donating my time.

Issue #690, opened 2026-07-16
MSRV bump to 1.88


okaneco:
The last MSRV bump was a year ago from `1.65` to `1.73` in June 2025 - https://github.com/image-rs/image-png/commit/c39b61bf1b769494ac6cf66d2b72f29ab6cdb8c7

Other than let-chains, the main draw of 1.88 is the `as_chunks` function family which optimizes more reliably than the `chunks_exact` functions which we rely on for areas like filter performance. After this version, 1.93 is probably the next MSRV to target due to the `as_[mut]_array` functions on slices.

90% of requests are 1.88+ according to [lib.rs](https://lib.rs/stats#rustc-usage).

_Rust release dates:_
- `1.65` - Nov. 3, 2022
- `1.73` - Oct. 5, 2023
- `1.88` - Jun. 26, 2025

I'm not aware of a crate or org MSRV policy other than waiting for a little while and bumping when there are specific functions we want to use. MSRV bumps are infrequent enough that I'm fine with how they're currently handled. I think as Rust has matured, it's also less likely that people are stuck on ancient toolchains so it's less of a worry.

PR #691, opened 2026-07-18
Upgrade MSRV to 1.88 and to 2024 Edition 

okaneco:
Based on the thumbs up in https://github.com/image-rs/image-png/issues/690, I figured it'd be worth seeing what this upgrade would look like.
- Bump MSRV to 1.88 in `Cargo.toml` and CI config
- Upgrade Rust edition from 2021 to 2024 since the edition was introduced in 1.85
- Apply clippy suggestions for functions that have been introduced since last MSRV (`is_multiple_of`, `io::Error::other`, let chains, removing unnecessary references)

Going from slices to arrays allowed us to get rid of `TryInto` casts and prevent bugs from mismatched sizes in this refactor due to type checking. I tried to split the unfilter/filter commits into similar chunks of logic.

I can split into other PRs if this is seen as too much.
- I'm mixed on the `is_multiple_of` clippy suggestion being better, this could just be unfamiliarity with using it.
- I reverted non-x86 Paeth filter 3 and 4 bpp special casing which was working around an LLVM issue. Now that we have `as_chunks`, I chose to use that since I was changing the impls around it. If this should be in its own PR, I can remove this commit.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ["1.73.0", stable, beta, nightly]
+        rust: ["1.88.0", stable, beta, nightly]
         os: [ubuntu-latest, windows-latest, macos-latest]
         features: [""]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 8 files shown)


okaneco:
The formatting changes after running `cargo fmt` are very minor. It's mostly reordering crate inputs, adding semi-colons in match arms, and putting some if-else on one line. 19 files are affected.

kornelski (approved):

PR #693, opened 2026-07-19
decoder: make output_buffer_size() respect configured Limits.bytes (fixes #692)

### LINKED ISSUE #692: output_buffer_size() ignores configured Limits.bytes — crate's own documented usage pattern allows ~256 PiB reported size from a 41-byte file
## Summary

`Reader::output_buffer_size()` computes the full output buffer size (`linelen * height`) straight from attacker-controlled IHDR width/height, and only rejects the result if it exceeds `isize::MAX`. It never consults the decoder's configured `Limits.bytes`. The crate's own top-level documented usage pattern (`vec![0; reader.output_buffer_size().unwrap()]`) therefore gives zero protection against a crafted file that declares a huge height while keeping actual compressed data (and hence the one `Limits`-checked quantity — a single row's byte length) tiny.

## Details

The only place `Limits.bytes` is enforced is `Reader::read_until_image_data`, via `self.decoder.reserve_bytes(buflen)` for a **single row's** byte length. `height` never enters any Limits-checked quantity, for both interlaced and non-interlaced images — the per-pass `frame_bytes` estimate is likewise derived purely from declared width/height and only feeds `UnfilteringBuffer`'s internal countdown.

**Proof of concept** (verified against the real published `png = "0.18.1"`): a 41-byte file — PNG signature + IHDR (width=8,000,000, height=4,294,967,295/`u32::MAX`, bit_depth=16, color_type=6/RGBA) + a truncated IDAT chunk header with no payload — makes `decoder.read_info()` succeed and `reader.output_buffer_size()` return `274877906880000000` (~256 PiB), with `Limits::default()` (64 MiB) left completely untouched.

`parse_ihdr` only rejects `width==0||height==0`, not the PNG-spec 2³¹-1 cap, widening the exploitable range — a separate, lower-priority gap not addressed by this fix.

## Impact

Any application following the crate's own documented usage pattern with default `Limits` is unprotected against attempting an astronomically large allocation from a tiny crafted file.

## Suggested fix

See the linked PR: `output_buffer_size()` now additionally checks the computed size against `Limits.bytes` (via a new `ReadDecoder::limits_bytes()` getter), returning `LimitsExceeded` instead of the raw huge number — since both existing call sites (`read_info()`, `next_frame()`) already treat `output_buffer_size() == None` as `LimitsExceeded`, this closes the gap for free at both. Verified: the 41-byte PoC above now fails in ~40-90µs with `LimitsExceeded` instead of succeeding; legitimate images (including a 2000x2000 RGBA8, well under the 64 MiB default) still decode correctly. Full test suite passes unchanged (91→96 tests, all green).

Note: image-rs/image-png#608 is relevant prior art — it already added a narrower check (`output_buffer_size()` rejecting only `isize::MAX` overflow); this PR extends that same code path to also respect the configured `Limits.bytes` budget, per the direction discussed in that issue.

Discovered by the [rust-in-peace](https://github.com/scadastrangelove/rust-in-peace/) security pipeline.

scadastrangelove:
Fixes #692.

## Summary

`output_buffer_size()` now additionally checks the computed `linelen * height` size against the decoder's configured `Limits.bytes` (via a new `ReadDecoder::limits_bytes()` getter), returning `None` (surfaced as `LimitsExceeded` at both existing call sites) instead of an astronomically large number when the budget would be exceeded.

## Design

- Minimal by construction: both `read_info()` and `next_frame()` already treat `output_buffer_size() == None` as `DecodingError::LimitsExceeded`, so this one change closes the gap at both call sites without touching either of them.
- Uses the *current remaining* `Limits.bytes` (post-decrement from any prior row/chunk reservations) as the ceiling, consistent with how the rest of the crate's `Limits` accounting works.
- Does not touch the separate, lower-priority missing PNG-spec 2³¹-1 dimension cap in `parse_ihdr` — that's a distinct, smaller issue that widens the exploitable range but isn't itself the resource-exhaustion bug.
- Callers who legitimately need larger buffers (e.g. multi-GiB rendered images, per the discussion in #608) just call `set_limits` with a larger budget, exactly as already supported — this adds a sanity ceiling on the *reported* number, it doesn't reserve/mutate the budget.

## Testing

- Full test suite: 91→96 tests (5 new), all passing, zero regressions. `cargo fmt --check` clean; `cargo clippy --all-targets` warning count unchanged.
- PoC verified end-to-end: the 41-byte crafted file from #692 (IHDR width=8,000,000, height=u32::MAX) — pre-fix, `read_info()` succeeds in ~44µs and `output_buffer_size()` returns `274877906880000000` (~256 PiB); post-fix, `read_info()` fails in ~40-90µs with `LimitsExceeded`.
- Legitimate decoding confirmed unaffected: 4x4, 16x16, and 2000x2000 RGBA8 round-trips (the latter ~16 MiB output, comfortably under the 64 MiB default) all decode successfully post-fix.

## Open questions for maintainer judgment

1. This uses the *remaining* `Limits.bytes` budget (after any prior reservations) as the ceiling — negligibly stricter than the originally-configured value, but worth confirming that's the intended semantics.
2. Whether the separate PNG-spec dimension cap (`parse_ihdr` not rejecting width/height > 2³¹-1) should land alongside this, since they're closely related — happy to open a follow-up if useful.

Discovered by the [rust-in-peace](https://github.com/scadastrangelove/rust-in-peace/) security pipeline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -663,8 +663,18 @@ impl<R: BufRead + Seek> Reader<R> {
     /// Return the number of bytes required to hold a deinterlaced image frame that is decoded
     /// using the given input transformations.
     ///
-    /// Returns `None` if the output buffer does not fit into the memory space of the machine,
-    /// otherwise returns the byte length in `Some`. The length is smaller than [`isize::MAX`].
+    /// Returns `None` if the output buffer does not fit into the memory space of the machine, or
+    /// if it would exceed the resource budget configured via [`Limits::bytes`] (see
+    /// [`Decoder::set_limits`]); otherwise returns the byte length in `Some`. The length is
+    /// smaller than [`isize::MAX`] and does not exceed the currently configured `Limits`.
+    ///
+    /// Note: unlike other buffers considered by `Limits` (e.g. single decoded rows), the buffer
+    /// whose size is computed here is allocated by the *caller*, not by this crate, so it isn't
+    /// literally reserved against the budget. However, this crate's own documented usage pattern
+    /// is to allocate a buffer of exactly this size (`vec![0; reader.output_buffer_size().unwrap()]`),
+    /// so a maliciously crafted file that declares a huge `height` (while keeping a single row
+    /// cheap) must not be able to make this function report a value wildly out of proportion with
+    /// the configured `Limits`, even though decoding hasn't allocated anything that large itself.
     pub fn output_buffer_size(&self) -> Option<usize> {
         let (width, height) = self.info().size();
         let (color, depth) = self.output_color_type();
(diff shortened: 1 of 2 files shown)


197g:
"Make snprintf error when buffer size would have exceeded `bufsz`".

PR #704, opened 2026-07-20
Remove manual SIMD unfiltering

fintelia:
I've now tracked down a low-end ARM system (a Raspberry Pi Compute Module 4), and run end-to-end benchmarks with [`corpus-bench`](https://codeberg.org/fintelia/corpus-bench) to see whether the manual `std::simd` unfiltering actually helps. And the results are rather underwhelming. Overall the hand-written SIMD version is slightly slower... 

main (rustc nightly + unstable feature flag):
```
image-png:           53.8 MP/s (average)    46.7 MP/s (geomean)
```

Without manual SIMD unfiltering (also rustc nightly + unstable feature flag):
```
image-png:           54.7 MP/s (average)    47.3 MP/s (geomean)
```
---

At this point I'm not quite sure what to do. The original PRs measure on entirely different chips that I don't have easily accessible and failed to address concerns about how we'd benchmark going forward. I really don't like the prospect of carrying a whole bunch of code and complexity when we can no-longer replicate the performance benefit. So maybe the right thing is just to drop this code knowing it can be added back if we get a repeatable way to confirm that it helps?


197g:
I'd say we leave them in but use less broad runtime detection for calling them. Afterall, these are all algorithmically equivalent alternatives. I don't see really any complexity in carrying them with us other than the selection logic. Actually, could we swap it at runtime? For large enough images there's little harm in running both variants once or twice and then using the faster between the two from that point onwards. That's not 100% perfect, could hit a power budget or data dependent costs later, but may be quite a good default.

fintelia:
The problem is that I haven't been able to find any hardware that currently benefits from the SIMD unfiltering so I don't know how to parameterize the selection logic or demonstrate a benefit from runtime selection. On x86, the scalar version _used_ to be slower but then a newer version of LLVM improved the codegen. The biggest benefit was supposed to be on low-end ARM devices, but I tracked one down and found it also didn't benefit from manual SIMD. However, since I tested on an Arm Cortex A72 rather than an Arm Cortex A520 like the original author did, I can't be totally confident that the other chip isn't still faster with SIMD.

If we keep the code assuming that it is still worthwhile on Cortex-A520, then we're in an awkward spot where we can't meaningfully work on this code since we have no way of knowing whether a seemingly minor tweak might regress performance on the CPU we're keeping it around for. For instance, the SIMD routines have scalar tail handling that duplicates code from our scalar unfiltering methods, but it has been left that way because the original PR alluded that reusing them could make performance a bit worse.

(This situation also annoys me because one of the less-visible things I've been doing for the last few years is periodically re-benchmarking various codec components to catch regressions / see if they can be optimized more. And I can't reasonably do that for code that's only meant for systems I don't have access to)

197g:
We could name this variant and treat it as effectively immutable with an API in the decoder to explicitly request it. This should work for Chromium? I really don't see us working on it either but it does not block us from having a separate other routine here, either. It just needs to be clear that you'll only be guaranteed to get this by explicitly asking for it, and we may fall back to anything else which we think might be faster. Having an explicit name would also be better for deprecating it if we really want to drop it, and we'd fallback to the LLVM scalar variant if it's not applicable (e.g. due to deprecation).

PR #708, opened 2026-08-04
Fix streaming APNG encoding

### LINKED ISSUE #411: Broken APNG is produced with `StreamWriter`
This code produces broken APNG without any errors:

```rust
fn main() {
    use std::io::Write;

    let file = std::io::BufWriter::new(std::fs::File::create("out.png").unwrap());
    let mut encoder = png::Encoder::new(file, 8, 8);

    // 8-bit grayscale, 2 frames
    encoder.set_color(png::ColorType::Grayscale);
    encoder.set_depth(png::BitDepth::Eight);
    encoder.set_animated(2, 0).unwrap();
    encoder.validate_sequence(true);

    let writer = encoder.write_header().unwrap();
    let mut stream_writer = writer.into_stream_writer().unwrap();

    // Frame 0
    stream_writer.set_frame_delay(1, 1).unwrap();
    stream_writer.write_all(&[0u8; 64]).unwrap();

    // Frame 1
    stream_writer.set_frame_delay(1, 2).unwrap();
    stream_writer.write_all(&[0xffu8; 64]).unwrap();

    stream_writer.finish().unwrap();
}
```

```console
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/png-animation-test`
$ magick identify -verbose out.png
identify: IDAT: unknown compression method `out.png' @ error/png.c/MagickPNGErrorHandler/1492.
$ ffprobe out.png
ffprobe version n6.0 Copyright (c) 2007-2023 the FFmpeg developers
(... snip ...)
[apng @ 0x560b1cafb580] inflate returned error -3
Input #0, apng, from 'out.png':
  Duration: N/A, bitrate: N/A
  Stream #0:0: Video: apng, gray(pc), 8x8, 100k tbr, 100k tbn
```

hjanuschka:
`StreamWriter` prefixed the first frame's `IDAT` payload with an APNG sequence number, selected `IDAT` again for the second frame, and did not finish animation frame accounting before validation.

Start frame controls and compression when the first pixels arrive, finish accounting after each compressed frame, and only add sequence numbers to `fdAT` chunks. This also makes frame settings apply to the intended frame and resets row filtering between frames. Tests cover all compression backends, small multi-chunk frames, and a separate default image.

This was found while investigating animated output in libjxl/jxl-rs#858. The jxl-rs workaround is libjxl/jxl-rs#859 upstream fix would be so much better, please let me know if you want me to address anything!

Fixes #411

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -1192,10 +1192,10 @@ impl<'a, W: Write> ChunkWriter<'a, W> {
         // TODO (maybe): find a way to hold two chunks at a time if `usize`
         //               is 64 bits.
         const CAP: usize = u32::MAX as usize >> 1;
-        let curr_chunk = if writer.images_written == 0 {
-            chunk::IDAT
-        } else {
+        let curr_chunk = if writer.images_written > 0 && writer.info.frame_control.is_some() {
             chunk::fdAT
+        } else {
+            chunk::IDAT
         };
         ChunkWriter {
             writer,
@@ -1236,17 +1236,18 @@ impl<'a, W: Write> ChunkWriter<'a, W> {
         assert_eq!(self.index, 0, "Called when not flushed");
         let wrt = self.writer.deref_mut();
 
-        self.curr_chunk = if wrt.images_written == 0 {
-            chunk::IDAT
-        } else {
+        self.curr_chunk = if wrt.images_written > 0 && wrt.info.frame_control.is_some() {
             chunk::fdAT
+        } else {
+            chunk::IDAT
         };
 
         match wrt.info.frame_control {
             Some(_) if wrt.should_skip_frame_control_on_default_image() => {}
             Some(ref mut fctl) => {
                 fctl.encode(&mut wrt.w)?;
-                fctl.sequence_number += 1;
+                fctl.sequence_number = fctl.sequence_number.wrapping_add(1);
+                wrt.animation_written += 1;
             }
             _ => {}
         }
@@ -1293,17 +1294,11 @@ impl<'a, W: Write> Write for ChunkWriter<'a, W> {
         }
 
         // index == 0 means a chunk has been flushed out
-        if self.index == 0 {
-            let wrt = self.writer.deref_mut();
-
-            // Prepare the next animated frame, if any.
-            let no_fctl = wrt.should_skip_frame_control_on_default_image();
-            if wrt.info.frame_control.is_some() && !no_fctl {
-                let fctl = wrt.info.frame_control.as_mut().unwrap();
-                self.buffer[0..4].copy_from_slice(&fctl.sequence_number.to_be_bytes());
-                fctl.sequence_number += 1;
-                self.index = 4;
-            }
+        if self.index == 0 && self.curr_chunk == chunk::fdAT {
+            let fctl = self.writer.info.frame_control.as_mut().unwrap();
+            self.buffer[0..4].copy_from_slice(&fctl.sequence_number.to_be_bytes());
+            fctl.sequence_number = fctl.sequence_number.wrapping_add(1);
+            self.index = 4;
         }
 
         // Cap the buffer length to the maximum number of bytes that can't still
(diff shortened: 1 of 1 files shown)


fallenmi (changes requested):
This review was prepared with OpenAI Codex assistance under the account owner's authorization.

Blocking: please preserve the missing-data error for a stream on which no pixels were written. `StreamWriter::new()` now leaves `to_write` at zero until the first non-empty `write()`, so an immediate `stream.finish()` reaches `finish_mut()` with `Wrapper::Chunk` and no outstanding byte count and returns `Ok(())`. The caller can then successfully call `Writer::finish()`, producing a 45-byte PNG containing only `IHDR` and `IEND`, with no `IDAT`. This regresses the base behavior and contradicts `StreamWriter::finish()`'s documented guarantee that abandoned image data is reported.

I reproduced this through the public API on exact PR base `4f873e4a628c0e8dbb4204fb263697298f89eb01`, head `cd6eca127b4cc54c8b745b33c0ceebbe88176e02`, current `master` `05ccdb9c8ae7fc50f4787ba2888a1c515adad944`, and current merge `ec5e4df8b619dddf91b685b4d8fb3fb9ba380832`: the empty/partial/complete stream contract test is 3/3 on base and current master but 2/3 on head and current merge, failing only the empty-stream case. The upstream unit, integration, binary, and example suites otherwise pass on all four revisions, and all 24 live CI checks are successful.

Keeping the lazy compressor/header start while initializing `line_len` and `to_write` from `chunk_writer.next_frame_info()` restores the empty-stream error; that minimal variant passes the 3/3 contract oracle, both new APNG tests, and both existing streaming round-trip tests.

hjanuschka:
Good catch. Initialized the expected frame byte count while keeping compressor/header startup lazy, and added regression coverage for finishing without image data.

fallenmi (approved):
This follow-up review was prepared with OpenAI Codex assistance under the account owner's authorization.

The changed head addresses my previous blocker. Relative to reviewed `cd6eca127b4cc54c8b745b33c0ceebbe88176e02`, `StreamWriter::new()` now initializes `line_len` and `to_write` from `ChunkWriter::next_frame_info()` while preserving lazy compressor/header startup. An untouched or partial stream therefore reaches `finish_mut()` with a positive remaining byte count and returns `MissingData`; a complete write still enters and finishes the frame normally. The new `stream_validation_rejects_missing_image_data` regression covers the empty-stream case.

I reviewed exact current `master`/base `05ccdb9c8ae7fc50f4787ba2888a1c515adad944`, head `5dac28ce87791331d27cbfaee30c0f9b69679b96`, and current merge `1f3db0a8203f82a32d4a91b2f67bd484f2f15f8d`. The PR is open, non-draft, and mergeable; all 24 live checks and both Actions workflows are successful. This follow-up was source/API-only; I did not rerun a local build or test suite.

src/encoder.rs
```diff
@@ -1470,12 +1426,11 @@ impl<'a, W: Write> StreamWriter<'a, W> {
...
         let filtered_buf = vec![0; in_len];
 
-        let mut chunk_writer = ChunkWriter::new(writer, buf_len);
+        let chunk_writer = ChunkWriter::new(writer, buf_len);
         let (line_len, to_write) = chunk_writer.next_frame_info();
-        chunk_writer.write_header()?;
 
         Ok(StreamWriter {
-            writer: Wrapper::from_level(chunk_writer, compression)?,
+            writer: Wrapper::Chunk(chunk_writer),
```
197g: What prompted removing the compression here?

src/encoder.rs
```diff
@@ -1664,62 +1619,92 @@ impl<'a, W: Write> StreamWriter<'a, W> {
...
+        }
         if self.to_write > 0 {
             let err = FormatErrorKind::MissingData(self.to_write).into();
             return Err(EncodingError::Format(err));
         }
 
+        if !matches!(self.writer, Wrapper::Chunk(_)) {
+            self.finish_frame()?;
+        }
         self.flush()?;
```
197g: Having this use `Write::flush` was already a little bit of a stretch as it needed it to also detects overwrite but since the encoder finish and stream flush is now happening inside of `finish_frame` this not only stretches it a bit too far but also the `match` inside this function seems redundant when we're now never calling it with `Wrapper::{Flate2, FDeflate}`. 

I might consider inlining the remainder of the body split between here and `finish_frame`. I'm not even sure if the `self.index > 0` branch was ever appropriate except for its use in `finish` since it also triggers if called in the middle of a line.

197g (commented):
The double duty of `Wrapper::Chunk` is quite hard to follow and there's now also its comment explaining it  (on `Wrapper`) that is outdated by its new usage. It alludes to it being an error-recovery state *after* a frame but now it also serves as the initial state of a new frame before any data is added (and before the compression is applied).

The latter part is fine with me! The symmetry of `start_frame` and `finish_frame` is far clearer. And yes the LLM review is right in saying that the fix resolves some state transitions better Maybe another explicit separate state in `Wrapper` would be better?

src/encoder.rs
197g: Sorry, misunderstood the new control flow with `Chunk`. That's now reviewed in the main comment.

src/encoder.rs
hjanuschka: Good point. I renamed `Wrapper` to `FrameState` and made the uncompressed state explicitly `Ready`, covering both the initial state and the point between completed frames. I also removed the redundant flush from `finish_mut()` and stopped treating a mid-scanline `Write::flush()` as an error, with a regression test for that case.

197g (approved):
@fallenmi Want to run Codex again? I'm not seeing anything obvious to have been missed and the extra regression tests also inspire confidence.

PR #709, opened 2026-08-06
Remove imports already in the prelude

linkmauve:
Since edition 2021 (bumped in af389624950f5a22d21389a8161b04874abdd43e), `std::convert::TryInto`, `std::convert::TryFrom` and `std::iter::FromIterator` are already imported in the prelude, so we don’t have to import them manually.

This was originally part of #670, but can go standalone.


197g (approved):
Interesting that clippy doesn't appear to lint it.

PR #710, opened 2026-08-06
Remove import of miniz_oxide, unused since 0.17.12

linkmauve:
It is still a direct dependency of `flate2`.

This was originally part of #670, but can go standalone.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ bitflags = "2.0"
 crc32fast = "1.2.0"
 fdeflate = "0.3.3"
 flate2 = "1.0.35"
-miniz_oxide = { version = "0.8", features = ["simd"] }
 
 [dev-dependencies]
 approx = "0.5.1"


Cargo.toml
```diff
@@ -25,7 +25,6 @@ bitflags = "2.0"
 crc32fast = "1.2.0"
 fdeflate = "0.3.3"
 flate2 = "1.0.35"
-miniz_oxide = { version = "0.8", features = ["simd"] }
```
197g, line 28: I believe we also want to make sure that the `simd` feature is active if this is the backend that's used. We have simd in `png` at all configurations anyways and this ensures an expected level of performance. But, debatable reasoning.

fintelia: Yeah, IIRC this feature flag makes `miniz_oxide` use the `simd_adler32` crate for checksum calculation. And since we depend on `simd_adler32` unconditionally via `fdeflate` it makes sense to use it everywhere 

Edit: though it looks like flate2 now directly enables the feature and switched to version 0.9 in the meantime... https://github.com/rust-lang/flate2-rs/blob/c8be6352ceadbed425335d150fb2dd9e6e0a96b6/Cargo.toml#L28

linkmauve: Since flate2 switched to version 0.9, this PR is even more useful, it prevents miniz_oxide from being built twice, once for flate2 and once for nothing.  This happens currently on image-png master, after running cargo update.

fintelia (approved):

PR #717, opened 2026-09-06
Interlaced encode

### LINKED ISSUE #716: Support Adam7 (interlaced) encoding
Apart from glycin supporting CMYK internally, missing interlacing support is currently the only remaining blocker for Inkscape to port to libglycin.

Looks like oxipng supports converting an existing PNG to Adam7, but it sounded like there would be interest in supporting it directly in the png crate.

197g:
Closes: #716 

In this implementation we do not support it for `StreamWriter` but only for `write_image_data`. We assume that you never want to have an arbitrary buffer (longer than a single line) in the former case. Having the caller `Write` data in order of its adam7 passes could also be very surprising, but we may relax it like that later (or via a separate type, or a separate method parallel to `into_stream_writer` etc.).

The bulk of this is code for efficiently extracting passes into a separate line buffer. The loops should optimize in considerably different manners for many combinations of sub-byte depths and pass (1..7 mostly). Further evidence in encoding performance might evolve this at a later point.

### DIFF
--- a/src/encoder.rs
+++ b/src/encoder.rs
@@ -127,7 +127,7 @@ impl From<io::Error> for EncodingError {
 
 impl From<EncodingError> for io::Error {
     fn from(err: EncodingError) -> io::Error {
-        io::Error::new(io::ErrorKind::Other, err.to_string())
+        io::Error::other(err.to_string())
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/encoder.rs
197g, line 997: Self-review: oops, forgot to restart here with interlaced encoding as well. The code structure seems sub-optimal for this reason but I really don't want to build type-algebra abstractions to double dispatch compression and lines. Eh.

